PR #6, opened 2014-08-06
Added 4 per glyph text rendering and sizing functions,

beoran:
This included support for getting size kerning information.
Example of use is in ex_font and ex_ttf.

The use of this is for custom text output. This is the minimal amount of functions needed to allow more efficient custom text output to be written by the users of Allegro.


beoran:
@SiegeLord @trentg Is there any interest in this pull request? I think it could be useful to let the user do their own complex text layout, but perhaps there isn't much demand for these functions? If you guys show no interest I'll close this pull request until someone else expresses interest.

trentg:
It makes sense, but I haven't looked at the patch yet. I'll look at it tonight.

trentg:
By the way, what is the use case of this? My mind says it's useful but I haven't thought of a concrete example.

SiegeLord:
The original motivation for this was for custom markup in games, e.g. to draw one particular word in a sentence some different color. This can be done using the current functions by drawing one character strings, but it's very inconvenient.

beoran:
I hope this commit cleans it up mostly. 

I added some ASCII diagrams to the documentation to explain al_get_glyph_dimensions better. 
There are currently 3 sizing functions. al_get_glyph_width is analogous with al_get_text_width, and al_get_glyph_dimensions is an analogue of al_get_text_dimensions. The latter is for when you only need the width of the glyph, the former is for when you need the detailed TTF size information. The example ex_ttf has a new string that tries to demonstrate it by drawing boxes.  

The function al_get_glyph_advance is specially for left-to-right writing of text, although with additional flags it could support right-to-left, up-to-down, etc as well later.

I didn't implement any flags on al_draw_glyph, because most flags on al_draw_text don't make sense for a single glyph and can be achieved though use of judicious placement with the use of al_get_glyph_dimensions or al_get_text_width.

beoran:
Finally if all is well, should I squash the commits or is it OK like this?

addons/ttf/ttf.c
```diff
@@ -785,7 +792,7 @@ ALLEGRO_FONT *al_load_ttf_font_stretch_f(ALLEGRO_FILE *file,
 
     if ((result = FT_Open_Face(ft, &args, 0, &face)) != 0) {
         ALLEGRO_ERROR("Reading %s failed. Freetype error code %d\n", filename,
-	   result);
+      result);
```
SiegeLord: This needs to be re-indented.

examples/ex_ttf.c
```diff
@@ -45,17 +45,36 @@ static void render(void)
...
     al_clear_to_color(white);
 
     al_hold_bitmap_drawing(true);
    
-    al_draw_textf(ex.f1, black, 50,  50, 0, "Tulip (kerning)");
-    al_draw_textf(ex.f2, black, 50, 100, 0, "Tulip (no kerning)");
+    al_draw_textf(ex.f1, black, 50,  20, 0, "Tulip (kerning)");
+    al_draw_textf(ex.f2, black, 50,  80, 0, "Tulip (no kerning)");
+    
+    x = 50 ; y = 140;
```
SiegeLord: This is very awkward looking.

addons/ttf/ttf.c
beoran: ok

examples/ex_ttf.c
beoran: @SiegeLord: What to do about it, though? I had to shift them up for the al_get_glyph_dimensions example. Should I extract the examples from ex_font and ex_ttf and move them to ex_glyph? Or...? Any suggestion is welcome.

SiegeLord: I just meant that I'd prefer it to be:

```c
x = 50;
y = 140;
```
:)

beoran: OIC, I'll fix that throughout tthe example then. :)

beoran:
@SiegeLord, I hope that fixes your remarks then. If no one objects I'll merge this change tomorrow.

elias-pschernig:
I think we should try and squash commits like this before pushing them... it doesn't really matter, but it's also not very useful to have all those intermediate commits in the history. And when bisecting something later it actually can cost more time.

elias-pschernig:
Also, personally I prefer rebasing over merging.

beoran:
Well, I'll take care of squashing the commits better next time, but a pull request is always a merge, AFAICS. I prefer to work with pull requests so I can get some discussion of the contribution.

elias-pschernig:
If you merge the pull request with normal git and not the github web interface you can also rebase it (github will still realize it was merged). Anyway, it's not that important and not sure we even want a policy one way or another... I suppose a lot of people like the merge commits for some reason.

PR #14, opened 2015-06-07
Enable logging for release modes by default

SiegeLord:


fatcerberus:
I'm not sure this is a good idea, the one time I tried to enable release logging MSVC locked up "Generating code".  And it wasn't a fluke either--it locked up consistently at the same point every time.  Turned off release logging, it built fine.

SiegeLord:
Yes, indeed. The same thing happened with GCC. The culprit is src/opengl/extensions.c. I believe I have worked around this adequately (I'll verify with MSVC myself at some point as well): https://github.com/liballeg/allegro5/pull/14/files#diff-786589e1d3a302d50dd4ee890fb8dcb9R276

PR #16, opened 2015-06-25
Constant color blending support

fatcerberus:
This is a generalized way to set a constant blend factor other than ZERO or ONE, and as it turns out is supported natively by both OpenGL and Direct3D9.

New blend factor constants:
* `ALLEGRO_CONST_COLOR`
* `ALLEGRO_INVERSE_CONST_COLOR`

New functions:
* `void al_set_blend_color(ALLEGRO_COLOR);`
* `ALLEGRO_COLOR al_get_blend_color(void);`

Usage:
```c
/* masking source pixels (like draw_tinted_bitmap!) */
ALLEGRO_COLOR const_color = al_map_rgba(0, 128, 0, 255); // dark green
/* out = src * const_color + dest * 1 */
al_set_blender(ALLEGRO_ADD, ALLEGRO_CONST_COLOR, ALLEGRO_ONE);
al_set_blend_color(const_color);

/* averaging src and dest pixels */
ALLEGRO_COLOR const_color = al_map_rgba(128, 128, 128, 255); /* avg. RGB, leave alpha as-is */
/* out = src * const_color + dest * const_color */
al_set_blender(ALLEGRO_ADD, ALLEGRO_CONST_COLOR, ALLEGRO_CONST_COLOR);
al_set_blend_color(const_color);
```

This implements constant color blending for all three backends: OpenGL, Direct3D and software (yes, I went the extra mile. :smile:) and has been tested with all three on Windows.  I don't have the means to test on other platforms, however support for the new modes has been added to ex_blend2 to facilitate such testing.


SiegeLord:
Awesome! The only small issue is that there are tabs all over the place, would you mind removing the ones you added (I know there are some in Allegro's source already).

fatcerberus:
Sorry about the tabs, I prefer them for my own projects and MSVC doesn't have an option to use different settings for different projects.  I'll go fix that now.

fatcerberus:
Okay, I fixed the tabs and also updated make_scanline_drawers.py to properly implement the software blending.  I hadn't noticed scanline_drawers.inc was autogenerated and edited it directly, that would have caused issues.

I only have one question, should `al_get_blend_color()` return an ALLEGRO_COLOR directly instead of taking a pointer?

SiegeLord:
Yes, returning it by-value would be more consistent with the rest of Allegro.

fatcerberus:
Okay, done. :smile:

fatcerberus:
There, I was a good boy and added the new additions to the API documentation. :-)

fatcerberus:
Stand by, don't merge this yet, something is borked.  The blending is all wrong, and al_get_blend_color() is returning ridiculous RGBA values (as in, in the millions).

fatcerberus:
False alarm, everything is good.  I just misunderstood how const color blending actually works and thought the results were wrong, but they're correct.  All three drivers agree on the output. :stuck_out_tongue_closed_eyes: 

I did get rid of a few tabs I missed though.

SiegeLord:
This was awesome work! Thanks! I had to rewrite all the commits to change the committer to my account, as the sourceforge pre-receive hook rejects non-sourceforge emails for that. This just means that this pull request will have to be closed manually, but the commits are all merged as is (with a few additions by me).

I've added some tests and disabled this feature for GLES 1, which doesn't have `glBlendColor`.

fatcerberus:
No problem, it was fun to implement, especially figuring out how the software blending worked. :smile:

PR #18, opened 2015-07-17
Add al_reparent_bitmap function.

allefant:
As well as al_get_bitmap_x/y. Also update documentation and add an
example.


src/bitmap.c
```diff
@@ -449,6 +449,25 @@ ALLEGRO_BITMAP *al_create_sub_bitmap(ALLEGRO_BITMAP *parent,
 }
 
 
+/* Function: al_reparent_bitmap
+ */
+void al_reparent_bitmap(ALLEGRO_BITMAP *bitmap, ALLEGRO_BITMAP *parent,
```
SiegeLord: This should check that bitmap is actually a sub-bitmap.

allefant: it does now

PR #19, opened 2015-08-02
Allow injection of allegro events into a queue.

rcorre:
Previously, an ASSERT in al_emit_user_event checked that the event was in the
user space.

However, there doesn't seem to be a good reason to limit users from injecting
regular allegro events into the queue.
This could be useful for testing and simulating input devices.

As a matter of fact, this is already possible in release builds, as the only
thing stopping you is the ASSERT.
I say we get rid of it unless there's a good reason for it to be there.

This was discussed in a forum thread a few months ago:
https://www.allegro.cc/forums/thread/615220

Everyone seemed to agree that it would be useful, and the changes this patch
applies to the source were suggested by Edgar Reynaldo in that thread.

This also includes an example of 'faking' events in ex_inject_events.

### DIFF
--- a/docs/src/refman/events.txt
+++ b/docs/src/refman/events.txt
@@ -817,7 +817,7 @@ See also: [ALLEGRO_EVENT_SOURCE]
 
 ## API: al_emit_user_event
 
-Emit a user event.
+Emit an event from a user event source.
 The event source must have been initialised with [al_init_user_event_source].
 Returns `false` if the event source isn't registered with any queues,
 hence the event wouldn't have been delivered into any queues.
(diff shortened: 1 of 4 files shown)


beoran:
Seems useful and harmless enough. The documentation of al_emit_user_event should be updated as well, though.

SiegeLord:
I'll have a few comments about this this evening, hold on.

rcorre:
Maybe this is one of the things you were going to mention, but I do like the idea you posted about providing functions like `al_emit_key_event`

SiegeLord:
Yes, that'd be nice, but what I mainly was concerned about is that the documentation should make it clear that this does not update the keyboard/mouse/joystick states.

beoran:
Yes, but shouldn't it update those states? Otherwise it's less useful for programs that use those states.

SiegeLord:
It's ideal, but it would require a large refactor of the event system to make that possible. For now, I'd be content with just a warning. In the future, the `al_emit_keyboard_event` or something could handle this update, while the `al_emit_user_event` will be the backend function.

beoran:
OK that's fair enough. So, @rcorre, could you please add the required documentation to this pull request? Oh, and one more thing, is the example code in line with the Allegro code standard?

rcorre:
I've updated the docs to note that `al_emit_user_event` can be used to emit any event type, and that emitting input events will not update the state structures.

Not sure about the code standard - where is that described?
I tried to imitate the style of other examples.

SiegeLord:
The format is fine, I think. Merged as ebcff378ef15cc313ac159772dae7211821e966b and b5082454eb4aa29b8ce4e8d9f9dfa7e9c3ca0b77. Thanks!

beoran:
Great, thanks for your efforts!

PR #472, opened 2015-08-18
Add support for grabbing the X11 window handle from an Allegro Display

Ragora:
I was recently doing some programming with Allegro and I noticed that according to the API there was no way to get the handle of an Allegro display on Linux: http://liballeg.org/a5docs/trunk/platform.html
There didn't even seem to be any semblance of this feature in the issue tracker as well (unless I somehow missed it!).

Which is quite odd, considering that there was ways to do it in Windows and OSX but not on *nix (with X11 that is anyway). This functionality is required on *nix if you want to try and initialize a full-fledged renderer in an Allegro window that implements such an initialization, but only with the window handle (such as Irrlicht, in my case).

With the changes proposed here, it makes the following code segment possible:

```
#include <iostream>
#include <sys/types.h>

#include <X11/Xlib.h>
#include <X11/Xatom.h>

#include <allegro5/allegro.h>
#include <allegro5/allegro_x.h>

#include <irrlicht/irrlicht.h>

int main(int argc, char* argv[])
{
    // Try to init Allegro!
    if (!al_init())
    {
        std::cout << "Failed to init Allegro" << std::endl;
        return -1;
    }

    // Create a display
    ALLEGRO_DISPLAY* allegroDisplay = al_create_display(640, 480);
    al_set_window_title(allegroDisplay, "This is an Allegro Window");

    XID allegroWindowID = al_get_x_window_handle(allegroDisplay);

    // Setup the creation request for Irrlicht
    irr::SIrrlichtCreationParameters creationParameters;
    creationParameters.WindowId = (void*)allegroWindowID;
    irr::IrrlichtDevice* device = irr::createDeviceEx(creationParameters);

    // Get the video driver and scene manager
    irr::video::IVideoDriver* video = device->getVideoDriver();
    irr::scene::ISceneManager* scene = device->getSceneManager();

    // Add something to the scene so we can see
    irr::scene::ISceneNode* cube = scene->addCubeSceneNode();
    cube->setPosition(irr::core::vector3df(0.0f, 0.0f, 30.0f));
    cube->setRotation(irr::core::vector3df(20.0f, 100.0f, 150.0f));
    cube->setMaterialFlag(irr::video::EMF_LIGHTING, true);

    // Add a light to provide illumination
    irr::scene::ILightSceneNode* light = scene->addLightSceneNode();
    light->setPosition(irr::core::vector3df(50.0f, 0.0f, 30.0f));
    light->setRadius(55.0f);
    light->getLightData().DiffuseColor = irr::video::SColor(255, 0, 0, 255);

    // Mouse control doesn't work as it should, but we put it in the scene so we can see.
    irr::scene::ICameraSceneNode* camera = scene->addCameraSceneNodeFPS();

    // Needs to hook an Allegro event to quit on window close
    while (device->run())
    {
        video->beginScene(true, true, irr::video::SColor(255, 255, 255, 255));
        scene->drawAll();
        video->endScene();
    }

    return 0;
}

```
And here's the output: http://i.imgur.com/aksTIu2.png

So I guess here's the question: Is there any specific reason as to why this is not supported?


src/x/xwindow.c
```diff
@@ -390,4 +390,13 @@ void _al_xwin_check_maximized(ALLEGRO_DISPLAY *display)
 }
 
 
+bool al_get_x_window_handle(ALLEGRO_DISPLAY *display, XID *out)
+{
+    if (!display)
```
SiegeLord: Could you use 3 space indents here?

src/x/xwindow.c
```diff
@@ -390,4 +390,13 @@ void _al_xwin_check_maximized(ALLEGRO_DISPLAY *display)
...
+        return 0;
```
SiegeLord: `return false;` here and `return true;` below.

cmake/FileList.cmake
```diff
@@ -315,3 +315,7 @@ set(ALLEGRO_INCLUDE_ALLEGRO_ANDROID_FILES
 set(ALLEGRO_INCLUDE_ALLEGRO_PLATFORM_FILES_GENERATED
     include/allegro5/platform/alplatf.h
     )
+    
+set(ALLEGRO_INCLUDE_ALLEGRO_X_FILES
+    include/allegro5/allegro_x.h
+    )
```
SiegeLord: Add a training newline here.

include/allegro5/allegro_x.h
```diff
@@ -0,0 +1,38 @@
...
+ *         \ \  __ \ \ \ \  \ \ \   /'__`\ /'_ `\/\`'__\/ __`\
+ *          \ \ \/\ \ \_\ \_ \_\ \_/\  __//\ \L\ \ \ \//\ \L\ \
+ *           \ \_\ \_\/\____\/\____\ \____\ \____ \ \_\\ \____/
+ *            \/_/\/_/\/____/\/____/\/____/\/___L\ \/_/ \/___/
+ *                                           /\____/
+ *                                           \_/__/
+ *
+ *      Header file for X specific functionality.
+ *
+ *      By Trent Gamblin.
```
SiegeLord: Remove this line (since that's not you) :).

include/allegro5/allegro_x.h
```diff
@@ -0,0 +1,38 @@
...
+#include <X11/Xlib.h>
+
+#ifdef __cplusplus
+   extern "C" {
+#endif
+
+/*
+ *  Public X-related API
+ */
+AL_FUNC(bool, al_get_x_window_handle, (ALLEGRO_DISPLAY *display, XID *out));
```
SiegeLord: I think just returning the XID is better. You can `ASSERT` that display is not NULL.

src/x/xwindow.c
Ragora: Sure, I use 4 unless otherwise instructed.

cmake/FileList.cmake
Ragora: Only reason I did that was I noticed that true/false can be not declared under older standards, but I would figure the rest of the codebase is using true/false anyway so I'm not sure why I bothered!

SiegeLord:
That's pretty sweet. The reason it wasn't there before was just because nobody added it.

include/allegro5/allegro_x.h
Ragora: Ah, I was following suit with the Windows equivalent: https://github.com/liballeg/allegro5/blob/5.1/src/win/wwindow.c#L1290

But alright!

Ragora:
And perhaps maybe the Allegro function should be renamed to al_get_x_window_id to more closely correspond with X11's naming?

SiegeLord:
Yeah. `al_get_x_window_id` is better.

Ragora:
Also squashed the commits down slightly.

SiegeLord:
Thanks!

PR #480, opened 2015-08-26
Fix #478 by implementing al_set_new_window_title() and al_get_new_window_title()

### LINKED ISSUE #478: Add al_set_new_window_title
Like it says above. Avoids an visually annoying split second wrong title.

beoran:
Fix #478. 
Also adds an example for the two functions and for al_set_window_tittle().

### DIFF
--- a/examples/ex_window_title.c
+++ b/examples/ex_window_title.c
@@ -26,7 +26,7 @@ int main(int argc, char **argv)
    ALLEGRO_FONT *font;
    int step = 0;
    const char *text;
-   char   title[TITLE_SIZE] = ""; 
+   char   title[TITLE_SIZE] = "";
    bool done = false;
    bool redraw = true;
 
@@ -43,9 +43,9 @@ int main(int argc, char **argv)
 
    al_init_font_addon();
    init_platform_specific();
-   
+
    text = NEW_WINDOW_TITLE;
-   
+
    al_set_new_window_title(NEW_WINDOW_TITLE);
 
    display = al_create_display(640, 480);
@@ -71,7 +71,7 @@ int main(int argc, char **argv)
    al_register_event_source(queue, al_get_timer_event_source(timer));
    al_register_event_source(queue, al_get_display_event_source(display));
 
-   
+
 
    while (!done) {
       ALLEGRO_EVENT event;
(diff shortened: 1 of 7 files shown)


examples/ex_window_title.c
```diff
@@ -0,0 +1,112 @@
+/* An example showing bitmap flipping flags, by Steven Wallace. */
```
SiegeLord: This comment is wrong.

include/allegro5/display.h
```diff
@@ -110,6 +110,9 @@ AL_FUNC(void, al_set_new_display_flags, (int flags));
 AL_FUNC(int,  al_get_new_display_refresh_rate, (void));
 AL_FUNC(int,  al_get_new_display_flags, (void));
 
+AL_FUNC(void, al_set_new_window_title, (char * title));
```
SiegeLord: Extra space after the `char *`

src/tls.c
```diff
@@ -181,6 +185,43 @@ ALLEGRO_EXTRA_DISPLAY_SETTINGS *_al_get_new_display_settings(void)
...
+
+
+
+/* Function: al_get_new_window_title
+ */
+const char * al_get_new_window_title(void)
+{
+   thread_local_state *tls;
+   
+   /*Return al_get_prgram name in case of error or if not set before. */
```
SiegeLord: `al_get_prgram` -> `al_get_app_name`

beoran:
With the cleanups above, hopefully I have fixed the points raised in SiegeLord's comments.

beoran:
@SiegeLord is it OK to merge now?

SiegeLord:
It occurs to me that this new string is never freed. I don't know if we leak other things like that, but perhaps we shouldn't. If you want, you could add a `_al_tls_shutdown` function to tls.c and free/NULL it when `al_uninstall_system` is called. Otherwise, it looks good to me.

beoran:
That's a good idea. I'll implement a way to clean up that string.

beoran:
@SiegeLord, there is another snag, al_restore_state and al_save_state, with the api they have cannot be used to manage allocated memory because there is no al_destroy_state. I'd say we just punt an use a static buffer instead, and document the limit of the buffer.

SiegeLord:
Ah, yikes. Yeah, a static buffer seems to be a good solution.

beoran:
@SiegeLord how is this? Sorry for stripping some spaces at the end of the lines.

beoran:
If no one has any more comments I will merge this tomorrow.

fatcerberus:
Looks good to me.  Static buffer is 255; nobody should need to set a window title longer than that. :-)

PR #483, opened 2015-08-29
Added al_premul_rgba[_f] function.

allefant:


beoran:
Yes, this is a useful function, not in the least because the fact that Allegro 5 uses pre multiplied alpha is a FAQ on the forums. This function us useful for someone who wants to use non pre multiplied alpha for colours, which is a rather common use case. The pull request looks fine, but let's await some more comments from the other devs before merging this.

SiegeLord:
It looks fine to me.

beoran:
Let me merge it then.

PR #486, opened 2015-09-03
Fix #446, support clipboard on Linux, Windows and OSX.

### LINKED ISSUE #446: copy&paste support/clipboard addon
Originally reported by: elias-pschernig

Original Ticket: "alleg/feature-requests/28":https://sourceforge.net/p/alleg/feature-requests/28

Imagine a game with a built-in chat, as many multi-player games have. Supporting copy&paste from other applications is quite important there. While A5 has not and will not have a GUI, what it should provide is a cross-platform way to exchange text with the clipboard.

See also: http://www.allegro.cc/forums/thread/599395

beoran:
I hope someone else will contribute mobile support as I am struggling with Android and can't develop for IOS.


trentg:
al_has_clipboard_text doesn't make sense. Desprite jonrafkind's grievances, I think al_is_clipboard_text is fine. You have C functions like isspace and so so on that have nothing to do with classes.

trentg:
Nevermind, I see you renamed it again.

beoran:
I made it al_clipboard_has_text() in commit 4e592e2 in analogy with _al_pixel_format_has_alpha(), al_ustr_has_prefix(), al_ustr_has_prefix_cstr(), al_ustr_has_sufffix(), and al_ustr_has_sufffix_cstr(). I assume that's OK then?

trentg:
Yeah that makes sense. Thanks.

trentg:
This doesn't require external programs does it? I've seen patches for copy and paste that used external programs on X.

trentg:
Also, I have/had basic functions for copy and paste on both Android and iOS... I can try to find them. Basically though they were something like:

    const char *get_clipboard();
    void set_clipboard(const char *);

Does that make sense?

trentg:
https://github.com/Nooskewl/monster-rpg-2/blob/master/src/iphone.mm

https://github.com/Nooskewl/monster-rpg-2/blob/master/src/android.cpp

https://github.com/Nooskewl/monster-rpg-2/blob/master/android-proj/src/ca/nooskewl/monsterrpg2/MO2Activity.java

Those are links to the code for iOS, C code for Android and complementing Java for Android.

beoran:
No, no external programs. It's pure X, lifted mostly from SDL (hence the credits in LICENSE). I must say copy paste in X is way too convoluted... That's one reason this is for clipboard text only. Arguably we could support bitmaps too but that's much harder and not that useful for now.

beoran:
And thanks for those links. I will try to implement those, but could you compile and test them when I do? I cannot compile IOS code, and I can't seem to get my android emulator to work...

trentg:
Sure not a problem.

beoran:
OK, @trentg. I tried to implement android and iphone support. I had to change a few things a bit from your example due to deprecated API's etc. The android support compiles, but is untested. I don't know if the iphone support even compiles, so please test and if possible correct it for me. Thanks!

trentg:
Android worked without changes. iOS needed a few small changes. Some of it was just to add the new files to the iOS Xcode project. Here's a patch:

https://gist.github.com/trentg/c18753424447fd20967d

beoran:
@trentg I applied your patch, thanks again for that! 

I grepped the Allegro src and include directories for tabs, and while I could find thousands of them (bstrlib.c is the main offender), I found no tabs in the other files I contributed, so I assume they are OK.

elias-pschernig:
Hm, since you mention bstrlib.c, we probably should go through their commit history and see if we should update: https://github.com/websnarf/bstrlib/commits/master

It seems they added utf8 support independent of us in the meantime, so maybe we could replace some of our utf8 code with just using the new bstrlib one.

fatcerberus:
Honestly I found Allegro's current UTF8 routines to be awkward to use and ended up using this awesomely compact decoder in minisphere:
http://bjoern.hoehrmann.de/utf-8/decoder/dfa/

elias-pschernig:
Allegro's routines wouldn't change, just the implementation (if bstrlib implements the required functionality, I didn't look myself).

beoran:
Well the new bstrlib might be worth updating to, however, please make a new issue if you think it's a good idea. Meanwhile I'll probably merge this PR tonight if no one has any objections.

android/allegro_activity/src/Clipboard.java
```diff
@@ -0,0 +1,88 @@
+package org.liballeg.android;
```
SiegeLord: This file has mixed tabs/spaces, and since it's completely new, it's worth keeping it consistent.

docs/src/refman/display.txt
```diff
@@ -679,3 +679,71 @@ or resets the system back to the default state
...
 It returns true if the state was set successfully,
 otherwise false.
+
+
+
+## Clipboard
+
+With the clipboard API of Allegro, text can be copied from and to the clipboard.
+Currentlly, only UTF8 encoded text is supported, and it only works on Linux,
+Windows and, OSX . Other platforms such as Android and IOS are currently
```
SiegeLord: Extra space before the period. The last sentence is now incorrect?

docs/src/refman/display.txt
```diff
@@ -679,3 +679,71 @@ or resets the system back to the default state
...
+
+### API: al_get_clipboard_text
+
+This function returns a pointer to a string, allocated with [al_malloc] with
+the text contents of the clipboard if available. If not text is available on the
+clipboard then this function returns NULL. You must call [al_free] on the
+returned pointer when you don't need it anymore.
+
+Beware that text on the clipboard on Windows may be in Windows format, that is,
+it may have carriage return newline combinations for the line endings in stead
```
SiegeLord: 'in stead' is typically written as 'instead'

docs/src/refman/display.txt
```diff
@@ -679,3 +679,71 @@ or resets the system back to the default state
...
+
+
+Since: 5.1.12
+
+See also: [al_get_clipboard_ustr], [al_get_clipboard_text],
+          [al_get_clipboard_text]
+
+### API: al_clipboard_has_text
+
+This function returns true if the clipboard has text available, false if not.
```
SiegeLord: No need to say 'false if not', as bool has only two states :).

fatcerberus: Going to play devil's advocate on that one: to fully preserve the meaning you'd have to change the text to "returns true only if".  I tend to document bool returns the same as in the text above.

include/allegro5/clipboard.h
```diff
@@ -0,0 +1,44 @@
...
+#include "allegro5/base.h"
+#include "allegro5/display.h"
+#include "allegro5/utf8.h"
+
+#ifdef __cplusplus
+   extern "C" {
+#endif
+
+AL_FUNC(char *, al_get_clipboard_text, (ALLEGRO_DISPLAY *display));
+AL_FUNC(ALLEGRO_USTR *, al_get_clipboard_ustr, (ALLEGRO_DISPLAY *display));
```
SiegeLord: As implemented right now, the ALLEGRO_USTR variants are a bit pointless since the platform functions are implemented in terms of `char *`. I think it's a bit misleading to do it this way, since e.g. a user might expect to be able to get/set a string with an internal NUL character to the ALLEGRO_USTR variants, but in fact have it be truncated. I don't know how well the platforms support the internal NUL character (I couldn't get it to work on Linux) on the clipboard, so ALLEGR_USTR variants might be pointless even if implemented correctly.

SiegeLord:
Could you also squash some of these commits before merging? I'm fine with just a single commit, or perhaps a commit per platform.

Great work, btw :+1:.

android/allegro_activity/src/Clipboard.java
beoran: I forgot to grep the android directory for tabs :p.  I'll fix it and make it all spaces.

docs/src/refman/display.txt
beoran: Yeah, I'll fix that.

docs/src/refman/display.txt
beoran: Ok, I'll fix that too.

docs/src/refman/display.txt
beoran: Ok I'll make it "This function returns true if and only if the clipboard has text available."

include/allegro5/clipboard.h
beoran: Point taken. I'll drop these USTR variants since they effectively do nothing you can't do with ustr yourself.

beoran:
There, that should integrate your suggestions, and it's all squashed down in  one commit.

beoran:
I merged it after fixing one last devious oversight. It's far from perfect, but already quite usable. Please feel free to open a new issue to improve on this.

fatcerberus:
I'm going to have to try this out now. :)

PR #490, opened 2015-09-07
Fix several OSX compilation and deprecation warnings.

beoran:
In particular Gestalt() has been replaced by NSAppKitVersionNumber.

### DIFF
--- a/addons/native_dialog/osx_dialog.m
+++ b/addons/native_dialog/osx_dialog.m
@@ -4,6 +4,12 @@
 
 #import <Cocoa/Cocoa.h>
 
+/* Normally we should include aintosx.h to get this prototype but it doesn't 
+* work. So repeat the prototype here as a stopgap measure.
+*/
+void _al_osx_clear_mouse_state(void);
+
+
 bool _al_init_native_dialog_addon(void)
 {
    return true;
@@ -509,6 +515,13 @@ static _AL_VECTOR displays = _AL_VECTOR_INITIALIZER(DISPLAY_INFO *);
 static volatile ALLEGRO_EVENT_QUEUE *queue = NULL;
 static ALLEGRO_MUTEX *mutex;
 
+/* Declare this here to avoid getting a warning. 
+* There seems to be no better way to do this...
+*/
+@interface NSApplication(AllegroOSX)
+- (void)setAppleMenu:(NSMenu *) menu;
+@end
+
 #define add_menu(name, sel, eq)                                          \
         [menu addItem: [[[NSMenuItem allocWithZone: [NSMenu menuZone]]   \
                                     initWithTitle: name                  \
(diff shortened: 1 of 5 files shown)


beoran:
This doesn't solve all problems, since there are harder cases where an API got deprecated and basically the functionality needs to be rewritten to use a newer API on newer OSX systems. But it's a start, so...

beoran:
It's defined from OSX 10.1 and onwards. AFAICS Allegro supports OSX 10.4 and onwards.

src/macosx/hidjoy.m
```diff
@@ -765,17 +765,18 @@ static bool get_joystick_active(ALLEGRO_JOYSTICK *joy_)
 
 #endif // Leopard+
 
+#ifndef NSAppKitVersionNumber10_5
+#define NSAppKitVersionNumber10_5 949
```
SiegeLord: Why do you need to define these manually?

addons/native_dialog/osx_dialog.m
```diff
@@ -509,6 +515,13 @@ - (NSMenuItem *) build_menu_item:(ALLEGRO_MENU_ITEM *)aitem
 static volatile ALLEGRO_EVENT_QUEUE *queue = NULL;
 static ALLEGRO_MUTEX *mutex;
 
+/* Declare this here to avoid getting a warning. 
+* There seems to be no better way to do this...
+*/
+@interface NSApplication(AllegroOSX)
```
SiegeLord: Could you move this next to the osx_clear_mouse declaration, so all declarations are in the same spot?

addons/native_dialog/osx_dialog.m
```diff
@@ -4,6 +4,12 @@
 
 #import <Cocoa/Cocoa.h>
 
+/* Normally we should include aintosx.h to get this prototype but it doesn't 
+* work. So repeat the prototype here as a stopgap measure.
```
SiegeLord: For these block comments, we typically add a space before the '*' from the second line on.

src/macosx/hidjoy.m
beoran: The constants are not defined when compiled using an earlier or equal version of the NSAppKit. For example, on NSAppKit 10.5, the constant above will not be defined. On 10.10  NSAppKitVersionNumber10_10 is not  defined. At least  that is what I understood from the dev docs and forum threads I read.

addons/native_dialog/osx_dialog.m
beoran: OK

addons/native_dialog/osx_dialog.m
beoran: OK

src/macosx/hidjoy.m
SiegeLord: Oh crud, that's a bummer. I guess this is fine then.

beoran:
@SiegeLord, it should be as per your instructions now.

beoran:
I merged this, but the macos port needs a lot more love. I'll really need some help from someone else who knows OSX better to clean it up better.

PR #491, opened 2015-09-07
Revert WinMain() definition in END_OF_MAIN()

dacap:
There is no different in the WinMain signature between x86 and x64
platforms. A HINSTANCE is a void*, and it's adjusted automatically to
a 32-bit or 64-bit pointer by the compiler.

Fix problem introduced in 234ac0ab5b5790ff6d63e0433c89007f491cefde


dacap:
This PR fix the compilation in both platforms. Anyway examples on x64 are not working yet.

trentg:
Thanks.

Yamakuzure:
I confirm that the change worked around a problem that was self-made and is not necessary. The WinMain signature update can be safely reverted.

Tested with Visual Studio 2013 and both 32bit and 64bit applications with and without debugging modes.

PR #492, opened 2015-09-07
Fix possible random crashes on Windows restoring DirectDraw surfaces

dacap:
This fix was introduced in my own Allegro 4.4 port included in Aseprite here:
https://github.com/aseprite/aseprite/commit/434059d517d786078c8cf528fd8bea70aa62fb73

There were several crash reports (memory dumps) about this specific problem so I think other people using Allegro 4.4 would love this fix.


trentg:
Looks like it makes sense.

PR #494, opened 2015-09-08
Implement al_resume_timer

rcorre:
Currently there is no easy way to pause/resume a long-running timer, as
`al_start_timer` will reset the counter.

`al_resume_timer` behaves like `al_start_timer`, but does not reset the counter.

Discussion:
https://www.allegro.cc/forums/thread/615715/1016397#target


trentg:
Looks pretty good but could you address my comments? Thanks!

src/timernu.c
```diff
@@ -159,6 +159,43 @@ static void shutdown_timers(void)
 
 
 
+// logic common to al_start_timer and al_resume_timer
+// al_start_timer : passes resetCounter = true to start from the beginning
+// al_resume_timer: passes resetCounter = false to preserve the previous time
+static void enable_timer(ALLEGRO_TIMER *timer, bool resetCounter)
```
SiegeLord: `resetCounter` should be `reset_counter`

rcorre:
Thanks @trentg and @SiegeLord. I've cleaned up the example and renamed the parameter.

beoran:
@trentg  Sorry but at least on android the only allowed signature for an allegro program is int main(int rgs, char **argv). The example program won't compile on android if it uses int main(void). I just found out about this myself when I tried it out. @rcorre, probably you will have to undo that one change, sorry for that.

rcorre:
@beoran thanks for explaining. I saw that in another example and copied just because I thought it was a style thing. I'll put that back in

beoran:
@trentg Is it OK now for you? I'll merge this tomorrow if no one objects.

trentg:
I think it's OK. Thanks.

beoran:
Merged! @rcorre, thanks for your contribution!

rcorre:
Thanks! Looking forward to being able to pause timers.

beoran:
You can now, if you use the git version of Allegro... :)

PR #498, opened 2015-09-10
Fix 455. Functions to get cpu amount and ram size.

beoran:
This allows Allegro's users to get some information about the hardware 
of the system they are running Allegro on.


examples/ex_cpu.c
```diff
@@ -0,0 +1,92 @@
+/* An example showing the use of al_get_cpu_count() and al_get_ram_size(). */
```
SiegeLord: I think it'd be sufficient to just use the log for this addon, and not bother with a display.

beoran: I thought so too at first, but the log doesn't display or even give debug feedback on Android, so I coulnd't see anything when testing it. I made a different issue for that point, but it will probably be a while before someone implements support for native dialogs in Android... So for now, I have to use the display to see the results on Android. Don't mind if I keep it like that for now until the Android port is fixed?

SiegeLord: Oh I see, carry on then.

trentg: There is one way to get log output on Android, and that's ALLEGRO_DEBUG but it only works with debug builds of Allegro.

beoran: Sure, but for an example app end user it's easier to see the results on the screen of their phone than to have to look into the adb debug logs. So I'd like to leave ex_cpu as an app with a display until the native dialog on androids work well enough to display the log on the phone.

trentg: Yeah, was just commenting, it's fine as it is.

beoran:
OK, I proceed like this then. Thanks for your comments!

PR #499, opened 2015-09-10
Fix an issue with formating of flags.

rlam1:
On several function definitions, the listing of the flags had wrong formatting, causing the final html doc page to only show the first flag in **bold** and the other ones part of this flag. That would probably confuse some users.

image


beoran:
Seems like a nice improvement.

rlam1:
Yeah... I think I used the correct formatting to produce the desired look, Took as an example the flag listing from `al_open_native_text_log` so I assume these edits will work.

SiegeLord:
Thanks a lot!

PR #502, opened 2015-09-15
warn more strongly about font fallback loops

elias-pschernig:
git remote rename github allefant

^ that hopefully makes me not confuse remotes anymore :) Still have to do the non-ttf fonts so opening a proper pull request now


elias-pschernig:
Yes, that's what I had at first. Then I decided I want to handle the case where you set a bitmap font as fallback for a TTF font (and vice versa). Which seemed way too much work so I almost gave up the idea - but then realized your newly added glyph functions make it somewhat easy :)

beoran:
Ah yes, I hadn't considered mutual recursion between BMP and TTF. That does make things a bit harder, I'd love to see how you'll handle that case.

PR #505, opened 2015-09-22
Fix unregistering a bitmap saver.

elias-pschernig:
There was an overzealous assert in there. And prevent future mistakes like that by removing some code duplicated four times.


SiegeLord:
Yay for less copy-pasted code!

PR #506, opened 2015-09-23
Added bitmap file identification.

### LINKED ISSUE #461: File Signatures
Originally reported by: nikakd

Original Ticket: [alleg/feature-requests/43](https://sourceforge.net/p/alleg/feature-requests/43)

When an extension is not known, or if a file has no extension, allegro should try to read the first bytes of the file and check if it knows the file signature, and load the file with the appropriate loader, instead of failing.
http://en.wikipedia.org/wiki/List_of_file_signatures

elias-pschernig:
Fixes #461. Identification is only done when opening a file with no
extension, or passing NULL as identification name to the _f variants. If
a file has an extension, or an extension is passed to a _f function, no
attempt to autodetect is made even if loading fails. This way there
should be little problems with backwards compatibility - those cases
meant an immediate failure to load before with no attempt to even open
the file.

### DIFF
--- a/addons/image/bmp.c
+++ b/addons/image/bmp.c
@@ -1180,4 +1180,29 @@ bool _al_save_bmp(const char *filename, ALLEGRO_BITMAP *bmp)
    return retsave && retclose;
 }
 
+
+bool _al_identify_bmp(ALLEGRO_FILE *f)
+{
+   uint16_t x;
+   uint8_t y[2];
+   al_fread(f, y, 2);
+   if (memcmp(y, "BM", 2) != 0)
+      return false;
+
+   if (!al_fseek(f, 14, ALLEGRO_SEEK_SET))
+      return false;
+
+   x = al_fread16le(f);
+   switch (x) {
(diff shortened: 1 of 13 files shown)


addons/image/bmp.c
```diff
@@ -1180,4 +1180,28 @@ bool _al_save_bmp(const char *filename, ALLEGRO_BITMAP *bmp)
    return retsave && retclose;
 }
 
+
+bool _al_identify_bmp(ALLEGRO_FILE *f)
+{
+   uint16_t x;
+   x = al_fread16le(f);
+   if (x != 'B' + 'M' * 256)
```
SiegeLord: I feel like doing a memcmp here just like with the other identifiers might be easier to look at.

SiegeLord:
I like it!

fatcerberus:
This'll be good to have, sometimes when loading images from, e.g. WAD-like packages you don't necessarily know the original file extension so it helps if the image loader can figure it out for you.

PR #507, opened 2015-09-26
Fix al_save_bitmap failing with uppercase filenames on Windows

SupSuper:
Use case-insensitive string comparisons for file extensions instead.


elias-pschernig:
According to the documentation the extension is matched case insensitively, so this looks good to me.

PR #509, opened 2015-10-01
mingw toolchain: Define CMAKE_FIND_ROOT_PATH for i686-w64-mingw32 too

amarillion:
Looks the definition of CMAKE_FIND_ROOT_PATH was just missing in this one instance
I've tested cross-compilation before and after, this solves the "directinput not found" problem reported here https://www.allegro.cc/forums/thread/615769/


SiegeLord:
Thanks, merged as 83a1c151eb8004579abb91a574b080ebf48e3a5e.

PR #513, opened 2015-10-20
Implemented get_monitor_info for Android

Orangeyness:
Hey,

Porting a project to android I had a problem with my method that gets the screen size,

```
ALLEGRO_MONITOR_INFO monitor_info;
bool success = al_get_monitor_info(0, &monitor_info);
if (!success)
    ...

int width = monitor_info.x2 - monitor_info.x1; 
int height = monitor_info.y2 - monitor_info.x1;
```

Taking a look at it the al_get_monitor_info will always return false on android as the system interface get_monitor_info isn't implemented. I couldn't see a reason why it couldn't be so I added it. Fill me in  if there is a reason :)

Cheers

### DIFF
--- a/android/allegro_activity/src/AllegroActivity.java
+++ b/android/allegro_activity/src/AllegroActivity.java
@@ -10,7 +10,9 @@ import android.content.res.Configuration;
 import android.os.Bundle;
 import android.os.Environment;
 import android.os.Handler;
+import android.graphics.Rect;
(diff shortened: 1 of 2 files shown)


SiegeLord:
Merged as 5f92f99df77c9b63f1deb6a08cef7bff3c9b9cbb. Thanks a lot!

PR #526, opened 2015-11-26
A few BMP image loader fixes

tehsausage:


elias-pschernig:
Before this is commited, we should make sure to run all three BMP test suites over it so we at least don't get any regression over all those working tests. And ideally we would find another test suite which can test the cases in those fixes as well.

I wonder if we can wget those BMP files on Travis CI and have it auto-run them...

SiegeLord:
We can easily wget those bitmaps, or even check them in (well, maybe not... there seems to be quite a few of them, we should make a backup of them locally however just in case). Does the bmp suite cover the fixes in this patch? If not, we should also check in the 3 types of bitmaps fixed by this patch.

tehsausage:
Good call, It didn't occur to be but there is a regression with 16/24/32-bit bmp files that use an optional palette, since it has to be seeked past manually now.

One other test fails (g08offs.bmp) which seems to be the one depending on the bmp file header offset value being used. A decent work-around might be only seeking to the offset if it's beyond the current file position (i.e. pixel data not overlapping the bitmap header/metadata), which seems to be the behaviour used in MS Paint and Firefox, and the most likely to be compatible with the most files.

tehsausage:
Added the fixes above, all current bmp tests pass, but there's one file that loads as 100% transparent in allegro that shows up correctly elsewhere which I'd like to investigate now.

I've also added a test using bmpsuite v2 ( http://entropymine.com/jason/bmpsuite/ ). All of the "good" tests pass except one which uses an unsupported bitmask.

tehsausage:
Everything here is all done and ready.

Non-standard bitmasks are probably the last major hurdle to compatability, but I think they're big enough to warrant a second branch/request.

tehsausage:
Turned out to be less destructive than I expected. Bitfields support has been added in here too. It might be controversial as it's doing colour scaling without using conversion tables since it has to cover an arbitrary number of bits per colour, which might be a performance regression for people using the previously supported 555/565/888 formats.

All bmp tests are passing. Sorry for the spam. ;)

elias-pschernig:
Oh, I see you figured out how to run tests already - but in case anyone else starts from scratch, this is how to get the required files. I just googled for the three pictures in the dead link, you did the same I assume :) I suppose we should put this into a script and run on Travis and also run the bmp test suites.

# bmp suite 1
wget -N http://entropymine.com/jason/bmpsuite/bmpsuite.zip
unzip -u -d bmpsuite bmpsuite.zip

# bmp suite 2
wget -N http://downloads.sourceforge.net/project/bmptestsuite/bmptestsuite/bmptestsuite-0.9/bmptestsuite-0.9.zip
unzip -u bmptestsuite-0.9.zip

# bmp suite 3
wget -N -P wvnet http://circuitebre.cat/diplomes/www/test/images/bmp/test32bfv4.bmp
wget -N -P wvnet http://circuitebre.cat/diplomes/www/test/images/bmp/test32v5.bmp
wget -N -P wvnet http://circuitebre.cat/diplomes/www/test/images/bmp/trans.bmp

# bmp suite 4
wget -N http://entropymine.com/jason/bmpsuite/releases/bmpsuite-2.4.zip
unzip -u bmpsuite-2.4.zip
mv bmpsuite-2.4 bmpsuite2

All tests succeed before your patch, but yes, I can see that one completely transparent BMP. It also is transparent in GIMP and other image viewers, but certainly not supposed to, so yes, we should just change that test.

I can confirm that all tests succeed here with the patch. And I'm amazed that you went ahead and added a 4th test suite. Great job, thanks a lot! With this Allegro5's bitmap support is better than the one in GIMP and Firefox I suppose :)

addons/image/bmp.c
```diff
@@ -231,6 +234,38 @@ static void read_1bit_line(int length, ALLEGRO_FILE *f, unsigned char *buf)
 
 
 
+/* read_2bit_line:
+ *  Support function for reading the 2 bit bitmap file format.
+ */
+static void read_2bit_line(int length, ALLEGRO_FILE *f, unsigned char *buf)
+{
+   unsigned char b[16];
+   unsigned long n;
```
SiegeLord: Could you make this `int32_t`? I get paranoid about the `long` type...

SiegeLord:
Merged as eada181d19b58fc3984d84281a3d6ab13ae25304, thanks a lot! This is great work.

PR #529, opened 2015-12-02
Fix for finding GDI+ on 64-bit windows

peterhull90:
As discussed on the ML. This allows to build to complete on MSVC and 64-bit Windows. Tested on Visual Studio Community 2015 and Windows 10 64-bit.

### DIFF
--- a/cmake/FindGDIPLUS.cmake
+++ b/cmake/FindGDIPLUS.cmake
@@ -17,7 +17,12 @@ else()
     set(GDIPLUS_LOWERCASE 1 CACHE INTERNAL "Is GdiPlus.h spelt with lowercase?")
 endif()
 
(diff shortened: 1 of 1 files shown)


fatcerberus:
Oh good, so this wasn't just me.  I'm curious why it fails though... CMake bug?

peterhull90:
As far as I can see, CMake has a [set of rules](https://cmake.org/cmake/help/v3.3/command/find_library.html) as to where it looks for libraries when you use `find_library()`. Recent Windows SDKs have a different directory layout which CMake doesn't understand. Microsoft's linker *does* understand it so if you just specify `gdiplus.lib` the linker knows where to look.

SiegeLord:
Thanks! Merged as e289a7f950dca29191a1240bab5eb4fa326b6d77

PR #537, opened 2015-12-19
Turn off allegro_acodec dynamic loading by default

fatcerberus:
Dynamic loading is a useful feature in some situations, but has limited utility.  It almost never makes sense when building Allegro in-house since you can just compile out the codecs you don't need; the only time it really makes sense is if you're installing the library system-wide and want to leave it up to individual games to provide the audio codecs they need.

Therefore the most logical default setting for this option is off, leaving the user to turn it back on in the limited circumstance where it's necessary.

### DIFF
--- a/addons/acodec/CMakeLists.txt
+++ b/addons/acodec/CMakeLists.txt
@@ -7,7 +7,7 @@ endif()
 option(WANT_VORBIS "Enable Ogg Vorbis support using libvorbis" on)
 option(WANT_TREMOR "Enable Ogg Vorbis support using Tremor" off)
 option(WANT_MODAUDIO "Enable MOD Audio support" on)
-option(WANT_ACODEC_DYNAMIC_LOAD "Enable DLL loading in acodec (Windows)" on)
+option(WANT_ACODEC_DYNAMIC_LOAD "Enable DLL loading in acodec (Windows)" off)
 
 #-----------------------------------------------------------------------------#
 


trentg:
Not sure that I agree. The default settings are there to be used to get up and running quickly, while changing settings is always required in production builds.

fatcerberus:
That "up and running quickly" is the reason for the change - every time I've asked someone to compile Allegro from source, this setting being on by default has caused confusion, i.e. the user can't figure out why no audio can be loaded despite allegro_audio and allegro_acodec being built and the relevant codecs being found by CMake.

The main source of confusion in my experience has been that, if you just type `cmake` on the command line to build Allegro, it expects that libvorbis, libflac, etc. be present, but because they are late-bound, audio loading will fail silently on another system without these libraries on them (and sometimes even on the same system!), rather than allowing the OS to produce an error on startup.  The better default setting, IMO, is to make these dependencies explicit and allow the user to compile out the codecs they don't actually need.

trentg:
Right, there's nothing distinguishing one error from another so it could be a corrupt file or missing dynamic library... we have al_get_errno but most things including this don't use it. Your solution averts the problem, but really we should be setting errno. Every professional grade library uses something like it. Even quick simple code can print the error number which can easily be looked up. I think that is the root problem. I have no opinion on whether runtime library loading or compile time linking is better, both have their advantages. I guess it depends on whether you want to just bang it with a hammer so it works or fix the problem. In my code I like to do things the right way... I'm less and less involved in Allegro now so I won't complain if one of the regulars just makes the change but for reasons like this (little things not done right) Allegro lags behind things like SDL. Sure, we have more features that are easy to use, but at some point in ones programming lifetime features become easy and stability and correctness are more important. For this reason it's hard for me to even use Allegro anymore because the little things aren't done right. Pointless little features get added instead sometimes (stuff that you could code in 5 minutes by yourself.) If everything was done right it would be a no brainer, Allegro would hands down be the best low level library. Right now IMO that is not the case. Just some food for thought.

fatcerberus:
For what it's worth I chose to build an entire game engine (minisphere, a modernization of the Sphere RPG engine from the early '00s) on Allegro after having tried both SDL and SFML and not really liking either one, so I wouldn't count Allegro out just yet. :)

trentg:
I've built a lot of engines on Allegro, once you release something and start fixing bugs (I've fixed thousands of bugs in Allegro due to people reporting bugs in my games) you may change your mind.

trentg:
I'm not counting Allegro out. It has great potential. But I've used Allegro 5 for a half dozen commercial games... this isn't a hobby for me, my livelihood depends on it... I don't know if I can convey the magnitude of this but a bug in a released commercial game is an unfathomably huge deal... I've lost so much time and money on it. Up until recently I've been vigilant and fixes every issue I came across... this took many thousands of hours, inspecting hardware specific bugs, buying devices just to fix bugs, etc... it's been pure hell, let me tell you. Allegro is actually fairly stable now and the work I've done is a big factor. It works well for the most part, but there are some key inadequacies when it comes to professional commercial games, and it's stuff that is not easy to fix. If SDL is missing something that Allegro has, OK, I can code that in a few days or a week. The bugs I'd have to fix in Allegro (the stuff that's left that hasn't been fixed yet) is stuff that makes my head numb, that would surely take weeks or months to fix, very simple things like fullscreen modes on Linux and Mac that SDL does without a hitch. No, I'm not counting Allegro out, I'm saying these little stability things that it gets wrong are vastly more important than some of the dumb little features that get added (some recent things that have been added are good though.)

SiegeLord:
Dynamic loading for the acodec addon was always for the purpose of making you not have to include individual DLLs for the file formats you weren't using on Windows. Now that the official binaries link the dependencies statically, the original motivation is just not there anymore (practically, 99% of people will use the official binaries, I think).

For people that build Allegro themselves, this behavior is also unnecessary as if you are using dynamically linked dependencies and are concerned about including unnecessary DLLs, you can just compile Allegro without them as Bruce points out. Ultimately, it was only useful for the situation when you didn't build Allegro yourself.

Back when I used to build Allegro myself, I found the dynamic linking to be more harm than good, as the dependency walker wouldn't uncover these dependencies and I'd send out my game without all the necessary DLLs.

To that end, I think it's fine to disable this by default. I probably wouldn't remove it outright, as perhaps it's still useful in certain situations.

fatcerberus:
By the way, AppVeyor build is failing because it can't find the freetype libraries, not due to anything in this pull (which is just a single one-line change to the CMake script).

SiegeLord:
Yeah, that's known. We're almost done creating Nuget packages for Allegro and dependencies, so that compile test will start passing (and testing more things) as soon as that's done.

SiegeLord:
Merged as 2ba0f3e4c8ea43f23eb2e954d9e738d2106d5672.

beoran:
@trentg 

Well, I agree with the feeling, I guess, although Allegro has always been more fully-featured than SDL. I think apart from back ends and drivers ,Allegro 5 almost has all the needed features. Once feature complete, we should focus 100% on stability and improving the code quality.

PR #538, opened 2015-12-20
Fix al_set_new_window_title() off-by-1

fatcerberus:
The last character was cut off of whatever title was set due to an off-by-1 error in the `strncpy`. I also made it const-correct.


SiegeLord:
Huh, `strncpy` is quite a confusing function to use. Thanks!

PR #540, opened 2015-12-20
fix some android crashes

elias-pschernig:
This change seems to fix quite a few crashes I got on various Android
devices on Amazon AWS device farm (unfortunately my 250 free minutes are
used up so can't do any more testing).

### DIFF
--- a/android/allegro_activity/src/AllegroActivity.java
+++ b/android/allegro_activity/src/AllegroActivity.java
@@ -131,6 +131,7 @@ public class AllegroActivity extends Activity
    {
       try {
          if (surface != null) {
+            // TODO: How can this ever happen?
             Log.w("AllegroActivity", "not recreating surface");
             return;
          }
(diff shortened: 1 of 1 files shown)


elias-pschernig:
Just putting this up here, it's certainly not a clean fix as I can't even figure out how it can happen - the surface gets set to null when it is destroyed... but some race condition makes Allegro try and create it a second time. By simply ignoring the second time (it seems to always happen within the first second of starting the game) my game works instead of crashing. So if nobody sees an obvious problem with it I'd like to apply anyway, it does make Allegro work on more Android devices after all - and hopefully at some point someone will clean up the surface/display handling on the Android port and actually fix the issue. I can't follow the jungle of all the undocumented d->first_run, d->created, d->recreated, d->is_destroy_display... fields in ANDROID_DISPLAY at all :P

SiegeLord:
Maybe there should be a comment along the lines of what you said above, so the future person knows that this isn't a proper fix :smile:.

elias-pschernig:
Indeed, I added a comment.

elias-pschernig:
Fixed this with commit c3493b instead I believe.

PR #541, opened 2015-12-21
Make OSX port work with older compilers (clang 3.1).

elias-pschernig:
I noticed that my regression tests stopped working for OSX at some point, and apparently the reason is the introduction of the @[] dictionary syntax. I don't know which XCode version officially is the oldest we support, but this patch would make Allegro compile with clang 3.1 (from ca. around 2012) again. I did not test on a newer OSX so not going to commit.


SiegeLord:
Hey @peterhull90, would you mind testing this if you have time? I know you did some recent OSX related changes.

peterhull90:
Will have a look later today. Those objective-C literals *should* work from Clang 3.1 onwards.[(ref)](http://clang.llvm.org/docs/ObjectiveCLiterals.html)

peterhull90:
Hmm, got an error

> 2015-12-21 20:34:22.752 ex_menu[16733:699154] -[ALWindow copyWithZone:]: unrecognized selector sent to instance 0x7f9d2481e040

will need to check a bit more. Sorry not got time today, will get onto it as soon as I can.

Why is the Appveyor build failing btw?

fatcerberus:
AppVeyor is failing due to an ongoing transition to Nuget packages for the MSVC build, apparently.

peterhull90:
Must have got some stale libraries mixed in somewhere; copyWithZone isn't mentioned anywhere in the current code base. Will continue to investigate.

peterhull90:
No, this isn't right. It's dictionaryWithObjectsAndKeys - object first then key.

    NSDictionary* newentry = [NSDictionary dictionaryWithObjectsAndKeys: menu, @"menu", window,@"window", nil];

elias-pschernig:
Um, since I rebased my branch to master instead of 5.1, the PR looks a bit strange now :P I'll make a new one against master.

PR #547, opened 2016-01-01
Fix the initial display size not being correct in tiling wms.

### LINKED ISSUE #542: Window size problems with tiling window manager
When a window is intially opened at a certain size using [dwm](dwm.suckless.org), the program doesn't seem to be informed of the window size being changed, resulting in this (window is 1200x1066):

ex_resize2

Resizing or moving the window will cause it to update.

Oddly enough, the problem is worse running a Windows program through Wine:

eomap

The program remains stuck in the corner even when moving and resizing the window as a tile, but updates correctly when using alt+right mouse to resize the window gradually as a floating window.

SiegeLord:
It appears that those wms send resize events that are lost due to the event
queue not being connected to the display yet. Fix it by acknowledging the
resize if such resizes were detected.

Fixes #542.


SiegeLord:
Merged as e58caffea3a3e4249997e42f5a28db5924c63fee.

PR #551, opened 2016-01-03
Reduce the MSVC warning level to 3 and fix all the 32 bit errors.

SiegeLord:
Level 3 is what's recommended by MS.


fatcerberus:
I was going to suggest this a while back, /W4 is way too picky and warns for stupid stuff.  Good that the 32-bit narrowing warnings are fixed, though.

SiegeLord:
I mean, fix warnings for the 32 bit build. There's a ton of narrowing warnings that show up when you build the 64 bit build. That's for a future commit D:.

SiegeLord:
Merged as b9aa6da1156ba6c0e7d160418f6add1673e30602

PR #553, opened 2016-01-03
Don't link the MSVC C runtime at all when requesting a static runtime…

SiegeLord:
… and building a static library.

This mirrors what happens with GCC.

### DIFF
--- a/cmake/Common.cmake
+++ b/cmake/Common.cmake
@@ -95,6 +95,12 @@ function(add_our_library target framework_name sources extra_flags link_with)
             else()
                 set(extra_flags "${extra_flags} /MT")
             endif()
+            if (NOT BUILD_SHARED_LIBS)
+                # /Zl instructs MSVC to not mention the CRT used at all,
+                # allowing the static libraries to be combined with any CRT version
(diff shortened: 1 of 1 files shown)


fatcerberus:
I didn't even know this was possible, I thought MSVC just automatically links it in regardless, hence all the linker drama with static/dynamic runtimes.

SiegeLord:
I didn't know it was possible until recently as well, but then I found out about it from posts such as this: http://stackoverflow.com/questions/3764716/c-visual-studio-libraries/3764884#3764884

I also tested this, and indeed I was able to vary the final runtime choice of the program independently of the `/MT` switch used during compilation of Allegro.

fatcerberus:
Perhaps we should do the same for Allegro's dependencies (libogg, DUMB, etc.).

SiegeLord:
That's done in https://github.com/SiegeLord/allegro_winpkg/commit/49cc4d3fb19a5a983e51d70b10618c577b7b47b5.

SiegeLord:
Merged as dab3113b94f8edc868c01f544728bd698f693a0f

PR #556, opened 2016-01-11
Fix debug lib crashing on start

BorisCarvajal:
A logging function was called before getting a system driver,
then al_get_time() pointed to nothing (at least on Windows).

### DIFF
--- a/src/file_stdio.c
+++ b/src/file_stdio.c
@@ -38,7 +38,6 @@
 #include <sys/stat.h>
 #endif
 
-ALLEGRO_DEBUG_CHANNEL("stdio")
 
 /* forward declaration */
 const struct ALLEGRO_FILE_INTERFACE _al_file_interface_stdio;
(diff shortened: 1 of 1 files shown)


beoran:
Sorry but your suggested patch doesn't realy fix the problem, just circumvents it.  Probably, al_get_time() and anything related to logging should be changed so that if the system driver isn't installed it doesn't crash but returns an appropriate value, perhaps -1 in this case.

fatcerberus:
To be fair, if Allegro is calling a function internally before that is allowed by the design (as in the case of this logging call), that's absolutely a bug, even if the function in question is changed to silently fail instead of crashing.

I'll agree the patch itself is questionable though.

BorisCarvajal:
Well this one random debug message just commited last month (fbe7ea6bf04e7063efb628274604c647e796d5fc) broke all the apps using the libreary in debug mode.

SiegeLord:
There've been a few changes to make this logging location actually work (e.g. c14c5adac60b318d07074d3132c306f6605a3ce2), I wasn't aware of it breaking windows though (I'm surprised I haven't hit it!). I'll investigate the `al_get_time` issue.

SiegeLord:
With b23aedab05aca909d618cfb0af85afa0482f65ae, there is no more crash at least for me. Thanks for bring this to our attention!

beoran:
Yes, that looks better. :)

Issue #495, opened 2015-09-08
TTF re-locking causes missing glyphs


goobliata:
This seems to be a very device specific problem, looking for leads on what's causing it. Whenever the page bitmap is re-locked, the first glyph drawn shows up blank on Samsung S6. I know this because I tried calling al_get_text_width on the entire translation and only the first character on a new row other than the first row is blank, which is how the locking works. I tried different sized fonts which results in different first characters on rows and those characters are also blank. When not calling al_get_text_width like this, many more characters are blank which I assume must be the first characters in each lock (but I don't have a good way to test this as I don't have the device in my possession.) This problem doesn't affect all Androids, right now the only one I know of is the S6. 
Potential suspects: ttf addon, OpenGL ES locking code, ???.

Attached the output from debug_cache in TTF addon using the same font size as used on S6. The first characters in the second, third and fourth rows are blank (O, 2, 0) on S6. (background has been coloured black so you can see the glyphs on the white webpage background...)

![font](https://cloud.githubusercontent.com/assets/766383/9747602/d5ea3be0-563b-11e5-89e7-957e7bd3b6db.png)

trentg:
A version of the game compiled with the font addon changes up to and including August 15, 2013 works fine.

trentg:
Got to thinking this may be a driver bug, so Googled to see if anyone has found that glTexSubImage2D followed by glFlush fixes any problems. Found the exact same issue with Cocos2D:

https://github.com/cocos2d/cocos2d-x/issues/9499

What I'm going to attempt is a single line patch that will try to recache any glyph if the locking fails initially... all it does is set glyph->page_bitmap to NULL before returning NULL from alloc_glyph_region. I'm hoping this fixes it...

trentg:
If not perhaps calling glFlush will fix it but that's a pretty horrible solution.

trentg:
Actually the solution I proposed doesn't make sense, since the locking itself just mallocs a buffer. Hopefully glFlush fixes it.

trentg:
One potential way to fix this would be to use accelerated drawing (al_draw_pixel) to draw the characters instead of locking.

trentg:
If anyone comes across this, here's how I've fixed it in my game... First you need to patch Allegro like this:

https://gist.github.com/trentg/4c79372c1da4f45b349b

Next, I load all translations (every language) and concatenate them all into one string, then call al_get_text_width on that string. Horrible. Ugly. But it works.

[...]

PR #562, opened 2016-01-26
Add a configuration switch to enable whole-page locking in the ttf addon.

SiegeLord:
It is significantly slower (7x+ on my system), but is likely to be less buggy
on some drivers.

This hopefully works around issue #495.

### DIFF
--- a/addons/ttf/ttf.c
+++ b/addons/ttf/ttf.c
@@ -83,6 +83,7 @@ typedef struct ALLEGRO_TTF_FONT_DATA
 
    int min_page_size;
    int max_page_size;
+   bool lock_whole_page;
 } ALLEGRO_TTF_FONT_DATA;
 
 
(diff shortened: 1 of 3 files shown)


trentg:
I think the unlock_page in ttf_get_glyph_advance will prevent this from working. The way I quickly tested my patch was to just put a printf in unlock_page and see how many unlocks there were. There should be exactly the same as the number of pages. I didn't test this patch though.

SiegeLord:
Yes, you're right. I need to think about it more what the expected behavior should be. Maybe you're right, and indeed adding an explicit caching function is best here rather than relying on `get_length` to never unlock the page (something which may break due to future code changes...)

It appears the current code never actually uses the `lock_more` mechanic, because the only time it is passed as `true` it is immediately unlocked...

Let me try something different.

PR #563, opened 2016-01-30
Rework TTF locking a bit.

SiegeLord:
- Change lock_more logic to be slightly simpler.
- Add a configuration option to pre-cache a set of user specified glyphs. Off by default (should we turn it on by default?)
- Add a configuration option to not cache glyphs on the fly (this is to work around some locking bugs)
- Fix one spot where the fallback font wasn't used.

This is an alternative to #562 which should actually work.

### DIFF
--- a/addons/ttf/ttf.c
+++ b/addons/ttf/ttf.c
@@ -70,7 +70,6 @@ typedef struct ALLEGRO_TTF_FONT_DATA
    int page_pos_x;
    int page_pos_y;
    int page_line_height;
-   REGION lock_rect;
    ALLEGRO_LOCKED_REGION *page_lr;
 
    FT_StreamRec stream;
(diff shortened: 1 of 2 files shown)


SiegeLord:
As I was thinking about this today, it occurs to me that it might be better to specify character ranges rather than a literal string (e.g. `cache_ranges = 32-127,512,540-550`).

trentg:
For me it's easier to cache characters. I already have the translation, so I just find the unique characters in it. With ranges I have to go look up Unicode ranges and what if I miss something? Also if I want to load only one language at a time to save memory, I have to keep an array of ranges for each language.

trentg:
Though I suppose I can just make a list of single character values separated by commas.

SiegeLord:
Yeah, you're right. It seems like the practical use case for this would indeed be to just go through your game text and just output all the unique characters.

SiegeLord:
Alright, I think I'll merge this as is. We can always play around with the API if necessary. Hopefully this bug will just go away with time.

PR #569, opened 2016-02-14
Unstable API experiment

SiegeLord:
This illustrates the ideas I had for combining unstable and stable API in a stable release. For illustration purposes I made the audio recording API unstable (since it's the only thing that came to mind as being obviously half-baked), but this doesn't necessarily mean that it truly has to be marked as such for the 5.2 release. The general idea is as follows:

From the user's point of view, the unstable API will not be declared by Allegro headers unless they define `ALLEGRO_UNSTABLE` macro. When that happens, in addition to revealing the unstable API, the version compatibility check by `al_install_system` will get more stringent in that the wip versions will have to match (i.e. 5.2.0 is only compatible with 5.2.0 if you're using the unstable API).

From our point of view, we will wrap the unstable API with this:

```c
#if defined(ALLEGRO_UNSTABLE) || defined(ALLEGRO_INTERNAL_UNSTABLE) || defined(ALLEGRO_*_SRC)
```

The first term is obvious. `ALLEGRO_INTERNAL_UNSTABLE` is for the case when you need to use unstable API between addons. It's there to not accidentally set the unstable bit (see below). The third term is so that when Allegro is compiled, all API is exported for shared libraries.

On the documentation side of things, I propose adding a little note explaining that an API is unstable and a short explanation as to why.

<s>The aforementioned more stringent version check is implemented as follows. When `ALLEGRO_UNSTABLE` is defined, Allegro's minor version goes from even to the next odd version (i.e. very similar to how things are right now). I.e. if you get a 5.2.0 shared library, but declare `ALLEGRO_UNSTABLE`, your library version will actually be 5.3.0.</s>

<s>I did it this way for a few reasons. First, `al_install_system` only accepts a single integer, and it documents it to be `ALLEGRO_VERSION_INT`. We also document the construction of `ALLEGRO_VERSION_INT`, so we can't really change it (e.g. by adding an unstable bit or something). Bumping the sub version by 1 seemed like an easy way to do it. I admit, it's a bit of a hack.</s>

<s>Since that the above behaviour is a change from what is done now I also bumped the version to 5.2.0.</s>

The aforementioned more stringent version check is implemented as follows. When `ALLEGRO_UNSTABLE` is defined, the MSB of ALLEGRO_VERSION_INT is set. When passed to `al_install_system`, it checks for that bit, and then performs the more stringent check accordingly.

What do you think about all this?

### DIFF
--- a/addons/acodec/ogg.c
+++ b/addons/acodec/ogg.c
@@ -4,6 +4,7 @@
  * author: Ryan Dickie (c) 2008
  */
 
+
 #include "allegro5/allegro.h"
 #include "allegro5/allegro_acodec.h"
 #include "allegro5/allegro_audio.h"
--- a/addons/audio/allegro5/allegro_audio.h
+++ b/addons/audio/allegro5/allegro_audio.h
@@ -38,7 +38,6 @@ extern "C" {
    #define ALLEGRO_KCM_AUDIO_FUNC      AL_FUNC
 #endif
 
-
 /* Internal, used to communicate with acodec. */
 /* Must be in 512 <= n < 1024 */
 #define _KCM_STREAM_FEEDER_QUIT_EVENT_TYPE   (512)
@@ -53,6 +52,7 @@ extern "C" {
 
 #define ALLEGRO_EVENT_AUDIO_RECORDER_FRAGMENT       (515)
 
+#if defined(ALLEGRO_UNSTABLE) || defined(ALLEGRO_INTERNAL_UNSTABLE) || defined(ALLEGRO_KCM_AUDIO_SRC)
 /* Type: ALLEGRO_AUDIO_RECORDER_EVENT
  */
 typedef struct ALLEGRO_AUDIO_RECORDER_EVENT ALLEGRO_AUDIO_RECORDER_EVENT;
@@ -63,6 +63,7 @@ struct ALLEGRO_AUDIO_RECORDER_EVENT
    void *buffer;
    unsigned int samples;
 };
+#endif
 
 
 /* Enum: ALLEGRO_AUDIO_DEPTH
@@ -173,9 +174,11 @@ typedef struct ALLEGRO_MIXER ALLEGRO_MIXER;
 typedef struct ALLEGRO_VOICE ALLEGRO_VOICE;
 
 
+#if defined(ALLEGRO_UNSTABLE) || defined(ALLEGRO_INTERNAL_UNSTABLE) || defined(ALLEGRO_KCM_AUDIO_SRC)
 /* Type: ALLEGRO_AUDIO_RECORDER
  */
 typedef struct ALLEGRO_AUDIO_RECORDER ALLEGRO_AUDIO_RECORDER;
+#endif
 
 
 #ifndef __cplusplus
@@ -386,6 +389,9 @@ ALLEGRO_KCM_AUDIO_FUNC(bool, al_save_sample_f, (ALLEGRO_FILE* fp, const char *id
 ALLEGRO_KCM_AUDIO_FUNC(ALLEGRO_AUDIO_STREAM *, al_load_audio_stream_f, (ALLEGRO_FILE* fp, const char *ident,
 	size_t buffer_count, unsigned int samples));
 
+
+#if defined(ALLEGRO_UNSTABLE) || defined(ALLEGRO_INTERNAL_UNSTABLE) || defined(ALLEGRO_KCM_AUDIO_SRC)
+
 /* Recording functions */
 ALLEGRO_KCM_AUDIO_FUNC(ALLEGRO_AUDIO_RECORDER *, al_create_audio_recorder, (size_t fragment_count,
    unsigned int samples, unsigned int freq, ALLEGRO_AUDIO_DEPTH depth, ALLEGRO_CHANNEL_CONF chan_conf));
@@ -396,6 +402,8 @@ ALLEGRO_KCM_AUDIO_FUNC(ALLEGRO_EVENT_SOURCE *, al_get_audio_recorder_event_sourc
    (ALLEGRO_AUDIO_RECORDER *r));
 ALLEGRO_KCM_AUDIO_FUNC(ALLEGRO_AUDIO_RECORDER_EVENT *, al_get_audio_recorder_event, (ALLEGRO_EVENT *event));
 ALLEGRO_KCM_AUDIO_FUNC(void, al_destroy_audio_recorder, (ALLEGRO_AUDIO_RECORDER *r));
+
+#endif
    
 #ifdef __cplusplus
 } /* End extern "C" */
--- a/addons/audio/allegro5/internal/aintern_audio.h
+++ b/addons/audio/allegro5/internal/aintern_audio.h
@@ -9,6 +9,42 @@
 #include "allegro5/internal/aintern_vector.h"
 #include "../allegro_audio.h"
 
+struct ALLEGRO_AUDIO_RECORDER {
+  ALLEGRO_EVENT_SOURCE source;
+  
+  ALLEGRO_THREAD           *thread;
+  ALLEGRO_MUTEX            *mutex;
+  ALLEGRO_COND             *cond;
+                           /* recording is done in its own thread as
+                              implemented by the driver */
+  
+  ALLEGRO_AUDIO_DEPTH      depth;
+  ALLEGRO_CHANNEL_CONF     chan_conf;
+  unsigned int             frequency;
+
+  void                     **fragments;
+                           /* the buffers to record into */
+
+  unsigned int             fragment_count;
+                           /* the number of fragments */
+
+  unsigned int             samples;
+                           /* the number of samples returned at every FRAGMENT event */
+                           
+  size_t                   fragment_size;
+                           /* size in bytes of each fragument */
+
+  unsigned int             sample_size;
+                           /* the size in bytes of each sample */
+  
+  volatile bool            is_recording; 
+                           /* true if the driver should actively be updating
+                              the buffer */
+                              
+  void                     *extra;
+                           /* custom data for the driver to use as needed */
+};
+
 typedef enum ALLEGRO_AUDIO_DRIVER_ENUM
 {
    /* Various driver modes. */
@@ -45,8 +81,8 @@ struct ALLEGRO_AUDIO_DRIVER {
    int            (*set_voice_position)(ALLEGRO_VOICE*, unsigned int);
    
    
-   int            (*allocate_recorder)(ALLEGRO_AUDIO_RECORDER *);
-   void           (*deallocate_recorder)(ALLEGRO_AUDIO_RECORDER *);
+   int            (*allocate_recorder)(struct ALLEGRO_AUDIO_RECORDER *);
+   void           (*deallocate_recorder)(struct ALLEGRO_AUDIO_RECORDER *);
 };
 
 extern ALLEGRO_AUDIO_DRIVER *_al_kcm_driver;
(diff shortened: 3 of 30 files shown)


elias-pschernig:
I don't really understand the "5.2.0 is 5.3.0 when unstable" bit, but I suppose that's just an internal implementation detail? I.e. our next Allegro versions will be: 5.2.0, 5.2.1, 5.2.2, 5.3.0, 5.3.1, 5.3.2, ... that is, we won't actually skip 5.3.0?

And then if you do not have ALLEGRO_UNSTABLE defined, when does something get stable? Like:

5.2.12 audio-rec=unstable
5.2.13 audio-rec=stable
5.2.14 audio-rec = stable
5.3.0 audio-rev = stable

If I compile with ALLEGRO_UNSTABLE, I suppose all 4 versions can use audio recording. But if I do not define it, can I use it in 5.2.14? Or only in 5.3.0?

trentg:
I think the point is to avoid the necessity of having odd versioned unstable branches. Instead, you just have the macro protecting stuff that might change. It means we don't have to backport to latest stable which is currently impossible for 5.0. That's my interpretation.

I have no qualms about this. The only issue is with Linux where Allegro is in the repos... So if a game uses unstable, bumping Allegro breaks it. But if you want your game in repos, you can still use stable just like now.

SiegeLord:
Right, that's the idea. The only releases we'd ever make would be the even versions, and odd versions would be something you select via a macro. Note that this isn't a requirement of any of this, it just seemed the easiest option to take such that we don't have to change what `ALLEGRO_VERSION_INT` means. That said, looking again at what is documented, I note that it's not actually documented what `ALLEGRO_VERSION_INT` is. The only thing that's documented is the return value of `al_get_allegro_version` and that you need to pass `ALLEGRO_VERSION_INT` to `al_install_system`.

So maybe what we could do is set the msb for unstable version in `ALLEGRO_VERSION_INT`, but remove that bit from `al_get_allegro_version()`. Then none of this hackery with randomly increasing the sub version will need to be done. Let me think about it a bit.

> And then if you do not have ALLEGRO_UNSTABLE defined, when does something get stable? Like:
> 
> 5.2.12 audio-rec=unstable
> 5.2.13 audio-rec=stable
> 5.2.14 audio-rec = stable
> 5.3.0 audio-rec = stable
> 
> If I compile with ALLEGRO_UNSTABLE, I suppose all 4 versions can use audio recording. But if I do not define it, can I use it in 5.2.14? Or only in 5.3.0?

It'd be available in 5.2.14.

SiegeLord:
Ok, I changed it to using the MSB instead. Now there's no confusion about the version being randomly bumped or not, there's always exactly one version.

fatcerberus:
I like this, it's similar to how a few SemVer-versioned packages I've seen work, there's a core "stable" API, but new features are phased in as "experimental", like deprecation but in reverse.  It might allow for a more rapid development cycle, too.

SiegeLord:
I made some more API unstable. Again, these are mostly proposals, I'm open to doing whatever:

- `al_toggle_menu_item_flags` because it seems redundant given that there's get/set versions of that API. We also removed a similar API from display flags at some point.

- `al_convert_bitmaps` because it's name is just confusing. You might think that it does something similar to `al_convert_bitmap` (without the s), but it does not. It probably just needs a rename.

- The mouse emulation API for touch events. I think it's of very limited usefulness, as touch input is typically very different from non-touch input (e.g. you typically can't hover a touch like you can do with the mouse), and just sending the same events without changing the code will often be wrong. It doesn't seem that hard to just handle the touch events manually and do whatever the appropriate thing seems right for your game.

I haven't made it unstable, but the entire haptic API seems very large and complex,and it's liable we didn't get it right, so maybe we should be conservative about it too.

Here's all of the new API added in 5.1, incidentally. We're almost at 1000 API entries :astonished:.:

```
al_is_touch_input_installed
al_install_touch_input
al_uninstall_touch_input
al_get_touch_input_state
al_set_mouse_emulation_mode
al_get_mouse_emulation_mode
al_get_touch_input_event_source
al_get_touch_input_mouse_emulation_event_source
al_lock_bitmap_blocked
al_lock_bitmap_region_blocked
al_for_each_fs_entry
al_remove_config_section
al_remove_config_key
al_set_new_window_title
al_get_new_window_title
al_set_blend_color
al_get_blend_color
al_get_cpu_count
al_get_ram_size
al_ferrmsg
al_vfprintf
al_fprintf
al_get_clipboard_text
al_set_clipboard_text
al_clipboard_has_text
al_register_bitmap_identifier
al_load_bitmap_flags
al_load_bitmap_flags_f
al_identify_bitmap_f
al_identify_bitmap
al_set_display_option
al_get_display_orientation
al_set_window_constraints
al_get_window_constraints
al_acknowledge_drawing_halt
al_acknowledge_drawing_resume
al_set_render_state
al_use_projection_transform
al_get_current_projection_transform
al_get_current_inverse_transform
al_build_camera_transform
al_translate_transform_3d
al_scale_transform_3d
al_transform_coordinates_3d
al_orthographic_transform
al_rotate_transform_3d
al_perspective_transform
al_horizontal_shear_transform
al_vertical_shear_transform
al_resume_timer
al_get_opengl_program_object
al_get_pixel_block_size
al_get_pixel_block_width
al_get_pixel_block_height
al_premul_rgba
al_premul_rgba_f
al_is_event_source_registered
al_pause_event_queue
al_is_event_queue_paused
al_reparent_bitmap
al_get_bitmap_x
al_get_bitmap_y
al_android_set_apk_file_interface
al_android_set_apk_fs_interface
al_android_get_os_version
al_register_trace_handler
al_install_haptic
al_uninstall_haptic
al_is_haptic_installed
al_is_joystick_haptic
al_is_mouse_haptic
al_is_keyboard_haptic
al_is_display_haptic
al_is_touch_input_haptic
al_get_haptic_from_joystick
al_get_haptic_from_mouse
al_get_haptic_from_keyboard
al_get_haptic_from_display
al_get_haptic_from_touch_input
al_is_haptic_active
al_get_haptic_capabilities
al_is_haptic_capable
al_get_haptic_gain
al_set_haptic_gain
al_get_haptic_autocenter
al_set_haptic_autocenter
al_get_max_haptic_effects
al_is_haptic_effect_ok
al_upload_haptic_effect
al_play_haptic_effect
al_upload_and_play_haptic_effect
al_stop_haptic_effect
al_is_haptic_effect_playing
al_get_haptic_effect_duration
al_rumble_haptic
al_release_haptic_effect
al_release_haptic
al_convert_bitmap
al_convert_bitmaps
al_set_mouse_wheel_precision
al_get_mouse_wheel_precision
al_win_add_window_callback
al_win_remove_window_callback
al_clear_depth_buffer
al_create_shader
al_attach_shader_source
al_attach_shader_source_file
al_build_shader
al_get_shader_log
al_get_shader_platform
al_use_shader
al_destroy_shader
al_set_shader_sampler
al_set_shader_matrix
al_set_shader_int
al_set_shader_float
al_set_shader_int_vector
al_set_shader_float_vector
al_set_shader_bool
al_get_default_shader_source
al_get_x_window_id
al_iphone_get_window
al_iphone_get_view
al_iphone_set_statusbar_orientation
al_get_d3d_texture_size
al_set_d3d_device_release_callback
al_set_d3d_device_restore_callback
ALLEGRO_FOR_EACH_FS_ENTRY_RESULT
ALLEGRO_NEW_WINDOW_TITLE_MAX_SIZE
ALLEGRO_TOUCH_INPUT_MAX_TOUCH_COUNT
ALLEGRO_TOUCH_INPUT
ALLEGRO_TOUCH_INPUT_STATE
ALLEGRO_TOUCH_STATE
ALLEGRO_MOUSE_EMULATION_MODE
ALLEGRO_RENDER_STATE
ALLEGRO_RENDER_FUNCTION
ALLEGRO_WRITE_MASK_FLAGS
ALLEGRO_HAPTIC_CONSTANTS
ALLEGRO_HAPTIC
ALLEGRO_HAPTIC_EFFECT
ALLEGRO_HAPTIC_EFFECT_ID
ALLEGRO_SHADER
ALLEGRO_SHADER_TYPE
ALLEGRO_SHADER_PLATFORM
al_get_font_ranges
al_draw_glyph
al_get_glyph_width
al_get_glyph_dimensions
al_get_glyph_advance
al_do_multiline_ustr
al_do_multiline_text
al_draw_multiline_ustr
al_draw_multiline_text
al_draw_multiline_textf
al_set_fallback_font
al_get_fallback_font
al_load_bitmap_font_flags
ALLEGRO_VIDEO_EVENT_TYPE
ALLEGRO_VIDEO_POSITION_TYPE
al_open_video
al_close_video
al_get_video_event_source
al_start_video
al_start_video_with_voice
al_set_video_playing
al_is_video_playing
al_get_video_frame
al_get_video_position
al_seek_video
al_get_video_audio_rate
al_get_video_fps
al_get_video_scaled_width
al_get_video_scaled_height
al_init_video_addon
al_shutdown_video_addon
al_get_allegro_video_version
ALLEGRO_AUDIO_RECORDER_EVENT
ALLEGRO_AUDIO_RECORDER
al_get_default_voice
al_set_default_voice
al_fill_silence
al_get_audio_stream_played_samples
al_create_audio_recorder
al_start_audio_recorder
al_stop_audio_recorder
al_is_audio_recorder_recording
al_get_audio_recorder_event
al_get_audio_recorder_event_source
al_destroy_audio_recorder
al_draw_polygon
al_draw_filled_polygon
al_draw_filled_polygon_with_holes
ALLEGRO_LINE_JOIN
ALLEGRO_LINE_CAP
ALLEGRO_PRIM_BUFFER_FLAGS
ALLEGRO_VERTEX_BUFFER
ALLEGRO_INDEX_BUFFER
al_draw_polyline
al_create_vertex_buffer
al_create_index_buffer
al_destroy_vertex_buffer
al_destroy_index_buffer
al_lock_vertex_buffer
al_lock_index_buffer
al_unlock_vertex_buffer
al_unlock_index_buffer
al_draw_vertex_buffer
al_draw_indexed_buffer
al_get_vertex_buffer_size
al_get_index_buffer_size
al_triangulate_polygon
ALLEGRO_MENU
ALLEGRO_MENU_INFO
al_create_menu
al_create_popup_menu
al_clone_menu
al_clone_menu_for_popup
al_build_menu
al_append_menu_item
al_insert_menu_item
al_remove_menu_item
al_find_menu
al_find_menu_item
al_get_menu_item_caption
al_set_menu_item_caption
al_get_menu_item_flags
al_set_menu_item_flags
al_toggle_menu_item_flags
al_get_menu_item_icon
al_set_menu_item_icon
al_destroy_menu
al_get_default_menu_event_source
al_enable_menu_event_source
al_disable_menu_event_source
al_get_display_menu
al_set_display_menu
al_popup_menu
al_remove_display_menu
```

beoran:
Although I wrote it , I wouldn't be against marking the haptic API as unstable. I'd like to get some feedback from people who are actually using it. Although it mirrors the rich API of DIrectX and of the Linux kernel, most joysticks these days just seem to have very simple vibration functionality with 2 rotors, small and large, and handheld devices seem to have only a single vibration motor. I can't afford more advanced controllers like steering wheels or flight joysticks so I have no way to test the extend functionality offered by the current API. 

Arguably, for all these reasons, the API could be reduced to a simpler one that just allows you to check how many vibration motors there are, and set an intensity for them.

SiegeLord:
Alright, made the haptic API unstable in its entirety. Again, I don't think that it's a bad API or that it's overly complex for the task (as the task is complex), I just thing Allegro got burned badly by the joystick API being complex and completely inadequate, and us being stuck with it, so I'd rather not make the same mistake again.

fatcerberus:
What's wrong with the joystick API?  I use it in minisphere, I've never had any complaints about it.  I found it much simpler and more intuitive than using, e.g. DirectInput directly (which I've actually done, once upon a time).

SiegeLord:
I filed an issue about it: #579. In short, it seems impossible to use without having a configuration screen as the axes/buttons etc are different between different platforms and different joysticks. The proposed solution was to do what SDL does, i.e. have a table of known joysticks. Unfortunately, due to the differences between platforms, it's seems difficult to solve without breaking existing code.

beoran:
Maybe we need a new joystick API as well, then? Anyway, thanks for marking haptics as unstable.

SiegeLord:
Alright, I'm going to commit this. Last call for comments!

PR #571, opened 2016-02-17
is_event_source_registered()

koro-xx:
Currently there is no way of knowing which event sources are registered to an event queue. 
Is this acceptable?

### DIFF
--- a/docs/src/refman/events.txt
+++ b/docs/src/refman/events.txt
@@ -676,6 +676,8 @@ Return true if the event source is registered.
 
 See also: [al_register_event_source]
 
+Since: 5.1.14
+
 ## API: al_pause_event_queue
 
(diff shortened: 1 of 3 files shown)


docs/src/refman/events.txt
```diff
@@ -670,6 +670,12 @@ source, they will no longer be in the queue after this call.
 
 See also: [al_register_event_source]
 
+## API: al_is_event_source_registered
+
+Return true if the event source is registered.
+
+See also: [al_register_event_source]
+
```
SiegeLord: Could you add a 'Since: 5.1.14' here?

SiegeLord:
This looks great!

SiegeLord:
Committed as 441572579f0a609746c6cafe1a2f2fbd95ac2c11. Thanks!

PR #572, opened 2016-02-19
Change vsync to moving vertical bar instead of flashing screen.

Sevalecan:


SiegeLord:
This is the best change ever. The old flashing was the most painful thing I've ever witnessed.

Committed as 695b705fd1f8617a9740cbaefbeea03df4800a95.

PR #589, opened 2016-03-13
Make toggling ALLEGRO_FRAMELESS work on OSX, with a few caveats.

### LINKED ISSUE #582: Toggling ALLEGRO_FRAMELESS doesn't work on OSX
Tested with 10.8 and 10.10

SiegeLord:
The reason it didn't work before was because toggling NSBorderlessWindowMask
does nothing, as it is defined to be 0 (i.e. nothing that uses it as a flag
works, so I fixed code that did). Using the correct flag however has some
issues, as described in the code. Still, this is better than not working at
all.

Fixes #582.

### DIFF
--- a/src/macosx/osxgl.m
+++ b/src/macosx/osxgl.m
@@ -1909,12 +1909,16 @@ static bool resize_display_win(ALLEGRO_DISPLAY *d, int w, int h)
    NSWindow* window = dpy->win;
    NSRect current = [window frame];
    NSRect content = NSMakeRect(0.0f, 0.0f, (float) w, (float) h);
+   float scale_factor = 1.0;
 #if MAC_OS_X_VERSION_MIN_REQUIRED >= 1070
    content = [window convertRectFromBacking: content];
+   if ([window respondsToSelector:@selector(backingScaleFactor)]) {
+      scale_factor = [window backingScaleFactor];
+   }
 #endif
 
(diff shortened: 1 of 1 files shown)


SiegeLord:
Hey, @peterhull90, could you check this over a bit? In particular, is the `setStyleSelector` trick I did the best that can be done?

peterhull90:
As far as I know there is no better way than what you did for `setStyleSelector` - it's a pity Obj-C can't do some sort of autoboxing to make these trivial methods unnecessary.
I also noted:
* In line 2201 you don't need to release because `+numberWithInt` creates an autorelease object - however, I'm not sure if we have a working autorelease pool on that thread. Might be better to stick to manual memory management (use alloc/initWithInt instead )
* I know it's not in this patch but the case for ALLEGRO_RESIZABLE doesn't call out to the main thread like the other ones do. As you know some flags seem to need it and some just work anyway, and that list can change between OS releases. Might be as well to make it consistent.
* Also not in this patch but the code never (AFAICS) actually updates the platform-independent `display->flags`. This means in ex_noframe, the code intended to toggle the frameless state doesn't work properly - the frame appears but can't be hidden again. 

Hope that helps,
Pete

SiegeLord:
Thanks Peter! That was very helpful, I appreciate it.

PR #590, opened 2016-03-14
add al_get/set_depth/samples functions

elias-pschernig:
This allows using a depth buffer as well as multi-sampling when drawing
into bitmaps with al_set_target_bitmap.


elias-pschernig:
This should do nothing as long as you don't use the 4 functions it adds. Those however definitely need testing. They will work with OpenGL 1.5 as long as the extensions used are available.

And it should be possible to make it work with OpenGL ES 2.0 (probably even 1.1, but I doubt anything still uses that). Maybe even with DirectX. I might attempt to add OpenGL ES support at some point as my game actually has use for drawing to an intermediate bitmap and in that case will need a depth buffer for it.

SiegeLord:
This is pretty sweet, although the `al_set_depth/samples` API make no sense:

- It takes effect at non-deterministic times (apparently only when you set up the FBO)
- Calling `al_get_depth/samples` returns basically bogus values if you call `al_set_depth/samples` after the settings take effect

I think those really should be in TLS (and be something like al_set_new_bitmap_depth/samples).

FBO finalization doesn't seem like it will work for bitmap locking: it seems to take effect when you switch target bitmaps, but that won't work if you lock the target bitmap after drawing to it. It seems that at the very least, you should finalize the FBO then as well.

elias-pschernig:
I don't like to use TLS, it doesn't really make sense to create more than one or two bitmaps with these special properties, but if you forget the setting on TLS you might not immediately notice and accidentally create all your game sprites with a depth buffer attached. I was thinking about a function al_create_bitmap_ex(w, h, depth, samples, flags) - but I don't like having _ex _ex2 _ex3 functions, it just feels messy. Operating *on* a bitmap seemed like the best bet.

The get variants will return 0 or whatever was set by a previous set call - they should never return bogus values.

We probably can make it so that you can change the values at runtime, destroying the depth-buffer/multisample-buffer. In fact it probably breaks right now when the FBO gets assigned to a different bitmap, so this has to be fixed anyway.

We could also have a more transparent API with separate ALLEGRO_DEPTH_BUFFER and ALLEGRO_MULTISAMPLE_BUFFER objects which you have to create and destroy and attach to an ALLEGRO_BITMAP to be used when drawing into it:

bitmap = al_create_bitmap(100, 100)
depth_buffer = al_create_depth_buffer(100, 100)
multisample_buffer = al_create_multisample_buffer(100, 100)
al_attach_bitmap_depth_buffer(bitmap, depth_buffer)
al_attach_bitmap_multisample_buffer(bitmap, multisample_buffer)

This way it would be clear thorugh the API itself that those are heavy objects. And we would leave it to the user to manage their lifetime.

SiegeLord:
> create all your game sprites with a depth buffer attached

It'll only get attached if you draw into them, no? Anyway, I think that's not particularly worse in this case than accidentally forgetting to revert `al_set_new_bitmap_flags(ALLEGRO_MEMORY_BITMAP)`. It might be nice to have a `al_create_bitmap_with_options` which is somehow future-proof, but that might be overkill. I still think TLS would be the most consistent option.

> The get variants will return 0 or whatever was set by a previous set call - they should never return bogus values.

What is set by the previous can get arbitrarily out of sync with what number of samples the bitmap actually has, since that's set in stone once you attach these things to it.

> We could also have a more transparent API

That seems a bit... excessive.

Anyway, it's probably fine for now, as we'd need to look at what the implementation looks like for Direct3D.

elias-pschernig:
Ok, I fixed a few issues and made it use TLS instead of bitmap setters. I couldn't figure out how to make the multisample buffers get realized on locking.

Incidentally, OpenGL ES has a different multisample extension which can directly draw to a texture so locking works there. Basically we have:

ARB_texture_multisample - locking works but requires OpenGL 3+, not used in my patch
GL_EXT_multisampled_render_to_texture - OES version of the same, used with OES in the patch
GL_EXT_framebuffer_multisample - no locking, but works all the way down to OpenGL 1.5, used in the patch

I updated the documentation to mention that you can't lock a multi-sample target while it is still the target. I think that is a very minor restriction. I don't mind if someone wants to make the changes necessary to allow locking the current target of course - it doesn't look that difficult, my brain just can't understand all the different paths for desktop GL and OES and iOS and raspberry... :P

src/opengl/ogl_fbo.c
```diff
@@ -93,11 +93,191 @@ void _al_ogl_reset_fbo_info(ALLEGRO_FBO_INFO *info)
...
+   ALLEGRO_BITMAP *b = info->owner;
+   int samples = al_get_bitmap_samples(b);
+
+   if (info->buffers.multisample_buffer != 0) {
+
+      if (info->buffers.samples != samples ||
+               info->buffers.mw != al_get_bitmap_width(b) ||
+               info->buffers.mh != al_get_bitmap_height(b)) {
+            detach_multisample_buffer(info);
+         }
```
SiegeLord: This has an extra indent.

src/opengl/ogl_fbo.c
```diff
@@ -93,11 +93,191 @@ void _al_ogl_reset_fbo_info(ALLEGRO_FBO_INFO *info)
...
+   ALLEGRO_BITMAP *b = info->owner;
+   int bits = al_get_bitmap_depth(b);
+
+   if (info->buffers.depth_buffer != 0) {
+
+      if (info->buffers.depth != bits ||
+               info->buffers.dw != al_get_bitmap_width(b) ||
+               info->buffers.dh != al_get_bitmap_height(b)) {
+            detach_depth_buffer(info);
+         }
```
SiegeLord: Another extra indentation.

examples/ex_depth_target.c
```diff
@@ -0,0 +1,236 @@
+
```
SiegeLord: Probably don't need this blank line :P.

SiegeLord:
This is awesome!

MarkOates:
I'm really excited about this PR.  I pulled a fresh copy just to try this out!

docs/src/refman/graphics.txt
```diff
@@ -1144,30 +1198,6 @@ Return the multi-sampling samples used by this bitmap if it is used with
 
 Since: 5.1.14
 
-### API: al_set_bitmap_depth
```
MarkOates: It's too bad this had to go.  I really liked the idea and simplicity of `al_set_bitmap_depth` and `al_set_bitmap_samples`. :blush:

SiegeLord:
Merged as b8672bdab2fc8598e4f18d0586b40609045014a8 (I squashed everything). Woohoo!

PR #594, opened 2016-03-19
Added documentation to ex_bitmap example.

wirecat:
While going through this example I thought it would be nice to have some documentation to help the next newbie with understanding the intent of the code.


elias-pschernig:
Thanks! One example down, 99 to go :) Eventually we should find a way to add the examples to the documentation like with Allegro 4.

PR #600, opened 2016-03-21
Added more inline documentation to ex_bitmap_flip

wirecat:


trentg:
My 2c is rather than repeat the same documentation for each example, one of the demos should be documented.

trentg:
Not that this is bad... I appreciate the effort.

examples/ex_bitmap_flip.c
```diff
@@ -138,6 +157,9 @@ int main(int argc, char **argv)
             if (event.keyboard.keycode == ALLEGRO_KEY_ESCAPE)
                done = true;
             else if (event.keyboard.keycode == ALLEGRO_KEY_SPACE) {
+				/* Spacebar toggles whether render from a memory bitmap 
+				 * or display bitamp.
+				 */
```
MarkOates: You gotta watch your spaces and tabs here.  3 spaces instead of tabs.

I don't know what editor you're using; I use VIM and had to setup the configuration so that whenever I use `TAB` it actually replaces it with 3 spaces.

MarkOates:
Even on single line comments we should always use C-style comments like so:
```c
/* single line comment */
```
Historically, Allegro has used these C-style comments throughout and we should continue that to be consistent with the rest of the code base.

wirecat:
@trentg I think that is a good idea. Would be much easier to keep up to date with any library changes.

wirecat:
@MarkOates Thanks! I'll keep a closer eye on that from now on.

trentg:
@wirecat that's my motivation. The examples get updated sometimes and I can see a lot of comments getting out of sync with the code if a lot of care isn't taken.

allefant:
Maybe we should group our current examples and move them into sub-folders like "tutorials", "feature-demos", "feature-tests"... then only comment the tutorial ones. Another idea I keep thinking about would be an example runner app - something with a simple guy which allows running all the other examples. That could be especially useful on platforms like iOS and Android where it's cumbersome installing and trying all the examples. But I'm getting off topic here...

wirecat:
@allefant I like the idea of having a tutorials section. Then it would be clear what their purpose is and it would make sense to have the documentation target a newbie.

SiegeLord:
Merged as 978e71fd57d94a2380a71cef59bfe14851960d1f.

I decided that this change was a very small net improvement. I'd rather not document every example, as per the points made above, but documenting a few of the common ones seems harmless.

Issue #602, opened 2016-03-24
al_set_sample_instance_length is poorly implemented


SiegeLord:
- Doesn't seem to work when playmode is set to loop
- Does weird things when set to longer than the sample length (static + crashing)

PR #607, opened 2016-03-31
Sample instance length

rcorre:
This addresses part of #602. It makes looping respect the length, but you have to restart the stream (adjusting the length while it is playing does nothing).


SiegeLord:
Merged as 4f71e69. Thanks!

PR #608, opened 2016-04-03
Add necessary includes for top level headers to be used in isolation.

woehr:
I've been playing around with a tool to automatically create bindings for allegro. The tool runs the C preprocessor on each of the top-level headers. However, in some headers this fails because they don't include the necessary dependencies. This pull request makes those dependencies explicit for the headers that I found had problems.

### DIFF
--- a/include/allegro5/allegro_opengl.h
+++ b/include/allegro5/allegro_opengl.h
@@ -22,6 +22,8 @@
    extern "C" {
 #endif
 
+#include <stdint.h>
+
 #if defined(ALLEGRO_WINDOWS)
 #include <windows.h>
 #endif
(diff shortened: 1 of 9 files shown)


include/allegro5/allegro_opengl.h
```diff
@@ -22,6 +22,8 @@
    extern "C" {
 #endif
 
+#include <stdint.h>
```
SiegeLord: Could you make this include allegro5/base.h instead? Technically we support platforms without stdint.h, and base.h takes care of the difference.

woehr: I've removed the include of stdint.h. The other includes include base.h so it ended up be superfluous.

SiegeLord:
Thanks!

PR #609, opened 2016-04-06
fix compilation on iOS

elias-pschernig:
I updated to OSX 10.11.4 and XCode 7.3 but needed these changes to compile.


elias-pschernig:
It seems @import only works when modules are enabled - but our CMake currently does not. So the other way to fix this would be to add a flag to CMake. Because I don't know where in CMake to add that flag, I think this commit is the better option for now. Also, if we are to switch to modules, we should use them for all frameworks, not just for a single one.

PR #613, opened 2016-04-10
make it work with clang 3.1

elias-pschernig:
Replaces #541 which was against the 5.1 branch.

### DIFF
--- a/addons/native_dialog/osx_dialog.m
+++ b/addons/native_dialog/osx_dialog.m
@@ -169,12 +169,14 @@ int _al_show_native_message_box(ALLEGRO_DISPLAY *display,
                                 ALLEGRO_NATIVE_DIALOG *fd)
 {
     (void)display;
-    @autoreleasepool {
+    {
+        NSAutoreleasePool *pool = [[NSAutoreleasePool alloc] init];
         NSValue* fdValue = [NSValue valueWithPointer:fd];
         
         [ALLEGAlertWrapper performSelectorOnMainThread: @selector(displayAlert:)
                                          withObject: fdValue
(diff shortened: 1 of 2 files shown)


peterhull90:
I've been looking at which version of clang goes with which version of XCode. From this
https://en.wikipedia.org/wiki/Xcode
AFAICS Apple Clang 3.1 goes with XCode 4.3, Clang 4.0 goes with XCode 4.4 and the latter drops support for OS X Snow Leopard (2009). (However both 3.1 and 4.0 were based on LLVM 3.1).
But I'm not sure how these *compiler* changes affect the target *runtime* version - i.e. does this patch allow people with relatively old versions of XCode to continue compiling Allegro or does it affect the end user (people writing games with Allegro and people playing games written with Allegro)

All in all, I'm confused :confused:

trentg:
Does this compile with newer clang as well? I'm not sure but I recall some of the lines you've changed were just changed recently... not sure if it matters but perhaps someone remembers what those changes were about and if these modifications don't undo them.

peterhull90:
Yes I am confident it will work on newer versions. The changes are just
removing uses of the shortcut methods for object literals which were added
in more recent Clangs.

elias-pschernig:
Basically @{ is a shortcut for creating an NSDictionary, @autoreleasepool is a shortcut for [NSAutoreleasePool] alloc/init/release, and "instancetype" is similar to C++ "auto" but you can still use "id" (which is similar to C++ "void *").

SiegeLord:
Merged as 74855c8 for consistency. That said, my preference would be to drop support for older clang versions and modernize the code wholesale at some point.

PR #616, opened 2016-04-11
Use PROJECT_*_DIR instead of CMAKE_*_DIR

### LINKED ISSUE #612: Use `PROJECT_*_DIR` instead of `CMAKE_*_DIR` in CMake files
By using `PROJECT_SOURCE_DIR` and `PROJECT_BINARY_DIR` instead of the `CMAKE_` prefixed equivalents this will allow Allegro to be built from source as a dependency simply by doing the following:

```
add_subdirectory ("path_to/allegro5")
include_directories("path_to/extern/allegro5/include")
include_directories("path_to/extern/allegro5/addons/main")
target_link_libraries(${PROJECT_NAME} allegro allegro_main ...)
```

Currently due to the usage of `CMAKE_*_DIR` the Allegro CMake files look in the wrong places for their files when Allegro is built by another project in this way.

peterhull90:
This change was just made by using sed. It passes Travis without any issues. Is this all that is required, @adtennant?

Fixes #612

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,7 +33,7 @@ option(WANT_ANDROID "Build for Android" OFF)
 option(WANT_ANDROID_LEGACY "Build for Android 4 (1.6)" OFF)
 option(WANT_GLES2 "Compile with GLES2 support" ON)
 if(WANT_ANDROID)
-    include(${CMAKE_SOURCE_DIR}/cmake/Toolchain-android.cmake)
+    include(${PROJECT_SOURCE_DIR}/cmake/Toolchain-android.cmake)
     set(ANDROID_TARGET "android-12" CACHE STRING "What Android target to compile for.")
 endif(WANT_ANDROID)
 option(ALLEGRO_SDL "Build using the SDL backend (experimental)" OFF)
@@ -48,14 +48,14 @@ string(REGEX MATCH "^[0-9]+[.][0-9]+" ALLEGRO_SOVERSION ${ALLEGRO_VERSION})
 string(REPLACE "." "" ALLEGRO_DLL_SHORTVER ${ALLEGRO_SOVERSION})
 
 # Search in the `cmake' directory for additional CMake modules.
-list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
+list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)
 
 # Search in `deps' directories for dependency files.
 file(GLOB deps_subdirs
-    "${CMAKE_SOURCE_DIR}/deps"
-    "${CMAKE_SOURCE_DIR}/deps/*"
-    "${CMAKE_BINARY_DIR}/deps"
-    "${CMAKE_BINARY_DIR}/deps/*"
+    "${PROJECT_SOURCE_DIR}/deps"
+    "${PROJECT_SOURCE_DIR}/deps/*"
+    "${PROJECT_BINARY_DIR}/deps"
+    "${PROJECT_BINARY_DIR}/deps/*"
     )
 foreach(subdir ${deps_subdirs})
     if(EXISTS "${subdir}/include" OR EXISTS "${subdir}/lib")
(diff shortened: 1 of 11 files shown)


adtennant:
That seems to work perfectly.

SiegeLord:
Thanks!

goobliata:
For me this breaks the build. PROJECT_SOURCE_DIR is empty (not sure about PROJECT_BINARY_DIR). I'm on Windows under MSYS using a recent version of CMake.

SiegeLord:
Is this an old build directory or a freshly made one?

elias-pschernig:
It's not really clear to me from the cmake documentation, but a short test looks like it should still work:

I put this in CMakeLists.txt:
```
message(STATUS "PROJECT_SOURCE_DIR=${PROJECT_SOURCE_DIR}")
message(STATUS "CMAKE_SOURCE_DIR=${CMAKE_SOURCE_DIR}")
message(STATUS "PROJECT_BINARY_DIR=${PROJECT_BINARY_DIR}")
message(STATUS "CMAKE_BINARY_DIR=${CMAKE_BINARY_DIR}")
```

Then running "cmake ." says (with cmake 3.5):
```
-- PROJECT_SOURCE_DIR=/tmp/blah
-- CMAKE_SOURCE_DIR=/tmp/blah
-- PROJECT_BINARY_DIR=/tmp/blah
-- CMAKE_BINARY_DIR=/tmp/blah
```

So it does not look like it is intended to ever be empty.

elias-pschernig:
Ah, found the problem, look at this example:
```
message(STATUS "PROJECT_SOURCE_DIR=${PROJECT_SOURCE_DIR}")
project(ALLEGRO C)
message(STATUS "PROJECT_SOURCE_DIR=${PROJECT_SOURCE_DIR}")
```
which produces:
```
-- PROJECT_SOURCE_DIR=
-- PROJECT_SOURCE_DIR=/tmp/blah
```
Oh the joy of CMake where nothing is fully documented and everything has uncalled for exceptions.

The solution for the Allegro CMakeLists.txt is to try and put the project() call before any references to PROJECT_*_DIR, if that is possible. Note though that some things in CMake have to be done before that - so I can't really say it's possible. I'd tend to revert the patch for now.

goobliata:
Yes, that's it, in fact I was building for Android which uses the variable before the project statement.

SiegeLord:
What if we do something like this:

```cmake
# Set the project name.
# We use C++ in a few cases.
project(ALLEGRO NONE)

if(WANT_ANDROID)
    include(${PROJECT_SOURCE_DIR}/cmake/Toolchain-android.cmake)
    set(ANDROID_TARGET "android-12" CACHE STRING "What Android target to compile for.")
endif(WANT_ANDROID)

enable_language(C)
enable_language(CXX)
```

Would that work?

goobliata:
-- Selected Android toolchain: C:/Users/trent/code/android-ndk-arm
-- The C compiler identification is GNU 4.9.0
-- Check for working C compiler: C:/Users/trent/code/android-ndk-arm/bin/arm-linux-androideabi-gcc.exe
CMake Error at C:/Program Files (x86)/CMake/share/cmake-3.5/Modules/CMakeTestCCompiler.cmake:47 (try_compile):
  The test project needs language RC which is not enabled.
Call Stack (most recent call first):
  CMakeLists.txt:45 (enable_language)

SiegeLord:
Wow... looks like RC is a hidden default language. Need to add `enable_language(RC)` too.

goobliata:
That fixes that problem, but now CMake is trying to compile a (Windows) dll in the C compiler test... for some reason, even though the Android toolchain is loaded first.

SiegeLord:
Hmm, alright, nevermind. Let's just stick the `CMAKE_SOURCE_DIR` in that one spot as clearly working around it will take a bit of effort.

All that said, perhaps we shouldn't even be including the toolchain from within `CMakeLists.txt` at all but rather require people to pass it via the cmake command (as we appear to do for iOS and RaspberryPi). Any opinions on this?

goobliata:
That's fair. It appears to be a bit hacky to load the toolchain from CMake. If nobody objects I can make the change and change the README.

goobliata:
So I guess if you specify a toolchain file, it doesn't get anything you specify with -D so -DANDROID_NDK_TOOLCHAIN_ROOT won't work. It has to be an environment variable. I guess Toolchain-android.cmake does say that it's an environment variable in its error message though, so maybe it was at some point.

goobliata:
I've got a patch that moves the toolchain to the command line. Not sure if that's ultimately what we want. Pick your poison, but here it is: https://github.com/goobliata/allegro5/commit/06bda91d40c0c60fc1d1a4007398e5f437588868

elias-pschernig:
I suppose this is the intended use of toolchain files. Makes building even more of a pain, but I think it is what we should do.

PR #618, opened 2016-04-12
Optimize bitmap hold blend mode

fatcerberus:
During a bitmap hold, this changes Allegro to defer the blend mode change until a vertex flush.  On some graphics chipsets (Intel, e.g.), `glBlend*()` calls are very expensive.  This defers them during `al_hold_bitmap_drawing` until the hold is released.  The change should be safe since it's already undefined behavior to change blend modes during `al_hold_bitmap_drawing`.

### DIFF
--- a/src/opengl/ogl_bitmap.c
+++ b/src/opengl/ogl_bitmap.c
@@ -364,7 +364,7 @@ static void ogl_draw_bitmap_region(ALLEGRO_BITMAP *bitmap,
       }
    }
    if (disp->ogl_extras->opengl_target == target) {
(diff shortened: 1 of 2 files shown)


fatcerberus:
Hm, GitHub says Travis failed but it actually didn't.  One of the clang jobs errored due to a package manager glitch.  The rest of the builds passed.

SiegeLord:
Merged as e7cbc439ed25dbaa1c479d75f57b3161531f74db. I added a commit after it to completely remove the call to `_al_opengl_set_blender` outside of the flush cache, so that it's called only once per draw call if bitmap drawing is not held.

PR #619, opened 2016-04-12
Fix GDI+ finding when using cmake-gui

fatcerberus:
CMake seems to be unable to find the Windows 8.1 SDK on Win10 automatically unless it's in the `PATH`.  As a result GDI+ usually doesn't get found unless cmake is run from the command line in an MSVC command prompt.

This PR updates `FindGDIPLUS.cmake` to use the same search method as for the DirectX libraries so GDI+ can be found regardless of the `PATH`, which allows it to work properly in `cmake-gui`.


SiegeLord:
Merged in ae16e2f608cbd8e058c8b4bb51df89ca1b97b892. This actually fixes it for command line too, in my testing.

PR #621, opened 2016-04-17
Opus codec support.

BorisCarvajal:
Based on Ogg Vorbis reader so expect the same functionality.


SiegeLord:
Awesome! There were just a few double-free errors that I had to fix, but otherwise this is pretty great. Merged as 9534c9d8be2171a7a90be52c406d46531b9a56c4. Thanks!

PR #624, opened 2016-04-20
Add an example of using Enet with Allegro.

### LINKED ISSUE #496: Add an example multiplayer network game
Either in the wiki or in the distribution we need an example of how to do a simple multi player game with an existing network library, such as enet, to show that Allegro doesn't have to invent the square wheel again by making a networking add on.  

Related to #465.

### LINKED ISSUE #554: Feature request: Networking/sockets
It would be great if Allegro provided basic support for networking.  It wouldn't have to be complicated, just a thin wrapper around the platform's native sockets implementation.  Winsock and POSIX sockets are different enough that writing platform-independent socket code is pretty tedious.

rcorre:
This should help address questions about why Allegro does not include
its own networking code. The example comes in two parts: ex_enet_client
and ex_enet_server.

The example coordinates movement between circles each controlled by a
separate client. The clients send changes in movement direction to the
server and rely on the server as an authoritative source of their
positions.

Resolves #554 
Resolves #496

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -243,6 +243,14 @@ if(WANT_CURL_EXAMPLE)
     endif(CURL_FOUND)
 endif(WANT_CURL_EXAMPLE)
 
+# The enet example won't compile if the enet library isn't available.
+# This example isn't important so it's disabled by default to prevent problems.
+option(WANT_ENET_EXAMPLE "Build ex_enet example" off)
+if(WANT_ENET_EXAMPLE)
+    example(ex_enet_client -lenet ${PRIM})
+    example(ex_enet_server -lenet)
+endif(WANT_ENET_EXAMPLE)
+
 # example(ex_ogre3d ex_ogre3d.cpp)
 # include_directories(/usr/include/OGRE)
 # target_link_libraries(ex_ogre3d OgreMain)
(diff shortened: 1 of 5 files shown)


examples/CMakeLists.txt
```diff
@@ -243,6 +243,14 @@ if(WANT_CURL_EXAMPLE)
     endif(CURL_FOUND)
 endif(WANT_CURL_EXAMPLE)
 
+# The enet example won't compile if the enet library isn't available.
+# This example isn't important so it's disabled by default to prevent problems.
+option(WANT_ENET_EXAMPLE "Build ex_enet example" off)
```
rcorre: Ideally I'd use `find_package`, but support for Enet isn't built into Cmake. I did find [this](https://github.com/stuntrally/stuntrally/blob/master/cmake/FindENet.cmake) which might work.

SiegeLord: Yeah, we created a few of our own Find* packages. Probably shouldn't use that specific one, as it's probably licensed as GPL.

Just grab `cmake/FindDUMB.cmake` in the Allegro's source and modify it to search for enet instead.

SiegeLord:
Merged as 9b454045837edb969d80d562e707a93f86afc147. These examples are pretty sweet! Thanks!

rcorre:
Thank you! I'd been meaning to learn enet anyways, and those issues were a good excuse.

PR #625, opened 2016-04-20
make enumerating connected joysticks more robust

peterhull90:
Developed by @NewCreature. As discussed on [a.cc](https://www.allegro.cc/forums/thread/616193/1021898#target)

### DIFF
--- a/src/macosx/hidjoy.m
+++ b/src/macosx/hidjoy.m
@@ -50,8 +50,11 @@ typedef enum {
 #define JOYSTICK_USAGE_NUMBER      0x04
 #define GAMEPAD_USAGE_NUMBER       0x05
 
+#define JOYSTICK_NAME_MAX           256
+
 typedef struct {
    ALLEGRO_JOYSTICK parent;
+   char name[JOYSTICK_NAME_MAX];
    IOHIDElementRef buttons[_AL_MAX_JOYSTICK_BUTTONS];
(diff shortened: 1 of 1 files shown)


SiegeLord:
Finally got around to looking at this, sorry for the delay.

One issue I found is that it somehow detects 1 extra button than the old code for my Logitech F310 controller. It detects 13 buttons, while the old code only detects 12 (and try as I might, I don't see where the 13th button is). I looked at the code, but it's not obvious to me where the issue is (or if it's a bug in the OSX itself).

NewCreature:
Do the analog sticks click down as buttons? My Wii U Pro controllers had 2 extra buttons I couldn't locate, but it turns out it was those.

SiegeLord:
Yeah, it detected the analog stick buttons fine.

I managed to find the 13th button, but I don't seem to get events from it (I do get events from it on Linux). The input callback doesn't even get called for it, so it's not like we're filtering it out somehow.

fatcerberus:
I used to have an F310 myself, I remember a lot of software (including Windows control panel) used to detect a seemingly phantom button on it.  It's the Analog button, isn't it?

SiegeLord:
It's the button between back and start buttons in the middle. The one with the logitech logo.

NewCreature:
Maybe try it with [Enjoyable](https://yukkurigames.com/enjoyable/) to see if it detects the 13th button and allows it to map to something.

SiegeLord:
Merged as 0828211. Thanks to you both!

I decided not to worry about the 13th button, this patch is still a strict improvement over the current situation.

PR #628, opened 2016-05-01
Make set_shader_uniform const-correct

fatcerberus:
`al_set_shader_matrix()` and `al_set_shader_*_vector()` don't modify the arrays passed to them, so they should be `const`.

### DIFF
--- a/include/allegro5/internal/aintern_shader.h
+++ b/include/allegro5/internal/aintern_shader.h
@@ -24,13 +24,13 @@ struct ALLEGRO_SHADER_INTERFACE
    bool (*set_shader_sampler)(ALLEGRO_SHADER *shader, const char *name,
          ALLEGRO_BITMAP *bitmap, int unit);
    bool (*set_shader_matrix)(ALLEGRO_SHADER *shader, const char *name,
-         ALLEGRO_TRANSFORM *matrix);
+         const ALLEGRO_TRANSFORM *matrix);
    bool (*set_shader_int)(ALLEGRO_SHADER *shader, const char *name, int i);
    bool (*set_shader_float)(ALLEGRO_SHADER *shader, const char *name, float f);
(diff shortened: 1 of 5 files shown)


SiegeLord:
This is an interesting change from the point of view of API stability. This doesn't affect the ABI, since the argument types are not encoded for C functions. However, it may cause code that previously compiled no longer compile, e.g.:

```c
bool (*ptr)(const char*, ALLEGRO_TRANSFORM*) = al_set_shader_matrix;
```

Historically, Allegro has never provided that kind of level of source compatibility though, so it's not like it'd be unprecedented.

Any opinions on this, @elias-pschernig?

elias-pschernig:
Well, I never was one who felt strong about ABI/API compatibility to begin with - so I'd say it's fine :)

fatcerberus:
Hm, that's true.  I just get frustrated in cases like this because my own code is const correct (to the best of my ability) and I don't like casting away the const as many compilers issue warnings for doing so.

elias-pschernig:
This reminds me, there is other places which technically could be made const-correct, for example it should be al_draw_bitmap(ALLEGRO_BITMAP const *bitmap, ...) since the bitmap is not modified. At the same time some drivers may require synchronizing a texture or other things inside of the ALLEGRO_BITMAP, or the texture may even get locked to the current thread and so on - so in reality it definitely may not be const at all. In any case, for parameters like the one in your patch I definitely think we should add the const as you did.

fatcerberus:
I tend to err on the side of caution for highly mutable objects like textures and such even in my own code so I know what you mean there.

SiegeLord:
Merged as 15a3f42f4da4a4354867839959a40d2f4ede5fd6.

I decided that we shouldn't provide this level of API stability, it's just too restrictive on our end, and doesn't enable any sort of useful code on the user's end.

PR #632, opened 2016-06-01
Fix Android app woken up during sleep

goobliata:
I spent 40 hours trying to figure out how to fix my music from starting when my Android TV system goes to sleep and I have backgrounded my game. Here's what I learned:

- Allegro generates a DISPLAY_RESUME event in surfaceChanged. It seems like it would be better placed in onResume, however threading issues make it (near?) impossible... I couldn't get it to work anyway, not without rewriting everything.
- surfaceChanged can be called at any time, it doesn't depend on the Android lifecycle... so onResume isn't always called afterwards. This leads to screenblank activating which for some stupid reason generates a surfaceChanged that wakes up the game, but the system doesn't wake up.
- surfaceCreated always gets called during a resume, so with this patch, the surface is always recreated and in my testing on phone and TV systems, this doesn't lead to a zombie state like https://github.com/liballeg/allegro5/commit/c3493b46eaf47f34b7393a70f9be1451c1e8d11e.
- Unlike the linked patch, this destroys the surface on the way into a pause, so there's no double free. And it gets recreated when the app resumes.

So all this effectively changes is one erroneous RESUME_DRAWING is avoided... but I welcome further testing. I guess where there is no surface, that surfaceChanged event is avoided...

The issue with putting the DISPLAY_RESUME in onResume is that onResume would have to wait for al_acknowledge_drawing_resume. However al_acknowledge_drawing_resume depends on surfaceChanged being finished... but they can happen in any order so the only potential way to work it out is a bunch of while (!flag); stuff (conditions don't work three way branches.)

### DIFF
--- a/src/android/android_system.c
+++ b/src/android/android_system.c
@@ -312,6 +312,9 @@ JNI_FUNC(void, AllegroActivity, nativeOnPause, (JNIEnv *env, jobject obj))
          _al_event_source_emit_event(&display->es, &event);
       }
       _al_event_source_unlock(&display->es);
+
+      ALLEGRO_DISPLAY_ANDROID *d = (ALLEGRO_DISPLAY_ANDROID*)display;
+      _al_android_destroy_surface(env, d, false);
    }
 }
 
--- a/src/android/android_system.c
+++ b/src/android/android_system.c
@@ -314,7 +314,8 @@ JNI_FUNC(void, AllegroActivity, nativeOnPause, (JNIEnv *env, jobject obj))
       _al_event_source_unlock(&display->es);
 
       ALLEGRO_DISPLAY_ANDROID *d = (ALLEGRO_DISPLAY_ANDROID*)display;
-      _al_android_destroy_surface(env, d, false);
+      if (d->created)
+         _al_android_destroy_surface(env, d, false);
    }
 }
 


elias-pschernig:
So basically with that patch we destroy the android surface in onPause and re-create it in onResume. That sounds like it should work in theory. However I'm a bit worried that if you slide down the settings from the top or slide in a side bar, the Allegro game will not work correctly anymore (as onPause will be called in those cases even though the game is still visible). Need to confirm that still works - I can try and test maybe during this week.

goobliata:
I've been implementing Google Play achievements over the past day or two and those work fine. They popup little banners and even some text input things.

(I had to patch AllegroActivity to extend FragmentActivity to base my code on Google's examples... but that requires an extra support library so I would not commit that to the master repo... however it's here: https://github.com/goobliata/allegro5/tree/android_fragment_activity, you just need the support-v4 library which was a pain to make usable with Ant projects :P)

elias-pschernig:
Do you have an .apk with the patch applied? I was more thinking about things like sliding in another app (from a different activity) on top of your game - I might be able to try it out during lunch break at work. I also want to test it on my game before applying. I once before applied a similar patch by someone (it destroyed the surface in onStop), then forgot about it and spent two entire weeks figuring out why my game sometimes would end up in a zombie state without a surface.

As I know now, the Android activity lifecycle is actually extremely simple:
                               
           onStart       onResume
stopped <-> paused <-> running
           onStop        onPause

It's a bit unfortunate that we made it such a mess of it in Allegro :P

goobliata:
I emailed you a link to an APK.

elias-pschernig:
It seems to work fine with my game, except once I got the below crash. I suppose we could simply check if the surface is already Null then guard against it. I unfortunately don't have more complete logs anymore and haven't been able to reproduce anymore - but it happened within a moment of starting the game, before anything was visible even.

I don't fully follow our display creation code on Android, but I see that in nativeOnResume we only re-create the surface if the "created" flag of our display is set. So there is the possibility of receiving onPause and onResume events where with the patch onPause will destroy the surface and onResume will not create it - and then the next onPause will lead to the crash? Just guessing.

Do you think it would make sense to only destroy the surface in onPause if the "created" flag is set?

I remember from hunting that last zombie-patch that on some devices Android will send many rapid onPause and onResume events to the app within the first milliseconds of starting (Samsung S6 and S7 especially) - which is exactly the time a game usually creates its display. (I used the free 200 Amazon test cloud or whatever minutes to try on all those devices... unfortunately I fully used those up back then so now I only have my personal phone to test on :P)

```
06-03 22:07:31.240 21797-21797/? D/AllegroActivity: destroySurface
06-03 22:07:31.241 21797-21797/? W/System.err: java.lang.NullPointerException: Attempt to invoke virtual method 'android.view.ViewParent org.liballeg.android.AllegroSurface.getParent()' on a null object reference
06-03 22:07:31.242 21797-21797/? W/System.err:     at org.liballeg.android.AllegroActivity.destroySurface(AllegroActivity.java:168)
06-03 22:07:31.243 21797-21797/? W/System.err:     at org.liballeg.android.AllegroActivity.nativeOnPause(Native Method)
06-03 22:07:31.244 21797-21797/? W/System.err:     at org.liballeg.android.AllegroActivity.onPause(AllegroActivity.java:374)
06-03 22:07:31.244 21797-21797/? W/System.err:     at android.app.Activity.performPause(Activity.java:6363)
06-03 22:07:31.245 21797-21797/? W/System.err:     at android.app.Instrumentation.callActivityOnPause(Instrumentation.java:1311)
06-03 22:07:31.246 21797-21797/? W/System.err:     at android.app.ActivityThread.performPauseActivity(ActivityThread.java:3367)
06-03 22:07:31.246 21797-21797/? W/System.err:     at android.app.ActivityThread.performPauseActivity(ActivityThread.java:3340)
06-03 22:07:31.247 21797-21797/? W/System.err:     at android.app.ActivityThread.handlePauseActivity(ActivityThread.java:3315)
06-03 22:07:31.247 21797-21797/? W/System.err:     at android.app.ActivityThread.-wrap13(ActivityThread.java)
06-03 22:07:31.248 21797-21797/? W/System.err:     at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1355)
06-03 22:07:31.248 21797-21797/? W/System.err:     at android.os.Handler.dispatchMessage(Handler.java:102)
06-03 22:07:31.249 21797-21797/? W/System.err:     at android.os.Looper.loop(Looper.java:148)
06-03 22:07:31.249 21797-21797/? W/System.err:     at android.app.ActivityThread.main(ActivityThread.java:5422)
06-03 22:07:31.250 21797-21797/? W/System.err:     at java.lang.reflect.Method.invoke(Native Method)
06-03 22:07:31.251 21797-21797/? W/System.err:     at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
06-03 22:07:31.251 21797-21797/? W/System.err:     at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
```

goobliata:
I added an `(d->created)` before destroying the surface as you suggested. Doesn't seem to hard anything (created is set false in surfaceDestroyed and is initially false so should work.) Seems odd that Android doesn't adhere to it's supposed lifecycle sometimes, but what can you do. I will commit the current branch after squashing it.

PR #636, opened 2016-06-10
Fix al_get_text_dimensions to get exact bounding box

koro-xx:
al_get_text_dimensions is supposed to return the text bounding box, but currently only the width is correct (for the height it returns the font line height instead of a tight bounding box).
This fixes it so that bbh is the exact height of the bounding box and bby is the offset from the top.

### DIFF
--- a/addons/ttf/ttf.c
+++ b/addons/ttf/ttf.c
@@ -677,7 +677,7 @@ static void ttf_get_text_dimensions(ALLEGRO_FONT const *f,
       }
 
       if (nch < 0) {
-         x += gx + gw + 2;
+         x += gx + gw;
       }
       else {
          x += al_get_glyph_advance(f, ch, nch);
@@ -1005,10 +1005,10 @@ static bool ttf_get_glyph_dimensions(ALLEGRO_FONT const *f,
       }
    }
    cache_glyph(data, face, ft_index, glyph, false);
-   *bbx = glyph->offset_x;
+   *bbx = glyph->offset_x + 1;
    *bbw = glyph->region.w - 2;
-   *bbh = glyph->region.h;
-   *bby = glyph->offset_y;
+   *bbh = glyph->region.h - 2;
+   *bby = glyph->offset_y + 1;
       
    return true;
 }
--- a/addons/ttf/ttf.c
+++ b/addons/ttf/ttf.c
@@ -663,7 +663,8 @@ static void ttf_get_text_dimensions(ALLEGRO_FONT const *f,
    bool first = true;
    int x = 0;
    int32_t ch, nch;
-
+   int ymin = f->height;
+   int ymax = 0;
    *bbx = 0;
 
    nch = al_ustr_get_next(text, &pos);
@@ -682,15 +683,23 @@ static void ttf_get_text_dimensions(ALLEGRO_FONT const *f,
          x += al_get_glyph_advance(f, ch, nch);
       }
 
+      if (gy < ymin) {
+         ymin = gy;
+      }
+
+      if (gh+gy > ymax) {
+         ymax = gh + gy;
+      }
+
       if (first) {
          *bbx = gx;
          first = false;
       }
    }
 
-   *bby = 0; // FIXME
+   *bby = ymin;
    *bbw = x - *bbx;
-   *bbh = f->height; // FIXME, we want the bounding box!
+   *bbh = ymax - ymin; 
 }
 
 


koro-xx:
After doing this I noticed this is not an exact bounding box, since it adds an extra "margin" pixel (because al_get_glyph_dimensions does). By now I understood that glyph->region adds a margin pixel to glyphs; that's fine.

But al_get_glyph_dimensions seems to return the exact width of the pixel in bbw while it keeps the extra margin pixel in bbh (see [here](https://github.com/liballeg/allegro5/blob/93e25245d7bfa77f2f3248fba738a8958189d582/addons/ttf/ttf.c#L998-L1002)). 
Is there a reason for this inconsistency?
(I noticed that al_get_text_dimensions even adds 2 to bbw to account for the missing margin pixels from the width in al_get_glyph_dimensions...)

goobliata:
I'm not sure this is your issue, but fonts all define the heights differently. One font could look the same size as another but Freetype will return a taller height for it.

koro-xx:
I've also corrected al_get_glyph_dimensions so that it returns the exact bounding box now.

goobliata:
The first patch looks reasonable. How does the second patch affect text drawing? I imagine it is more condensed? I'd have to check it out...

addons/ttf/ttf.c
```diff
@@ -1005,10 +1005,10 @@ static bool ttf_get_glyph_dimensions(ALLEGRO_FONT const *f,
       }
    }
    cache_glyph(data, face, ft_index, glyph, false);
-   *bbx = glyph->offset_x;
+   *bbx = glyph->offset_x + 1;
    *bbw = glyph->region.w - 2;
```
goobliata: why is this left at -2 and the others were all changes?

koro-xx: The text rendering shouldn't be affected since this function is not used for rendering (as far as I can see).

The reason for the change in the other numbers is that glyph->region adds a 1 pixel margin to the glyphs, so to get the exact glyph bounding box you need to add 1 to the offsets and substract 2 to the width and height.

Apparently someone remembered this fact but only substracted 2 from region.w (at least that's the only explanation I can think of...).

The old version gets bbh, bbx and bby off by 1 or 2 pixels from the real position of the bounding box for the glyph, and the proposed patch should give a tight bounding box (in my tests it worked nicely).

goobliata:
I had a look and it looks good. ex_ttf has an error in it that it draws "Tulip"'s boxes 1 pixel too far to the right and bottom (it should be x + w - 1) but besides that it looks perfect.

addons/ttf/ttf.c
koro-xx: You're right, there's an issue with the offsets bbx and bby. The reason is that gylph->offset_x and glyph->offset_y don't include the extra pixel added to glyph->region.*.
So the return values should really be *bbx = glyph->offset_x and *bby = glyph->offset_y.
(so everything is shifted one pixel forward, accounting for the "error" that you saw in ex_ttf).

Is there a way I can amend this without opening a new pull request?

goobliata: Like I said, the problem is not here but in the example (ex_ttf) as far I I know. When you draw a bounding box you always draw x + w - 1 not x + w (think about it for a moment.) If you look closely at ex_ttf, the x position is right but the box is drawn 1 pixel too wide and tall.

goobliata: I fixed ex_ttf. It lines up perfectly for me now.

elias-pschernig:
Actually, if you have a bounding box from a/b "inclusive" to c/d "inclusive", the proper way to do it with Allegro is: al_draw_rectangle(a + 0.5, b + 0.5, c + 0.5, d + 0.5, 1). Remember, our default projection assumes pixel centers are at 0.5/0.5 and a line-width of 1 means a rectangle at the center spans the entire pixel. E.g. al_draw_rectangle(0.5, 0.5, 1.5, 1.5) will modify exactly these 4 pixels: 0/0, 1/0, 0/1, 1/1. If you want to instead clear a rectangle at x/y with size w/h it would be: al_draw_filled_rectangle(x, y, w, h) - so that case makes a bit more sense at least.

goobliata:
With 0.5 pixel offsets it should be x + 0.5, x + w - 0.5. Think about this: you have a box that is 2 pixels wide that starts at 0. It should be drawn on pixels 0 and 1 not 0 and 2 because 0 to 2 is 3 pixels.

addons/ttf/ttf.c
koro-xx: I have tested this extensively and I am sure I did make a mistake by returning glyph->offset_x+1. That's not right. The value "glyph->offset_x" is exactly the position where the glyph begins. You can see in [these lines](https://github.com/liballeg/allegro5/blob/master/addons/ttf/ttf.c#L521-L528) of ttf.c that the glyph is drawn with offsets exactly (offset_x, offset_y) (they grab the glyph image without including the 1 pixel margin). 

For example, this code:
```c
    char t = 'a';
    int x = 1;
    al_get_glyph_dimensions(default_font, t, &bbx, &bby, &bbw, &bbh);
    al_draw_glyph(default_font, WHITE_COLOR, x, x, t);
    x+=0.5; // to get pixel-aligned rectangle
    al_draw_rectangle(x+bbx, x+bby, x+bbx+bbw-1, x+bby+bbh-1, al_map_rgba(255,0,0,140), 1);
```
Produces this with the current version (after this patch):
ex_current

And this is what it looks like if one fixes the return values of bbx and bby (as glyph->offset_x without adding anything):
ex_shouldbe

goobliata: It looked example right to me in ex_ttf but I'll take a second look.

goobliata:
OK I think you're right koro-xx. I removed the + 1's. I also changed ex_ttf to use 0.5 offsets. Please check it out. However for me in ex_ttf to top blue line of the u is visible but it shouldn't be.

fatcerberus:
Regarding box drawing: A solid rectangle drawn at (0,0)-(2,2) is 2x2.  This is because with hardware rendering integer coordinates are at intersections of pixels, not their centers, unlike in pure raster drawing.

The only reason you need to add 0.5 for line drawing is because you want the lines to go through the center of pixels, otherwise they won't be visible.  If you just want to calculate a bounding box, (x,y)-(x+w,y+h) is indeed correct for GPU rendering.

goobliata:
x+w is never correct, it's more or less a fluke that it works WITH OPENGL. The only reason it works with D3D is Allegro adds a 0.5, 0.5 matrix to the transformation.

goobliata:
Just to expand on that, turn on multi-sampling and you'll see why x+w is never correct.

goobliata:
Sheesh, even without multisampling it doesn't draw a bounding box...

fatcerberus:
Right, D3D9 aligns (0,0) with the center of the first pixel when using an identity transform.  I gather they did that to make pixel-perfect drawing easier.  From what I've gleaned D3D10+ behaves like OpenGL, where (0,0) is the top left of the first pixel.

To clarify, I agree that x+w is not correct for rendering--if you're drawing an outlined rectangle.  If you're drawing a _solid_ rectangle (or blitting), then the geometric definition holds.

goobliata:
Yes, but we weren't talking about drawing sold boxes, so that's just random. You want a box to go to the edge of the pixel.

fatcerberus:
Ah, then yes, you're right.  Sorry, I thought the discussion was about calculating the bounding box (I.e. The mathematical definition), not drawing it.

For *drawing* the bounding box, assuming thickness 1, the correct coordinates to use are (x+0.5,y+0.5)-(x+w-1,y+h-1).  Spoken from experience. :)

goobliata:
No, you're still wrong. It's x + 0.5, x + w - 0.5;

goobliata:
What you said about "D3D9 aligns (0,0) with the center of the first pixel" is wrong too. 0,0 is the left edge of the first pixel.

elias-pschernig:
What goobliata says. I think we need to make http://liballeg.org/a5docs/trunk/primitives.html#pixel-precise-output more prominent, and probably strew some more examples into the documentation. E.g. al_draw_rectangble probably should have an example:

al_draw_rectangle(0.5, 0.5, 2.5, 2.5, 1)

With a picture how it draws a 3x3 square. Because obviously this is very confusing to people :)

fatcerberus:
Haha, sheesh, I can't even read my own code right.  I was basing my post above on this bit in minisphere:
https://github.com/fatcerberus/minisphere/blob/master/src/engine/screen.c#L686-L689

I subtract 1 there, but only because I've already added the 0.5 to x1.  I _do_ understand how this works, I swear! :P

As for D3D, I had run across this before:
http://drilian.com/2008/11/25/understanding-half-pixel-and-half-texel-offsets/

That seems to imply D3D puts (0,0) at the center of the pixel (unless I read it wrong, which is a distinct possibility today).  I assumed that's why Allegro needed to add the 0.5 translation matrix.

goobliata:
Yeah you read that wrong too, read again. Allegro add's 0.5 transformation because OPENGL pixels are 0, 0 at the center, and we need to be consistent.

fatcerberus:
Getting back to drawing bounding boxes, I sidestepped the whole +/- 0.5 mess in minisphere's standard library by tweaking the algorithm for thickness compared to Allegro, see here:
https://github.com/fatcerberus/minisphere/blob/sphere-2/assets/system/modules/prim.js#L127-L131

goobliata:
That's right but I find that to be written confusingly. No wonder why you were confused. I would use `x2 = x + w - t;` but that's just me.

fatcerberus:
I'll probably do that, yeah.  The way it's written now is an artifact of me originally truncating x and y to integers, so I just wrote the end points in terms of the already-normalized x1/y1.  I can change it now that I support floats.

PR #643, opened 2016-07-05
Make al_set_new_window_title work on OS X.

goobliata:


src/macosx/osxgl.m
```diff
@@ -1003,7 +1003,7 @@ +(void) initialiseDisplay: (NSValue*) display_object {
    [win setDelegate: view];
    [win setReleasedWhenClosed: YES];
    [win setAcceptsMouseMovedEvents: _osx_mouse_installed];
-   [win setTitle: @"Allegro"];
+   [win setTitle: [NSString stringWithUTF8String:al_get_new_window_title()]];
```
SiegeLord: This doesn't actually work, because this function is called on the main thread, rather than the user thread. This already affects the new window position. Currently that's done via two globals (`new_window_pos_x` and `_y`), which the top of the file admits is a hack.

Perhaps we could solve both this correctly here and now? We can just pass an NSArray pointer to this function and extract the display pointer, new title and position from it.

SiegeLord:
Merged as 52bd7db92f38c03f5baf8b169d0a6ee6f09c9c86.

PR #644, opened 2016-07-05
Linux new window title

goobliata:
The extra system locking that is now done on GTK that wasn't done before is unfortunate but I don't think it's a big deal.


SiegeLord:
This works great, thanks! We can merge this after the OSX half of this is fixed.

SiegeLord:
Merged as f12d7aa384f0d8455f03687ec10a1ea5f2b23407.

PR #645, opened 2016-07-09
lowp fragment shader improves performance a lot on mobile and

goobliata:
doesn't seem to affect anything.


SiegeLord:
Seems to work fine on my devices, and the internet seems to agree that using lowp for fragment shaders is standard practice.

PR #648, opened 2016-07-14
Rename LONG_LONG to AL_LONG_LONG

billyquith:
- LONG_LONG clashes with other 3rd party library.
- Please prefix Allegro types with AL_.

### Fix install

- This also fixes install as a header is missing from internal.

### DIFF
--- a/include/allegro5/inline/fmaths.inl
+++ b/include/allegro5/inline/fmaths.inl
@@ -110,7 +110,7 @@ AL_INLINE(al_fixed, al_fixsub, (al_fixed x, al_fixed y),
  *
  * PS. Don't move the #ifs inside the AL_INLINE; BCC doesn't like it.
  */
-#if (defined ALLEGRO_I386) || (!defined LONG_LONG)
+#if (defined ALLEGRO_I386) || (!defined AL_LONG_LONG)
    AL_INLINE(al_fixed, al_fixmul, (al_fixed x, al_fixed y),
    {
       return al_ftofix(al_fixtof(x) * al_fixtof(y));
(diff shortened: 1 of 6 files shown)


elias-pschernig:
Maybe we can just use int64_t instead?

include/allegro5/inline/fmaths.inl
```diff
@@ -110,20 +110,20 @@ AL_INLINE(al_fixed, al_fixsub, (al_fixed x, al_fixed y),
...
-      LONG_LONG lx = x;
-      LONG_LONG ly = y;
-      LONG_LONG lres = (lx*ly);
+      AL_LONG_LONG lx = x;
+      AL_LONG_LONG ly = y;
+      AL_LONG_LONG lres = (lx*ly);
 
       if (lres > 0x7FFFFFFF0000LL) {
-	 al_set_errno(ERANGE);
+     al_set_errno(ERANGE);
```
SiegeLord: Indentation got messed up here.

SiegeLord:
Yeah, let's just use int64_t.

billyquith:
I don't know the history of this type name. This change fixes my issue. I think standardising types to stdint is a different problem, although one I agree with. Perhaps you could accept this and raise that as a different issue?

SiegeLord:
Yeah, that's fair. Merged as a779890. Thanks!

billyquith:
Taking a quick look at the code, I think this define is a compiler patch for missing types, e.g. BCC. Changing it might break other things. 

There also seem to be other defines which could clash with user code (e.g. `INLINE`). Perhaps a review of the API wrt defines and types would be advantageous. Allegro 5.2 is becoming a very nice API, although some bits seem a little over engineered/confusing, some simplification (stdint, define tidy/reduction) would help clarity and avoid these clashes in the future.

Thanks for your efforts.  👍

beoran:
Some older compilers don't support stdint. Are we sure we want to drop
support for those?
On 16 Jul 2016 6:36 pm, "SiegeLord" <notifications@github.com> wrote:

Yeah, let's just use int64_t.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/liballeg/allegro5/pull/648#issuecomment-233138766>, or mute
the thread

billyquith:
Don't you already have a wrapper for that? Couldn't you just add stdint types where they are missing? You used to have to do this for MSVC because it was missing. At least it gives a standard for the type names. Once you have this it might simplify other parts of the header configuration.

beoran:
True, we could supply our own stdint header if unavailable. Although this
might again bring other complications. But yes, I guess it's nicer to be
able to use stdint than Allegro specific integer types.

fatcerberus:
For what it's worth, we already do something similar for stdbool, using the standard header when available and providing a custom one if needed.

PR #650, opened 2016-07-16
Make al_set_new_window_title work on OSX

SiegeLord:


SiegeLord:
This is a fixed version of #643.

PR #652, opened 2016-07-17
OSX mouse movement and warp fixes.

SiegeLord:
@goobliata, could you test this on your OSX machine? In particular, ex_mouse_warp should work and ex_mouse_events should still register the mouse leaving the display.


goobliata:
It mostly works, but there is one problem. In ex_mouse_warp, when right clicking normally the red line will go where your mouse would be. On OS X, it initially goes to about the opposite of where your mouse would be (e.g., if your mouse is top left, right clicking puts the line bottom right) and then the line never moves.

goobliata:
Actually I never actually checked it out, let me try it again...

goobliata:
It works now that I'm using the right code. I don't see any problems.

SiegeLord:
Thanks for testing this.

PR #653, opened 2016-07-17
Add al_backup_dirty_bitmap and al_backup_dirty_bitmaps to give

goobliata:
the user some control over when their bitmaps get backed up
to system memory.


fatcerberus:
> Since: 5.2.2

Should that be 5.2.1, or have the binaries for 5.2.1 been built (but not released) already?

goobliata:
You're right, I've changed it to 5.2.1.

goobliata:
Anybody have any problem with the API here? I've tested this on Windows/Android and it works same as before.

SiegeLord:
No complaints from me, although I'll probably want to make this API unstable for a bit. I can do this myself afterward though.

PR #654, opened 2016-07-17
Don't stop and join the timer thread when stopping the last timer.

goobliata:
On Windows were al_rest is very inaccurate, al_stop_timer could
cause a long unwanted delay when joining the thread before this.


SiegeLord:
We should probably add a condition variable inside that thread that waits until _al_vector_size goes above 0 so that we can suspend that thread entirely.

goobliata:
I did that, just have to test it when I get back to my Windows machine.

goobliata:
Didn't find any problems on Windows, Mac or Linux in my games.

SiegeLord:
This looks great, I'm glad we could fix this without adding more API.

PR #655, opened 2016-07-18
Move macro READ3BYTES/WRITE3BYTES into memdraw.c as only used there

billyquith:
This change moves a few macros into an implementation file which the only place they are used. This avoids use API pollution with macros.


billyquith:
Polluted with previous commits.

billyquith:
They are used in internal header on a platform other than mine.

PR #656, opened 2016-07-18
Move macros to implementation file & namespace other macros.

billyquith:
- `READ3BYTES` macro prefixed with `AL_` 
- `bmp_*` macros moves to only file they are used.


SiegeLord:
This looks great, thanks. Could you change the prefix for those macros to `_AL` instead, as they're in principle private.

billyquith:
There we go. Apologies for the mess. I will use the branch workflow next time!

SiegeLord:
No worries, thanks for these changes! Merged as  349cedc.

PR #657, opened 2016-07-22
Add ALLEGRO_GLYPH structure and al_get_glyph.

goobliata:
The structure contains all the info needed (including the bitmap)
to draw text. The ttf addon's render_glyph was used to make the
vtable function for the ttf addon.

This should allow the user the draw their own text. This hasn't
been tested much, I'm just looking for feedback on the API
right now.

### DIFF
--- a/examples/ex_ttf.c
+++ b/examples/ex_ttf.c
@@ -61,6 +61,8 @@ static void render(void)
     ALLEGRO_USTR *tulip = al_ustr_new("Tulip");
     ALLEGRO_USTR *dimension_text = al_ustr_new("Tulip");
     ALLEGRO_USTR *vertical_text  = al_ustr_new("Rose.");
+    ALLEGRO_USTR *dimension_label = al_ustr_new("(dimensions)");
+    int prev_cp = -1;
 
     al_clear_to_color(white);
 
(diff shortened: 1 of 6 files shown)


SiegeLord:
I think it'd be cleaner to stick the `advance` into the glyph structure itself. Additionally, that'd let you return a bool to say if glyph has been found or not.

goobliata:
I moved advance into the structure, documented things and used it in ex_ttf.

examples/ex_ttf.c
```diff
@@ -81,7 +83,17 @@ static void render(void)
        x += al_get_glyph_advance(ex.f2, cp, ALLEGRO_NO_KERNING);
     }
     al_draw_line(50.5, y+0.5, x+0.5, y+0.5, red, 1);
-    al_draw_textf(ex.f2, black, x + 10, y, 0, "(dimensions)");
+
+    for (index = 0; index < al_ustr_length(dimension_label); index++) {
+       int cp = ustr_at(dimension_label, index);
+       ALLEGRO_GLYPH g;
+       if (al_get_glyph(ex.f2, prev_cp, cp, &g)) {
+       printf("YIP\n");
```
SiegeLord: This is probably leftover debugging code, right?

goobliata: Yip.

goobliata:
So besides those changes which I've made, how does it look? It passes all font tests. I'm open to suggestions on making it just right before committing.

SiegeLord:
I think it looks fine. I think the only concern I have is that by using a by-value struct, we'll be making it very tricky to ever add fields to it if we for some reason decide to return more information.

But I don't think that concern is big enough to fix now, we can just add it as unstable and think about it some more at a later point.

Maybe one change that should be made is that we should require that the struct is zeroed out before being sent to the function? This'll help us upgrade the API in the future.

PR #661, opened 2016-08-06
embed qt and allegro in d3d

vkensou:
I post a topic on allegro forum [https://www.allegro.cc/forums/thread/616419](url).

It only work in d3d. I don't know how to resize the display in opengl.

It works like this:
p


include/allegro5/platform/aintwin.h
```diff
@@ -92,6 +92,9 @@ struct ALLEGRO_DISPLAY_WIN
    int toggle_w;
    int toggle_h;
 
+   //allegro don't manager the display
```
SiegeLord: "manage"

include/allegro5/platform/aintwin.h
```diff
@@ -92,6 +92,9 @@ struct ALLEGRO_DISPLAY_WIN
...
+   bool aldontmanagerdisplay;
```
SiegeLord: Could you rename this to `dont_manage_display` or alternatively, flip call it `manage_display` and flip all the conditions in the code?

src/win/d3d_disp.cpp
```diff
@@ -95,6 +95,7 @@ typedef struct D3D_DISPLAY_PARAMETERS {
    int window_x, window_y;
    /* Not owned. */
    const char *window_title;
+   int exist_window;
```
SiegeLord: `existing_window` would make it clearer what this is. Could this be HWND as well instead of an int?

include/allegro5/display.h
```diff
@@ -69,6 +69,7 @@ enum ALLEGRO_DISPLAY_OPTIONS {
    ALLEGRO_SUPPORTED_ORIENTATIONS = 32,
    ALLEGRO_OPENGL_MAJOR_VERSION = 33,
    ALLEGRO_OPENGL_MINOR_VERSION = 34,
+   ALLEGRO_USE_EXISTDISPLAY = 35,
```
SiegeLord: Maybe call it... ALLEGRO_USE_EXISTING_WINDOW?

SiegeLord:
This is interesting, have you tested how this handles display resets? E.g. start your program, press ctrl-alt-delete, cancel out of that and then see if your program still works?

vkensou:
Oh... I didn't handle this. I'll fix this.

vkensou:
I made a new commit. Now it could handle device lost. But there are some limits. If let allegro use existing  window, you must handle device lost and resize window by yourself. So people must create their own qt or mfc widget. I made a qt widget to help others, so where can i upload it?

SiegeLord:
How big is this widget? If it's small enough, we could make a `ex_qt` example that uses this.

vkensou:
The widget is not big. Just 2 files and less than 200 lines. But it dependence qt.

SiegeLord:
That's fine, we have an `ex_ogre` that depends on Ogre3D. It'll be an optional dependency.

vkensou:
So, should I upload a ex_qt.cpp? But I don't know how to use cmake.

vkensou:
I upload 3 new files. Compile and run succeed in qt 5.5.

vkensou:
There is another problem. My widget can't render bitmap unless use ALLEGRO_MEMORY_BITMAP. I am fixing it....

vkensou:
lol   
I have fixed that problem. Now my widget could render bitmap normally.
I added a init() method in QAllegroCanvas.    So we can load bitmap or something in this method.

Don't load bitmap in QAllegroCanvas' Constructor. I call init() in showEvent(), and it worked.

goobliata:
This could get the window size from the HWND and use it rather than the values passed to al_create_display. Those would be ignored kind of like they are for FULLSCREEN_WINDOW.

vkensou:
You mean ignore FULLSCREEN_WINDOW when use exist window?

goobliata:
No I meant do you think it would be better if the patch determined the size of the ALLEGRO_DISPLAY based on the HWND the user uses?

vkensou:
I have make a new commit. It worked.
```
m_display = al_create_display(0, 0);
int w = al_get_display_width(m_display);
int h = al_get_display_height(m_display);
```
The code will get right width and height.

goobliata:
Thanks for changing that!

vkensou:
A new commit!
Fix QAllegroCanvas can't init al_display's size correctly.

PR #662, opened 2016-08-14
OSX joystick enumeration changes

SiegeLord:
This is a revival of the currently reverted OSX changes, so we don't forget about them. There's some issue with the XBox controllers.

(cherry picked from commit 0828211c7d1b1e4647efe228c5b6e08e49f16bac)

### DIFF
--- a/src/macosx/hidjoy.m
+++ b/src/macosx/hidjoy.m
@@ -50,8 +50,11 @@ typedef enum {
 #define JOYSTICK_USAGE_NUMBER      0x04
 #define GAMEPAD_USAGE_NUMBER       0x05
 
+#define JOYSTICK_NAME_MAX           256
+
 typedef struct {
    ALLEGRO_JOYSTICK parent;
+   char name[JOYSTICK_NAME_MAX];
    IOHIDElementRef buttons[_AL_MAX_JOYSTICK_BUTTONS];
    IOHIDElementRef axes[_AL_MAX_JOYSTICK_STICKS][_AL_MAX_JOYSTICK_AXES];
    IOHIDElementRef dpad;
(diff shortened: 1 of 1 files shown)


fatcerberus:
Does the bug only affect OS X/macOS?  I'm using an Xbox One controller on Windows with Allegro 5.2.1 and it seems to work fine.

SiegeLord:
This is a re-introduction of the change that got reverted, so there's no bug per se. But either way, yes, it only affected OSX.

goobliata:
What needs to be done here? I recall some joysticks worked better with this code than the reverted code... what was it that did/didn't work then/now? I might have a joystick somewhere that exhibits the problems.

SiegeLord:
Here's the conversation that caused the reversion: https://github.com/liballeg/allegro5/commit/0828211c7d1b1e4647efe228c5b6e08e49f16bac It appeared that XBox controller didn't work for you with this patch :).

goobliata:
I've since noticed that even Windows messes up my joysticks. I'll have a trigger where an axis is supposed to be etc. Right now all I have is Xbox and a Logitech F710 (which is like an Xbox but has a directinput mode too) and a couple bluetooth controllers. I'll take a look how they work across platforms but it'll probably be next week.

fatcerberus:
Don't know if this is your issue, but I vaguely recall that Windows normally treats the L and R triggers on an Xbox controller as a single axis.  Each trigger cancels out the other so that it's impossible to tell if both are pressed simultaneously.

goobliata:
The issue I had was the right stick showed up as a trigger plus one axis of a stick.

goobliata:
I had a look at this. I can't do a whole lot because I don't have any of the type of joysticks that Todd Cope was using (with multiple joysticks on one device.) I dunno if it's correct but I have a hunch we need to check the Collection type when enumerating elements and not use all collections. I made some changes to this pull request here: https://github.com/goobliata/allegro5/tree/osx_joysticks. I found if I check for Application collections (Logical collections works as well) it'll work with my gamepads. The buttons are ordered oddly though, so I also sort them by usage which gives the same layout as before Todd's patch.

Anyway, not sure this is much use unless someone can test/fix it with one of those Wii U controllers or something similar.

NewCreature:
I tested the version of the patch located here: [https://github.com/goobliata/allegro5/tree/osx_joysticks](url).

It is working perfectly with my Mayflash Wii U Pro Controller adapter. Each controller is detected as a separate joystick in Allegro as expected.

PR #667, opened 2016-08-18
Add WANT_GLES3 CMake option

goobliata:
This hides multisampled render surfaces behind WANT_GLES3 on Android.


goobliata:
Any objections to this? It fixes the problems I mentioned on Kindle Fire tablets and probably others, and shouldn't be too difficult to use -DWANT_GLES3=on.

elias-pschernig:
Right now if multi-sampling is not supported (because the OpenGL extension is missing, which is quite common) it fails silently - I don't think we need a debug message in the GLES3 case either. I'm just always a bit afraid of those messages we print every frame.

goobliata:
OK I did error checking in case of missing support for multisampling.

elias-pschernig:
Thanks! But I meant, the failing silently is good, so just remove the debug messages. Multi-sampling in my game is not a critical feature, I just want it on when it is available as the game looks much better. With your change now if I turn on multi-sampling and compile without GLES3 it just breaks, right? Or maybe I misunderstand what you did with that bool return value.

I figure what I have to do eventually is actually this:
https://developer.android.com/guide/topics/graphics/opengl.html#version-check

I.e. check the version at run-time. Because I can't really disable GLES3 at compile-time, but I need the game to work on GLES2 only devices, the only way is a runtime check.

Anyway, I'll implement that once I get to it, so I'm fine with applying your commit, but please clarify what the new error check is for.

goobliata:
Yeah that makes sense. I reverted those changes, kept the debug prints out and kept the warnings fixed.

PR #669, opened 2016-08-25
Disable DEPTH_TEST Render State Before Drawing HUD

1Hyena:
When playing around with the ex_camera example I ran into a weird bug that made the checkerboard to be drawn on top of the debug texts in the upper left corner of the screen. This happened when the control style was switched to the spaceship or airplane mode and I moved the camera halfway inside the checkerboard. Since I debugged the situation for quite some time I figured that I should better commit a fix so that others would not waste their time on debugging the same exact issue in the future.


goobliata:
Thanks for the fix!

PR #674, opened 2016-09-08
Fix a crash when activity is destroy on Android.

goobliata:
The crash occurred when onDestroy was called before main exited.
This can happen when memory is low and possibly other instances.
The only way I could reproduce it was setting "Don't keep activities"
in developer options. Previously, games would crash as soon as
you closed them with this option on. Presumably the activity isn't
in a complete state when onDestroy is called like this: the crash
occurred in android_cleanup at the calls to finish_activity and
DetachThread. The exit(0) is needed or else resuming (restarting)
the game will not work.

Some ALLEGRO_DEBUGs were also removed for good measure: anything
after Allegro has been shut down or the activity is dying has
been removed.

### DIFF
--- a/src/android/android_system.c
+++ b/src/android/android_system.c
@@ -120,6 +120,8 @@ static void android_cleanup(bool uninstall_system)
       return;
    }
 
+   already_cleaned_up = true;
+
    if (uninstall_system) {
       /* I don't think android calls our atexit() stuff since we're in a shared lib
          so make sure al_uninstall_system is called */
@@ -129,8 +131,6 @@ static void android_cleanup(bool uninstall_system)
    finish_activity(_al_android_get_jnienv());
 
    (*javavm)->DetachCurrentThread(javavm);
-
-   already_cleaned_up = true;
 }
 
 static void *android_app_trampoline(ALLEGRO_THREAD *thr, void *arg)
@@ -362,6 +362,7 @@ JNI_FUNC(void, AllegroActivity, nativeOnResume, (JNIEnv *env, jobject obj))
    }
 }
 
+/* NOTE: don't put any ALLEGRO_DEBUG in here! */
 JNI_FUNC(void, AllegroActivity, nativeOnDestroy, (JNIEnv *env, jobject obj))
 {
    (void)obj;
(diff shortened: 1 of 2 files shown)


goobliata:
@elias-pschernig, perhaps you could test this with your Android game?

goobliata:
I also removed the call to atexit on Android in this patch. al_uninstall_system is called manually on Android.

goobliata:
I've found this only fixes half the problem. I'm working on the other half...

MaxSavenkov:
Got the same crash on Xiaomi Redmi 3S (but not on any other phone we have). Happens every time the app goes into background and comes back. For some reason, their Android variant tries to re-create activity by destroying & creating it again, and it all goes to hell in onDestroy. One thing I can see is that on this Xiaomi the app rotates when it goes into background for some reason, so when it comes back to foreground, it tries to rotate back to proper landscape variant, and here onDestroy happens.

goobliata:
Hello Max. I've found the previous patch a while ago to destroy the surface in onPause causes crashes sometimes, especially when you enable "Don't keep activities" mode in developer options. You had previously suggested a similar patch so please check this pull request as it has changed if you have a chance as it undoes that patch.

When I applied the patch in question, it was because music would start as the game would go through onChange and get a RESUME_DRAWING. Well I had to change my games and the way they resumed music to fix this with this patch removed.

elias-pschernig:
I applied the pull request as it is right now and I think it improves behavior, so please apply :) Without the patches I don't get crashes, but sometimes when switching in and out of my game it would end up in a messed up state. My guess is because the surface was destroyed in onPause and failed to properly recreate in onResume (sometimes you get 10 pause/resume pairs within a few milliseconds on Android - so never should do something time consuming like re-creating the surface anyway).

With the patch it haven't been able to recreate the issue anymore. Unfortunately I'm very low on test devices at home so that doesn't mean too much...

goobliata:
Thanks for testing!

elias-pschernig:
Thanks for fixing things! :D

goobliata:
For reference NVIDIA closed the bug with SHIELD TV crashing so everything appears fine now.

PR #675, opened 2016-09-08
Fix to ex_camera: camera's axes are now properly normalized

1Hyena:
This fixes a bug that used to freeze the camera when the user rotated it at a very rapid pace. Attached are screenshots of the bug.
ex_camera_1
ex_camera_2

### DIFF
--- a/examples/ex_camera.c
+++ b/examples/ex_camera.c
@@ -103,15 +103,14 @@ static void camera_rotate_around_axis(Camera *c, Vector axis, double radians)
    ALLEGRO_TRANSFORM t;
    al_identity_transform(&t);
    al_rotate_transform_3d(&t, axis.x, axis.y, axis.z, radians);
-   al_transform_coordinates_3d(&t, &c->xaxis.x, &c->xaxis.y, &c->xaxis.z);
    al_transform_coordinates_3d(&t, &c->yaxis.x, &c->yaxis.y, &c->yaxis.z);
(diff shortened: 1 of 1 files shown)


elias-pschernig:
Instead of normalizing the third vector, would it be better to set it as the cross product of the first two?

1Hyena:
@elias-pschernig you are right. By using cross product here it becomes trivial to also maintain perpendicularity of the axes. I have made the respective changes to this branch.

1Hyena:
@elias-pschernig 
Actually this can be optimized even more like this:

```
al_transform_coordinates_3d(&t, &c->zaxis.x, &c->zaxis.y, &c->zaxis.z);
al_transform_coordinates_3d(&t, &c->yaxis.x, &c->yaxis.y, &c->yaxis.z);

c->zaxis = vector_normalize(c->zaxis);
c->xaxis = vector_cross_product(c->yaxis, c->zaxis);
c->xaxis = vector_normalize(c->xaxis);
c->yaxis = vector_cross_product(c->zaxis, c->xaxis);
```

Unfortunately I can test this code and make a new commit on Monday the earliest. I'd suggest waiting until then before merging.

1Hyena:
I also made the skybox move with the camera's y-position because in airplane or spaceship mode it was possible to fly through the skybox. @elias-pschernig let me know if there is anything else you think needs modification here before merging.

elias-pschernig:
I always found it funny that I could reach the top of the skybox, but I agree that for the sake of being an example it should be a real skybox where this is not possible. Applied it all, thanks!

PR #686, opened 2016-09-27
add two Android convienience functions

elias-pschernig:
al_android_get_jni_env: returns Allegro's JNI environment
al_android_get_activity: returns Allegro's Activity

### DIFF
--- a/docs/src/refman/platform.txt
+++ b/docs/src/refman/platform.txt
@@ -123,6 +123,35 @@ Android platform that the calling Allegro program is running on.
 
 Since: 5.1.2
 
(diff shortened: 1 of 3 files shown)


elias-pschernig:
@goobliata Is this a good idea? In my game I need to call into Java for some Google API integration stuff, and exposing these two functions saves me from duplicating some code Allegro already has. I suppose in your games you also call into Java somehow?

goobliata:
Yes, this was once on my todo but I forgot about it. ATM I'm using the internal functions. :)

PR #687, opened 2016-09-28
cmake build should never litter in the git folder

elias-pschernig:
This calls ant inside the build folder instead for the Android example.
Also added my android build folders to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,10 @@
 
 # ignore the build/ directory
 build/
+build_android_arm/
+build_android_arm64-v8a/
+build_android_armeabi-v7a/
+build_android_mips64/
+build_android_mips/
+build_android_x86/
+build_android_x86_64/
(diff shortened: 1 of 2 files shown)


elias-pschernig:
Btw. Google says on https://developer.android.com/studio/tools/help/android.html that:
> This tool is no longer supported. You should instead use Android Studio to create AVDs and create projects.

Android Studio projects are extremely simply and use the Gradle build system. So I think we probably want to change our make_android_project.py script to create an android gradle project instead of an unsupported ant project, and do the same with the example.

goobliata:
I'm OK with this as long as gradle supports aar libraries (I think they're actually MEANT for gradle so shouldn't be a problem) as well as the old ant type of libraries (a directory with lib src res etc) or at least an easy way to make those ant libraries into aar. Reason is for things like Google SDKs and Amazon SDKs you need to use different types of libraries.

PR #688, opened 2016-10-30
Fix transform docs typo.

rcorre:
al_use_transform takes a pointer, just like the other transform
functions. Depict this correctly in the docs.


SiegeLord:
Thanks!

PR #690, opened 2016-11-06
Allow 32-bit build on OSX

MarcusCalhoun-Lopez:
On 64-bit systems, NSPoint is an alias of CGPoint.
NSPoint is an alias of CGPoint if any of the following are true:
__LP64__, TARGET_OS_EMBEDDED, TARGET_OS_IPHONE, TARGET_OS_WIN32, or
NS_BUILD_32_LIKE_64.
See
/System/Library/Frameworks/Foundation.framework/Headers/NSGeometry.h.
If NSPoint and CGPoint are different data types, convert between them.

### DIFF
--- a/src/macosx/qzmouse.m
+++ b/src/macosx/qzmouse.m
@@ -342,7 +342,11 @@ static bool osx_set_mouse_xy(ALLEGRO_DISPLAY *dpy_, int x, int y)
 
       CGPoint point_pos = CGPointMake(x, y);
 #if MAC_OS_X_VERSION_MIN_REQUIRED >= 1070
+#if __LP64__ || TARGET_OS_EMBEDDED || TARGET_OS_IPHONE || TARGET_OS_WIN32 || NS_BUILD_32_LIKE_64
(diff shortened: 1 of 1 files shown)


goobliata:
Can you simply add

```typedef CGPoint NSPoint;```

Somewhere? C allows a duplicate typedef.

goobliata:
If not I'd still prefer not having that pile of `#ifdefs` there. Perhaps just use the second block always.

MarcusCalhoun-Lopez:
NSPoint is declared in a System header file (/System/Library/Frameworks/Foundation.framework/Headers/NSGeometry.h)
On 64-bit systems,
`typedef CGPoint NSPoint;`
On 32-bit systems,
`typedef struct _NSPoint {
    CGFloat x;
    CGFloat y;
} NSPoint;`
So on 32-bit systems, I would think attempting a duplicate typedef would cause a conflict.

Always using the second block seems fine.
I am very new to GitHub, so please forgive me while I attempt to modify the pull request.

goobliata:
Thanks!

Issue #691, opened 2016-11-09
Python 3 bindings


gjvnq:
I was trying to use the allegro5 python bindings on my Arch Linux laptop and I came across two problems:

  * These bindings are not compiled by default.
  * There were a few syntax errors when I used the binding generation scripts.

The [diffs](https://github.com/liballeg/allegro5/files/581816/diffs.txt) are attached.

goobliata:
You'll probably want to keep the bindings off by default. You can enable them by passing -DWANT_PYTHON_WRAPPER=on to CMake. About the other changes, I don't know python but if nobody else comments I'll test them out myself and see.

SiegeLord:
The two argument `str`/`bytes` don't exist in Python 2, so we can't use them. You'll want to use the `encode`/`decode` methods which work on both. Also, would you mind making a pull request with these changes?

gjvnq:
No, I would not mind.

I will try to make the pull request as soon as I can.

PR #694, opened 2016-11-27
Python 3 bindings #691

gjvnq:
Solves issue #691 by making the code compatible with both python2 and python3.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,7 +147,7 @@ option(WANT_MONOLITH "Include all addons in the main library" off)
 # Wrappers.
 #
 
-option(WANT_PYTHON_WRAPPER "Enable generation of the Python wrapper" on)
+option(WANT_PYTHON_WRAPPER "Enable generation of the Python wrapper" off)
 
 #
 # Documentation.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,7 +147,7 @@ option(WANT_MONOLITH "Include all addons in the main library" off)
 # Wrappers.
 #
 
-option(WANT_PYTHON_WRAPPER "Enable generation of the Python wrapper" off)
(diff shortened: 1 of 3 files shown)


elias-pschernig:
What happens if someone has no Python installed? And what is the minimum version of Python 2 that is required now?

gjvnq:
I've used Python 2.7.12 and Python 3.5.2 for testing. So, I imagine that Python 2.7 is the minimum version.

However, I simply forgot the possibility of no python being available. The simplest fix is to keep the line 150 of the file CMakeLists.txt as it is (i.e. not building python bindings by default)

The best solution, in my opinion, would be to automatically install the python bindings only if python is available and simply ignore/not build the bindings if it python is not available.

beoran:
I would prefer if the Python bindings stay not built by default as they are
now. Even though Python is installed on my system, I have no use for the
bindings. So I don't want to have wait for them to compile.

Kind Regards, B.

On 27 Nov 2016 10:37 pm, "Gabriel Queiroz" <notifications@github.com> wrote:

I've used Python 2.7.12 and Python 3.5.2 for testing. So, I imagine that
Python 2.7 is the minimum version.

However, I simply forgot the possibility of no python being available. The
simplest fix is to keep the line 150 of the file CMakeLists.txt as it is
(i.e. not building python bindings by default)

The best solution, in my opinion, would be to automatically install the
python bindings only if python is available and simply ignore/not build the
bindings if it python is not available.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/liballeg/allegro5/pull/694#issuecomment-263149392>, or mute
the thread

gjvnq:
@eoran, I reverted `WANT_PYTHON_WRAPPER` to the previous state (i.e. `off`).

beoran:
Thanks. I will leave it to someone else to merge this, though...

SiegeLord:
Thanks!

PR #704, opened 2017-01-10
Improvements to save dialogs on systems using Gtk

oitofelix:
These are some improvements to file save dialogs on systems using Gtk.  Developing [MININIM](http://oitofelix.github.io/mininim/) and getting [feedback from users](http://forum.princed.org/viewforum.php?f=127) led me to them.


oitofelix:
Rebranched head fork to pull request #705.

PR #705, opened 2017-01-10
Improvements to save dialogs on systems using Gtk

oitofelix:
These are some improvements to file save dialogs on systems using Gtk.  Developing [MININIM](http://oitofelix.github.io/mininim/) and getting [feedback from users](http://forum.princed.org/viewforum.php?f=127) led me to them.

(rebranched head fork from previous pull request #704)

### DIFF
--- a/addons/native_dialog/gtk_filesel.c
+++ b/addons/native_dialog/gtk_filesel.c
@@ -46,6 +46,10 @@ static gboolean create_gtk_file_dialog(gpointer data)
 
    _al_gtk_make_transient(display, window);
 
+   if (save)
+     gtk_file_chooser_set_do_overwrite_confirmation
+       (GTK_FILE_CHOOSER(window), true);
+
    if (fd->fc_initial_path) {
      if (fd->flags & ALLEGRO_FILECHOOSER_FILE_MUST_EXIST)
        gtk_file_chooser_set_filename
--- a/addons/native_dialog/gtk_filesel.c
+++ b/addons/native_dialog/gtk_filesel.c
@@ -47,8 +47,23 @@ static gboolean create_gtk_file_dialog(gpointer data)
    _al_gtk_make_transient(display, window);
 
    if (fd->fc_initial_path) {
-      gtk_file_chooser_set_filename(GTK_FILE_CHOOSER(window),
-         al_path_cstr(fd->fc_initial_path, ALLEGRO_NATIVE_PATH_SEP));
+     if (fd->flags & ALLEGRO_FILECHOOSER_FILE_MUST_EXIST)
+       gtk_file_chooser_set_filename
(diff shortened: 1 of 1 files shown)


addons/native_dialog/gtk_filesel.c
```diff
@@ -46,6 +46,10 @@ static gboolean create_gtk_file_dialog(gpointer data)
 
    _al_gtk_make_transient(display, window);
```
elias-pschernig: Should always use {} around an if block.

addons/native_dialog/gtk_filesel.c
```diff
@@ -47,8 +47,23 @@ static gboolean create_gtk_file_dialog(gpointer data)
...
 
    if (fd->fc_initial_path) {
-      gtk_file_chooser_set_filename(GTK_FILE_CHOOSER(window),
-         al_path_cstr(fd->fc_initial_path, ALLEGRO_NATIVE_PATH_SEP));
```
elias-pschernig: Should use {} around the if here as well.

addons/native_dialog/gtk_filesel.c
oitofelix: Personally I don't agree with the premise "should always use {} around an if block".  I find single statement blocks inelegant and unnecessary.  That's an stylistic issue.  In respect to regards like "it's error prone", my answer is: first, that's debatable; second I use Emacs, and its indentation is very clear e consistent; third, good C programmers should be very acquainted with that, making errors quite unlikely, not worth the wasteful non-elegant alternative.  Said that, of course, this is a pull request, so I should use the coding style of the host project.  What's the official convention for Allegro?

beoran: The convention is first and foremost to stick to the convention used by the file you are in, since unfortunately we're not completely consistent. We do generally tend to use 3 spaces for indents and use {} around if blocks, amongst others.

oitofelix: > [...] amongst others.

Is there any publicly available documentation on that?  I'd like to avoid future stylistic mismatches.

elias-pschernig:
What Beoran said. In your patch the indentation was messed up and with statements spanning several lines it just was very hard to see where the if block ended - no big deal either way, it's something the person applying the patch can easily fix. However there is a more serious issue I get when trying ex_native_filechooser with the patch applied:

`(ex_native_filechooser): Gtk-CRITICAL **: gtk_file_chooser_default_set_current_name: assertion 'impl->action == GTK_FILE_CHOOSER_ACTION_SAVE || impl->action == GTK_FILE_CHOOSER_ACTION_CREATE_FOLDER' failed`

Any idea how to fix that?

oitofelix:
Thank you.

PR #712, opened 2017-01-14
Place a copy of allegro5.cfg into the docs.

### LINKED ISSUE #702: document configuration
Right now we provide a sample allegro5.cfg which documents all possible settings. But going to the "Configuration files" chapter in the documentation does not mention any of those settings at all. I think it would be nice if we had a new section describing allegro5.cfg (and link to it from "Configuration files").

SiegeLord:
Fixes #702.

What do you think of this, @elias-pschernig?


elias-pschernig:
Does not seem to work for me... I applied it and rebuilt the docs, but no mention of allegro.cfg still.

SiegeLord:
I stuck it into `al_get_system_config` documentation. There's a listing of the default allegro5.cfg at the end (unless it didn't work :P).

elias-pschernig:
Oh, indeed, perfect!

PR #716, opened 2017-01-17
[color addon] Add XYZ, xyY, Lab and LCH color spaces.

allefant:


addons/color/color.c
```diff
@@ -575,4 +583,227 @@ uint32_t al_get_allegro_color_version(void)
...
+   return pow(x, 1 / 2.4) * (1 + a) - a;
+}
+
+
+/* Function: al_color_xyz_to_rgb
+ */
+void al_color_xyz_to_rgb(float x, float y, float z,
+    float *red, float *green, float *blue)
+{
+   double r = 3.2406 * x + (- 1.5372 * y) + (- 0.4986 * z);
```
SiegeLord: What's up with the weird space after the negation here?

examples/ex_color2.c
```diff
@@ -0,0 +1,348 @@
...
+}
+
+static void draw_range(int ci)
+{
+   float l = example.color[ci].l;
+   int cx = (example.color[ci].a * 511.0 / 2) + 511.0 / 2;
+   int cy = (-example.color[ci].b * 511.0 / 2) + 511.0 / 2;
+   int r = 2;
+   /*
+    * 1234561
```
SiegeLord: Hah! What's this?

allefant: I need to comment this. Basically we loop around the selected color in the center to find the pixels with a certain ciede2000 distance. We start around the center then go 1, 2 to the right. Then we go 1, 2 down. Then 1, 2 left. Then 1, 2 up. Then we jump to a bigger radius and start over with 1, 2, 3, 4 right, 1, 2, 3, 4 down... The diagram shows how exactly :)

addons/color/color.c
allefant: You clearly must not like Haskell :P

PR #717, opened 2017-01-19
Mac OS X 10.6 fixes

oitofelix:
- Fix build in Mac OS X 10.6. The default compiler provided by XCode 4.2 is llvm-gcc-4.2 which fails to compile modules for which gl_ext.h redefines type ALLEGRO_OGL_EXT_LIST. This isn't an issue for later versions of GCC which allow for such redefinitions. This fix is important, though, because that's the latest GCC officially released by Apple for such system, and I couldn't build an working Allegro using compilers provided by MacPorts, since although builds were successfully completed, Allegro systematically segfaulted at misaligned_stack_error system function for any program.

- Fix warnings "__NSAutoreleaseNoPool(): Object ? of class ? autoreleased with no pool in place - just leaking" when building or modifying menus under Mac OS X 10.6.

### DIFF
--- a/addons/native_dialog/osx_dialog.m
+++ b/addons/native_dialog/osx_dialog.m
@@ -119,6 +119,7 @@ void _al_shutdown_native_dialog_addon(void)
 bool _al_show_native_file_dialog(ALLEGRO_DISPLAY *display,
                                  ALLEGRO_NATIVE_DIALOG *fd)
 {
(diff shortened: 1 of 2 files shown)


SiegeLord:
Merged as 53a775e1c6aec1b0d67563a21682d9cf36c139fe, thanks!

Issue #721, opened 2017-01-21
Allegro doesn't update keyboard state when focus switch out from display


oitofelix:
While implementing an work around for issue #720, I came across another issue regarding key presses, in respect to ALLEGRO_EVENT_KEY_UP and ALLEGRO_EVENT_KEY_DOWN this time.

When the user calls al_get_keyboard_state, Allegro doesn't retrieve the real current state of the keyboard, but instead it returns a copy of its internal model which is maintained in a key press/release event driven-fashion.  Thus when Allegro detects a key down event, it marks the respective key as pressed until it receives the key up event for that very key, updating its internal state accordingly.  This means that Allegro reports a key as being pressed down if it has received a key down event, but not yet a key up one for that key.

The problem I'm facing is a result of Allegro's internal keyboard state getting out of sync with the real keyboard state.  This happens because Allegro is not always able to get a key release event for each key press event detected.  For instance, if Ctrl+L triggers a dialog box (which takes the keyboard focus out of the current display) and the user releases both keys, Allegro misses their respective key up events and never realizes both keys were released, reporting otherwise in subsequent calls to al_get_keyboard_state.  If Allegro had at least a function that allowed the user to influence its internal keyboard model, one could easily workaround this bug by forcing it to reset its state in every action likely to take the focus out of the current display.  Unfortunately there isn't any.

Thinking in retrospect now, this bug was one of the reasons I originally decided to use ALLEGRO_EVENT_KEY_CHAR instead for processing most of MININIM's key bindings.  Too bad I'm having difficulties both ways.

SiegeLord:
For reference, here's a discussion about this a few years ago: https://mail.gna.org/public/allegro-developers/2015-05/msg00006.html

I guess it just never got implemented.

beoran:
I'd like to implement that function, but I simply don't have much time left
to work on Allegro these days. All help would be. greatly appreciated...

oitofelix:
I started implementing it already.  I'll make a pull request soon.

oitofelix:
Done.  See pull request #722.

PR #722, opened 2017-01-22
Implement al_clear_keyboard_state function in order to remedy issue #721

oitofelix:
See issue #721.

**PS:** I didn't mark al_clear_keyboard_state with "Since: 5.2.3"


fatcerberus:
I suggested implementing something like this on the forums about a year ago and got all kinds of flak for it. :(
https://www.allegro.cc/forums/thread/615411

Oh well, good to see it finally got implemented anyway.

SiegeLord:
Nah, we discussed it on the mailing list (https://mail.gna.org/public/allegro-developers/2015-05/msg00006.html) and decided that it's probably fine, just needs to be implemented.

As for this PR, it's a fine start but I'll very likely add a boolean parameter to that function whether to send the KEY_UP events or not when it is called, for people that maintain their own state and would like to just receive the virtual key events instead of tracking the internal Allegro keyboard state, or doing manual reset logic. To be clear, you don't need to implement this, but if you start using the function right away be aware that it'll be modified a bit soon after this is merged.

oitofelix:
Optional key up batch event dispatch has been implemented, as suggested by @SiegeLord, with a minor modification: it has an ALLEGRO_DISPLAY * parameter (instead of a simple boolean) for a more general and consistent behavior, given that al_clear_keyboard_state is mainly intended to be used in response to ALLEGRO_EVENT_DISPLAY_SWITCH_OUT events, that carry a display source.  Furthermore, ALLEGRO_EVENT_KEY_UP events must provide a display anyway.

> `al_clear_keyboard_state (ALLEGRO_DISPLAY *display)`
>
> Clear the state of the keyboard, emitting ALLEGRO_EVENT_KEY_UP for each currently pressed key.  The given display is regarded as the one which had the keyboard focus when the event occurred.  In case display is NULL no event is emitted.  For most keyboard drivers Allegro maintains its own state of the keyboard, which might get out of sync with the real one.  This function is intended to remedy such situation by resetting Allegro's keyboard state to a known default (no key pressed).  This is particularly useful in response to ALLEGRO_EVENT_DISPLAY_SWITCH_OUT events.

oitofelix:
While reimplementing MININIM's keyboard handling code, I came across a bug in my implementation of al_clear_keyboard_state for the X11 driver.  It has been fixed in the latest commit.

oitofelix:
While testing MININIM on Mac OS X, I realized that I had forgotten to implement al_clear_keyboard_state backend for that system.  It has been implemented in commit 51247b7.

oitofelix:
Just for the record, I've tested MININIM's new keyboard handling code which makes use of al_clear_keyboard_state on GNU/Linux (Debian 8 and Ubuntu 16.04), Windows XP and Mac OS X 10.6, and it works as expected.

Is there anything else I can help with regarding al_clear_keyboard_state implementation before merging this PR?

gh0stwizard:
As I see there are left some issues with platform-specific code. For instance, static variable `modifiers` in file `src\win\wkeyboard.c` was not updated to 0 by `clear_keyboard_state(void)`.

The second notice is that `unichar` element of the struct `ALLEGRO_KEYBOARD_EVENT` was not updated by `al_clear_keyboard_state(ALLEGRO_DISPLAY *display)`. As I see `unichar` is filled by platform-specific code too. This is minor issue but anyway, just to note.

Everything else is cool, except some minor optimization of `al_clear_keyboard_state(ALLEGRO_DISPLAY *display)`:
```C
         /* because of _al_event_source_emit_event() makes a copy of event's struct
          * we fill-in elements only once and modify only valuable of them 
          * in the loop below. */
         ALLEGRO_EVENT event;
         event.keyboard.type = ALLEGRO_EVENT_KEY_UP;
         event.keyboard.timestamp = al_get_time();
         event.keyboard.display = display;
         event.keyboard.modifiers = 0;
         for (keycode = ALLEGRO_KEY_A; keycode < ALLEGRO_KEY_MAX; keycode++) {
            if (al_key_down(&ks, keycode)) {
               event.keyboard.keycode = keycode;
               event.keyboard.unichar = 0; /* TODO: platform-specific code */
               _al_event_source_emit_event(es, &event);
            }
         }
```
But may be I am wrong, just my thoughts.

SiegeLord:
The unichar field is only filled in ALLEGRO_KEY_CHAR events which are emitted contemporaneously to the ALLEGRO_KEY_DOWN events, so I think the omission is fine here. I'm pretty happy with the code as it is, I'll just do more tests this weekend (and check out that `modifiers` issue) and then it's good to go.

SiegeLord:
So I tested this on Windows, Linux and OSX and found the following oddities:

On OSX, the if you keep holding a key while switching in and out of the window, the first char event will be a repeat, while on Linux and Windows it will correctly be a non-repeat char event.

On Windows, if you keep holding shift + letter and then switch in and out, the modifiers will not be set correctly after a switch in (this isn't the same thing pointed out above). OSX and Linux handle this correctly.

In all places, you get unpaired KEY_UP events if you switch in with a held key... surely we can use the cleared state to detect them and not send them.

But either way, those are not blocking issues... I'll file a bug about them and we'll fix them as time goes on, while merging this meanwhile. Thanks for your hard work making this! Merged as aba9e6aa9dc436e635dfc887a71e00b53feb6dc0.

SiegeLord:
And here's the issue with the above observations: #728

PR #724, opened 2017-01-25
Fix al_remove_config_section and al_remove_config_key ASSERT crash

oitofelix:
Fix al_remove_config_section and al_remove_config_key ASSERT crash in al_ref_cstr, when invoked with a NULL section (which conventionally represents the global section).


elias-pschernig:
Thanks! I didn't know you can use NULL instead of "", but it does make sense to allow it.

PR #725, opened 2017-01-28
patch al_set_window_constraints() when the flag ALLEGRO_MAXIMIZED is set

gh0stwizard:
If a window (display) was created with the `ALLEGRO_MAXIMIZED` flag then call to `al_set_window_constraints()` will be a cause of strange behavior as depicted below:

a5_issue_maximized

The patch is fixing such problem.
Tested only under Windows 7, 64-bit.
Thanks!


gh0stwizard:
Additional notes:
1. Better to say if a display was created like that: `ALLEGRO_WINDOWED` | `ALLEGRO_RESIZABLE` | `ALLEGRO_MAXIMIZED` and not just `ALLEGRO_MAXIMIZED`.
2. Added an example to reproduce a bug. It may be wrong because I don't know yet how to find out the default adapter id for all platforms, I just hardcoded `0`. Just to clarify the moment.

elias-pschernig:
Should I not be allowed to constrain a window to a certain size while it is maximized? For example if I want my window to never be smaller than 100x100 but it is currently maximized. There may be a mismatch between the function name (al_set_window_constraints) and what it does (do a one-time resize if the size is outside the passed range)?

gh0stwizard:
Indeed. I have changed the code and now the new window is maximized and restricted to be never less than minimum values specified by `al_set_window_constraints()`.
Changes related only to Win32. See information about other platforms below:
* `raspberrypi_set_window_constraints()` is not implemented
* `xgtk_set_window_constraints()` is not implemented
* `xdpy_set_window_constraints_default()` probably is need a patch and testing. As I see, it contains a code to choose the appropriate width & height of the window, but this code does incorrect decision:
```C
   int w = glx->display.w;
   int h = glx->display.h;

   if (min_w > 0 && w < min_w) {
      w = min_w;
   }
   if (min_h > 0 && h < min_h) {
      h = min_h;
   }
   if (max_w > 0 && w > max_w) {  /* <-- how is this possible? resize "down"? */
      w = max_w;
   }
   if (max_h > 0 && h > max_h) {
      h = max_h;
   }

   /* Resize the display to its current size so constraints take effect. */
   al_resize_display(display, w, h);
```
Anyway, I have to build Allegro under GNU/Linux to see how is really working. This will take a time.
I had not found out anything about other platforms. May be I had missed something.

elias-pschernig:
I think if your window is already 1000x1000, then you do al_set_window_constraints(500, 500), it will resize the window down. Anyway, I have never used this function and I'm not clear what the purpose is. There is two general ideas:

A. Do what the name says and just set constraints which would restrict a future resize. This should then work even while maximized as the current size would not matter.

B. Only do an adjustment of the current size but don't restrict future resizes. This would of course make no sense while the window is maximized.

And I suppose both of those could be combined into one function which always does both. The documentation right now is not clear on which it is (but the name hints at just A) - so before changing any code we have to decide what the function _should_ do. Then document it and fix it if it is not working as it should.

gh0stwizard:
> I think if your window is already 1000x1000, then you do al_set_window_constraints(500, 500), it will resize the window down.

The `al_set_window_constraints()` contains four parameters: min_w, min_h, max_w, max_h. So, if I have a window 1000x1000 and then call `al_set_window_constraints(500, 500, 1000, 1000)`, what do we have to expect?
* A) should the window to be resized to 500x500
* B) or should it be left as is with 1000x1000

The current version of Allegro is choosing the variant A. Meanwhile, a human (i.e. me) must decide when I wish it or do not. I may call al_resize_display() if I need this. In case of maximized window I wish that Allegro has choose the variant B: either resize the window again to 1000x1000 or left it untouched.

But. There is one more issue. Under Win32 after un-maximizing such a window, the previously set upper limits starts working as intended (i.e. limiting and creating "broken" maximized windows). 

Also I see that for X11 there is special code exists to handle a maximized window in the file `src\x\xwindow.c`, function `void _al_xwin_maximize(ALLEGRO_DISPLAY *display, bool maximized)`, which sets correct maximum values for window _bypassing_ constraints. In comparison with Win32, where constraints are workine: there is possibility to create "broken" maximized windows after their re-maximizing. It is ok, because I can pass to `al_set_window_constraints()` correct values, but it should not touch the already maximized window (patch above does).

All of this leads to the point that currently the function `al_set_window_constraints()` resizing the window when it is needed and not needed. So, my point is that `al_set_window_constraints()` must set constraints only and never touch the size of the current display, no unexpected resizes. If somebody needs to resize a window it can handle it by themselves. 

In other words, the current implementation of the function `al_set_window_constraints()` is working only with lower limits (mins) and does not take into account the upper limits (maxs). Either more complicated logic must be created (and described) or keep it as simple as possible. We can resize the current window only and only when the current sizes are out of the specified constraints.

My case is simple:
```C
al_set_new_display_flags(ALLEGRO_WINDOWED | ALLEGRO_RESIZABLE | ALLEGRO_MAXIMIZED);
display = al_create_display(DISPLAY_W, DISPLAY_H);

/* I set constraints to lower limits only, but.... */

/* Quote from docs: */
/* A value of 0 for any of the parameters indicates no constraint for that parameter. */
al_set_window_constraints(display, DISPLAY_W/2, DISPLAY_H/2, 0, 0);
/* does not work */

/* OK: computer listen to my command! */
al_set_window_constraints(display, DISPLAY_W/2, DISPLAY_H/2, DISPLAY_W, DISPLAY_H);
/* does not work */
```

IMHO.

goobliata:
I think @gh0stwizard's summary looks good. If it's maximized then apply the constraints once it has been un-maximized. Otherwise apply to future resizes but do not resize right away. This could be an issue though. If you have a 1200x1200 window and set a max 1000x1000 size and then drag the window border, what happens? Perhaps the function should resize the window to fit in the max dimensions (ie, if it's 1200x800, size it to 1000x800.))

gh0stwizard:
I have found a real cause of this bug. There is no fear to call `al_resize_display()` inside of `al_set_window_constraints()` to apply new constraints (as its described in the comment) if... if width & height of the display have correct values when `ALLEGRO_MAXIMIZED` is used. Prototype patch is showed below.

`src\win\wwindow.c`
Function: `static LRESULT CALLBACK window_callback(HWND hWnd, UINT message,
   WPARAM wParam, LPARAM lParam)`

```C
/* we go a message from OS telling that ... */
switch (message) {
case WM_SIZE: /* the size of the window was changed. */
  if (wParam == SIZE_RESTORED || wParam == SIZE_MAXIMIZED || wParam == SIZE_MINIMIZED) {
       /*
	* Delay the resize event so we don't get bogged down with them
	*/
     if (!resize_postponed) {
	resize_postponed = true;
	_beginthread(postpone_thread_proc, 0, (void *)d);
    }
    d->flags &= ~ALLEGRO_MAXIMIZED;
    if (wParam == SIZE_MAXIMIZED) {
      d->flags |= ALLEGRO_MAXIMIZED;

      /* new code starts here. just update ALLEGRO_DISPLAY width & height to new ones */

	RECT wRect;
        RECT cRect;
        int wWidth;
        int wHeight;
        int cWidth;
        int cHeight;

        GetWindowRect(hWnd, &wRect);
        GetClientRect(hWnd, &cRect);
        wWidth = wRect.right - wRect.left;
        wHeight = wRect.bottom - wRect.top;
        cWidth = cRect.right - cRect.left;
        cHeight = cRect.bottom - cRect.top;
        ALLEGRO_DEBUG("changed resolution from w: %d, h: %d ---> w: %d, h: %d",
                                     d->w, d->h, cWidth, cHeight);
        d->w = cWidth;
	d->h = cHeight;
   } /* if (wParam == SIZE_MAXIMIZED)  */
}
```
After applying this patch the `al_set_window_constraints()` is working as expected even been unmodified.
Now, we have to decide what to do. What you would say?

goobliata:
We should either resize the window to fit in the constraints (eg if it's too small, make it bigger, if it's too big, make it smaller) or add a note in the documentation that you should make sure to resize the window before applying constraints if it's currently not within the constraints to avoid an inconsistent state. IMO the latter would be better because it's more consistent with maximized window behaviour and also fully backwards compatible.

gh0stwizard:
New patch applied. I have made it in the same manner as it does by `xdpy_set_window_constraints_default()` from `src/x/xdisplay.c` now:
* find out current window size via `GetClientRect()`
* compare values from `GetClientRect` with that user has defined, i.e min_w, min_h, max_w, max_w
* call `al_resize_display()` with w & h appropriate to their own ranges:
    * min_w < w <max_w (min_w > 0, max_w > 0)
    * min_h < h < max_h (min_h > 0, max_h > 0)

Results are:
* when called `al_set_window_constraints(display, min_w, min_h, 0, 0)` for a window with `ALLEGRO_MAXIMIZED`: the window left untouched, but `min_w`, `min_h` will be applied as constraints after window's restoring.
* when called `al_set_window_constraints(display, 0, 0, max_w, max_h)` for a window with `ALLEGRO_MAXIMIZED`: the window will be resized to `max_w`, `max_h` as intended.

And the rule from docs is working too: "A value of 0 for any of the parameters indicates no constraint for that parameter."

elias-pschernig:
I agree with goobliata, the latter option sounds better to me (mainly because of how the function is named so in my mind it should only set the constraints but not apply them). That is, al_set_window_constraints should never do a resize (even if the current size is smaller/larger). It would just prevent the window from ever getting resized to a size smaller than the minimum or larger than the maximum constraint in **future** resizes.

If you maximize a window I think that should just ignore the constraint altogether. Same with toggling fullscreen. If you don't want the window to be maximized or fullscreened - just don't enable the buttons for it. The constraints would only apply to the user resizing a user-resizable window by dragging its borders/corners.

goobliata:
I was trying to articulate this potential issue with what you said @elias-pschernig. It depends on how the windowing system works though which I don't know. And that is: say you have a window that's 1000x1000 and you then constrain it to 500x500. If the window isn't resized automatically, what's going to happen when you drag the border? Will it jump to 500x500? That's the tricky bit.

elias-pschernig:
If possible I'd like to leave that up to the underlying windowing system (assuming Windows, OSX and X11 all have a concept of size constraints). But I don't know either - would have to experiment. Basically if Allegro has to do a ton of work to "fight" the underlying window manager I think we are doing it wrong. Otherwise I'm really fine with any solution as long as it is documented.

gh0stwizard:
I assume that we may introduce a new flag to enable or disable automatic calls to `al_resize_display()` inside of `al_set_window_constraints()`. First, it will not break already existed code. The second, it give us a flexibility. But may be this is not good idea?

At the moment of writing this I see that there is still open an issue with how we must control upper limits for constraints. Let's assume the next case:

* we have a maximized window.
* we call `al_set_window_constraints(min_w, min_h, max_w, max_h)`; where *max_w* & *max_h* are lower than sizes of the maximized window.
* under Win32 this leads to creating "ugly" maximized window with update width/height (see picture in the first post). And later maximizations & de-maximizations keeps such state.
* under GNU/Linux this does nothing: maximized window stays maximized as before (if de-maximize such a window then constraints are working properly; and the next maximization will maximize the window _bypassing_ constraints to the first/original maximized state).
* under Mac OS: I don't know what's happening. I have to install Darwin, not sure if my PC is able to emulate it under vmware...

I am asking this because this is responsibility of `al_set_window_constraints()`. I am not sure what is best. Probably we have to provide a flexible way to control the behavior above too.

Anyway, I may code everything, just tell me what to do.

Current state of changes in the pull request: 
1. It is fixing the original issue under Win32
2. Added the example to testing.

goobliata:
I think to get consistent behaviour is going to require a lot of code that taps when the window gets maximized/unmaximized and applies constraints at that time. If we were to do that then the current Linux behaviour is obviously better to go with.

So basically if a window has constraints and it gets maximized, clear the constraints (but save the info) and then if/when it's unmaximized apply the saved info.

goobliata:
I forget what was happening with this, what's the status right now?

gh0stwizard:
@goobliata 

1. Well, the plan was that we will have the same behavior for major three: Win32, *nix (X11) and MacOS.
2. Currently, this patch is for Win32 only. And it repeats the behavior of X11 (*nix).
3. I could not found a time to see what is situation under Mac OS. So, the ticket is stalled at the moment.

It would be great if someone with Mac OS will perform tests and will answer to this topic how constraints are working with trunk or stable. The patch also includes a simple test program (it uses only one key "spacebar", also this is mentioned in comments of the program) for those who are lazy to write its own.

The behavior under X11 was described here: https://github.com/liballeg/allegro5/pull/725#issuecomment-275936826

goobliata:
I think I agree with Elias that al_set_window_constraints shouldn't ever resize the window. It sounds from what I'm reading that with this patch it will resize the window if it's maximized and you then set a max_w/max_h constraint? I think it should probably just ignore constraints when maximized (again like Elias said). To do this it might have to clear the constraints before maximizing and re-apply them afterwards.

If this is already the way it is, let me know, I may have misread. Once this is done I can check it on macOS.

gh0stwizard:
@goobliata 
>I think I agree with Elias that al_set_window_constraints shouldn't ever resize the window.

That's OK, because I am happy with both behaviors.

>It sounds from what I'm reading that with this patch it will resize the window if it's maximized and you then set a max_w/max_h constraint?

Yes, indeed.

>If this is already the way it is, let me know,

No, it does not. I will do my best ASAP.

goobliata:
I've tested this as-is on macOS. The current results are:

If I use 320x240 as min size, I can maximize/unmaximize and the constraint has no effects (makes sense.) If I set the min_w, min_h bigger than my screen, when I un-maximize it resizes to min_w, min_h (I used 5000x5000 and un-maximizing creates this huge window.) Also I can properly maximize the window with these large constraints, it takes the size of the screen.

With max_w, max_h set I cannot really maximize the window, clicking maximize expands the window to 640x480.

So case 1 (min_w, min_h) seems correct. Case 2 (max_w, max_h) needs to clear the constraints when you click maximize then re-apply them after un-maximizing.

goobliata:
I thought we were going to remove the al_resize_display and let the window manager handle it? You could still do it manually if you wanted the constraints applied right away. Seems like the most flexible way to do it.

gh0stwizard:
@goobliata 

>I thought we were going to remove the al_resize_display and let the window manager handle it?

This is not an issue. I assume we have to play with it when everything will work as expected. I don't very like remove `al_resize_display()`, because it may be a cause of strange behaviors, especially if one forgot to call `al_resize_display()` right after `al_set_window_constraints()`. You may try it already on win32.

There is another reason why I did not removed it. Meantime constraints on win32 may work without `al_resized_display()` due its architecture (I have tested it already), the X11 system will still use `al_resize_display()` due some hacks (until someone resolve them).

And thanks for report for MacOS. As I see the problem is the same as it was on win32.

goobliata:
I'm not completely against having an al_resize_display in there.

I had a look at it on Windows with this patch. It doesn't look like the window is maximized properly in ex_window_maximized? Here's a screenshot:

image 001

There is a border around the window that wouldn't be there if it was properly maximized.

gh0stwizard:
>I'm not completely against having an al_resize_display in there.

As I said, this is not an issue. At the moment I am trying to install MacOS because it's need a fix.

>I had a look at it on Windows with this patch. It doesn't look like the window is maximized properly in ex_window_maximized?

Thanks for report. It is not a bug of the patch. As I see, the bug exists in Allegro 5.2.2 also. This is caused by `WS_EX_OVERLAPPEDWINDOW` when `ALLEGRO_RESIZABLE` is set:

```c
static void display_flags_to_window_styles(int flags,
   DWORD *style, DWORD *ex_style)
{
   if (flags & ALLEGRO_FULLSCREEN) {
      *style = WS_POPUP;
      *ex_style = WS_EX_APPWINDOW;
   }
   else if (flags & ALLEGRO_RESIZABLE) {
      *style = WS_OVERLAPPEDWINDOW;
      *ex_style = WS_EX_APPWINDOW | WS_EX_OVERLAPPEDWINDOW;
   }
   else {
      *style = WS_CAPTION | WS_SYSMENU | WS_MINIMIZEBOX;
      *ex_style = WS_EX_APPWINDOW;
   }
}
```
I suppose it is better to create another patch for this, IMHO.

goobliata:
OK. If you can do the same for macOS great. I'd like for that to be done before I accept this, as the documentation changes would suggest differently than the current OS X behaviour. Thanks.

gh0stwizard:
Just want to tell that I have managed to install MacOS & all stuff needed for the development. Until today I was thought that it would be an easy cake, but now I see there are multiple issues related to windows sizing & placing under MacOS. And fixing them will consume some amount of time. Can't say how good and fast I would be, I hope I will done till the next weekend.

goobliata:
It's not a rush, I've got all the time in the world.

gh0stwizard:
I have fixed all issues related to window constraints & `ALLEGRO_MAXIMIZED` flag under MacOS. Sorry, that it took so long, I have got a new job and just did not have a time to complete this task.

Can you test it?

We can get rid of `al_resize_display()` as planned. All three platforms are ready to this step. So, if you will like the patch, then I will remove al_resize_display() and fix the docs too.

goobliata:
Looks good. Works fine for me on OS X. So if you remove the al_resize_display I'll merge it. Thanks!

goobliata:
I think it would be good to have a function that would resize a window to the constraints but as a separate function. So you could set the constraints and then if you want it applied right away call al_respect_constraints(display); (example function name).

gh0stwizard:
@goobliata 
>Looks good. Works fine for me on OS X. So if you remove the al_resize_display I'll merge it. Thanks!

Yes, I will do it. There is a need to change docs still.

>I think it would be good to have a function that would resize a window to the constraints but as a separate function. So you could set the constraints and then if you want it applied right away call al_respect_constraints(display); (example function name).

Good. Am I correct that the new function will do the same as does `al_set_window_constraints()` now? I mean, it will set new constraints and apply them right away by calling `al_resize_display()`.

goobliata:
> Good. Am I correct that the new function will do the same as does al_set_window_constraints() now? I mean, it will set new constraints and apply them right away by calling al_resize_display().

The idea was that it would just resize the window to the constraints, it wouldn't set any constraints. It would just use the existing ones.

gh0stwizard:
>The idea was that it would just resize the window to the constraints, it wouldn't set any constraints. It would just use the existing ones.

To resize a window to previously set constraints it would be enough to call:
```c
al_resize_display(display, al_get_display_width(display), al_get_display_height(display));
```

That is, if the current w & h of a window do not in a range of the constraints, then the window will be resized. I am not sure that such a simple thing should be implemented. At least it can be a macro. What do you think? Are we really need it?

Actually, I thought about this (without brackets to simplicity):
```c
#define al_respect_constraints(display, min_w, min_h, max_w, max_h) \
    if (al_set_window_constraints(display, min_w, min_h, max_w, max_h))
        al_resize_display(display, al_get_display_width(display), al_get_display_height(display));
```
Almost the same, just for another use case.

I think that if we clarify thee new behavior of `al_set_window_constraints()`. then developers will do it as they wish themselves.

gh0stwizard:
To summarize and that we all be agree, next goals are:

* `al_set_window_constraints()` just remembers user's constraints;
* a new function `al_apply_window_constraints(DISPLAY *display)` will apply new constraints (that is, pass them to OS level) and if there is a need resize a window;
* update documentation and tell about such changes;
* update examples (otherwise, they stop working).

Is "apply" better then "respect" in this case? I mean about a new function: `al_apply_window_constraints()` or `al_respect_window_constraints()`. I am not a native English speaker and it is hard to distinguish which is best to me.

goobliata:
Actually just forget about al_apply_window_constraints. If al_resize_display() with the window size will work then there is no need for it. I think all that needs to be done is to remove the resize and update the documentation.

goobliata:
After reading this over again maybe you're right and there should indeed be an `al_apply_window constraints`. `apply` would be a better word.

gh0stwizard:
@goobliata 

Alright, I am ready to finish this task today. I willing to ask about one thing:

If we will implement the `al_apply_window constraints` function I see no other way except adding a new field to `struct ALLEGRO_DISPLAY`, for instance, `bool use_constraints`. Also, possibly (I am not sure yet), a new virtual function should be added to `struct ALLEGRO_DISPLAY_INTERFACE`, for instance, `apply_window_constraints`.

The `use_constraints` field will be used so if it is set to true, then OS specific code will use constraint's values. This is required because Win32 and OS X set constraints by asking application about current sizes of the window via asynchronous "messages". Also, this field might be used by the `al_resize_display` function.

The virtual function `apply_window_constraints` will call platform specific code to set or remove constraints from OS level. I am not sure about is it really necessary to implement it, but it might be useful to simplify implementation.

Such changes will break ABI compatibility as I know. It is ok to do so?

If yes, I will start to change the code. If no, then I simply remove the `al_resize_display` from the `al_set_window_constraints` function, but **min.** constraints will be automatically applied to a window if set (I have explained why early).

Implementation notes:
```c
/* al_apply_window_constraints: enable or disable using window constraints.
 * If enabled then previously set window constraints will be used by a window.
 * If disabled, then a window will not use constraints.
 */
void al_apply_window_constraints(DISPLAY *display, bool onoff)
{
   display->use_constraints = onoff;

   /* Platform specific code to set or remove constraints.
    * This is required at least on OS X, for instance, by calling 
    * [window setContentMaxSize: max_size] and [window setContentMinSize: min_size].
    */
   if (display && display->vt && display->vt->apply_window_constraints)
      display->vt->apply_window_constraints(display, onoff);
}
```

gh0stwizard:
I have finished implementation of `al_apply_window_constraints()`. Here is the report about overall changes:
* Added documentation about changes: updated `al_set_window_constraints`, added information for `al_apply_window_constraints`;
* Updated examples: `ex_projection2.c`, `ex_window_constraints.c`: added call to `al_apply_window_constraints()`;
* Added a new example: `ex_window_maximized.c`: shows how works a maximized window w/wo constraints;
* Behavior of `al_set_window_constraints()` function has been changed: it just sets values of constraints only and does not try to resize a window;
* A new function `al_apply_window_constraints()` has been added: it applies previously set constraints to a window, and if current sizes of the window do not fit in ranges of constraints, then the window **may** be resized (see documentation for more information);
* Fixed bug under Win32: if a window is maximized it has no border;
* Fixed bug under OS X: `al_create_display()` with `ALLEGRO_MAXIMIZED` creates maximized (zoomed) window now;
* Changes was made and tested under Windows 7 64-bit (MSVS 2015), GNU/Linux (GNOME 3, gcc-6.2), OS X Mavericks (XCode 6.2).

New issues & bugs:
* OS X: fail to apply constraints  when a window restores its state from the fullscreen mode to windowed one. I was unable to found out how to fix it in a short period of time; there is a events (messages) hell. Tested under Mavericks (10.9.5). I will try to fix it again in this weekend.

A basic example how constraints are working with new changes:
```c
/* set constraints to a display:  minimum 300x200, maximum 800x600 */
if (!al_set_window_constraints(display, 300, 200, 800, 600)) {
   fprintf(stderr, "failed to set constraints to display\n");
   exit(1);
}

/* apply constraints so they take effect right not */
al_apply_window_constraints(display, true);

/* your code [...] */

/* completely disable constraints */
al_apply_window_constraints(display, false);
```

Is anything else I have to do (change)? May be I had missed something. And if you don't like it I can revert changes and create another implementation or change existing one.

goobliata:
The only change I'd like to see is to move the new vtable methods and variables to the end of the DISPLAY structures to maintain ABI compatibility.

gh0stwizard:
@goobliata 
>The only change I'd like to see is to move the new vtable methods and variables to the end of the DISPLAY structures to maintain ABI compatibility.

Done.

goobliata:
Have a couple issues now. On Windows resizing the window in ex_window_maximized is not extremely laggy, the window gets filled with garbage and takes a good 30 seconds to resize. Also ex_resize2 when maximised the window has a border like before.

I'm wondering for the first issue if there's a missing acknowledge_resize or something. Not sure why this particular window has a border when maximised either.

gh0stwizard:
@goobliata 

If you are testing ABI compatibility, then such issues do not surprise me :) Are you sure that you are running examples with the correct version of *.dll's? Please, check out the PATH environment variable.

P.S. I just did test the ex_resize2 example and it does have such an issue.

goobliata:
I wasn't testing ABI compatibility. That is with a regular build of this pull request. Whether the window is orange or green it messes up when I drag the corner to expand it.

goobliata:
One other thing I was thinking of was to make the apply_constraints state a display option, but I can do that separately after this gets sorted out.

gh0stwizard:
@goobliata 

>Whether the window is orange or green it messes up when I drag the corner to expand it.

The cause of it was that exposure events were not handled. I have fixed it recently. I confirm that there is some issues with resizing a window under Win32, but they are not related to this patch. Probably I have coded the example in a wrong way.

Back to your issues.

>On Windows resizing the window in ex_window_maximized is not extremely laggy, the window gets filled with garbage and takes a good 30 seconds to resize.

I confirm that there was a garbage when resizing the window. But I did not understood about lags and 30 seconds. Can you explain this case better?

>Also ex_resize2 when maximised the window has a border like before.

Do you still have this bug? If ex_window_maximized does not have this bug and ex_resize2 does, then this is extremely strange. I assume in this case that ex_resize2 uses old allegro *.dll files.

In case if the both examples have this bug, then it is possible that I had missed something. Did you test under Windows 10? Or compiled with MSVS 2017? Perhaps there is some WINAPI changes, and I don't know about them.

goobliata:
Sorry, I was on the wrong branch.

Thanks for your hard work on this update. I'll merge it now!

PR #726, opened 2017-02-02
Fix newlines being added ad-infinitum to config files

oitofelix:
Prevent config_write_section from adding superfluous newlines for empty comments, which spontaneously arise when using al_remove_config_* functions.

When saving settings to its main configuration file, MININIM (my Allegro-based game) uses al_get_first_\*, al_get_next_\* and al_remove_config_\* functions to remove any duplicate keys (irregardless of case) to ensure each entry is unique.  Oddly enough, this results in new lines being added before reinserted entries for each removal.  Over time the configuration file gets bigger, sparse and thus confuse to human readers.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -408,9 +408,9 @@ static bool config_write_section(ALLEGRO_FILE *file,
                al_fputs(file, "# ");
             }
             al_fputs(file, al_cstr(e->key));
+            al_fputc(file, '\n');
+            *should_blank = false;
          }
-         al_fputc(file, '\n');
-         *should_blank = false;
       }
       else {
          al_fputs(file, al_cstr(e->key));


SiegeLord:
So the reason the code is written this way is because the config system is meant to preserve blank lines in the configuration files when loading and re-saving them. Blank lines are inserted as empty comments into the internal data-structure and that's why the code unconditionally adds a newline for each one of them.

Could you describe your problem a bit more? Do you load a config file and then just save it? Do you generate multiple config entries, then prune the duplicates and then save it?

oitofelix:
> So the reason the code is written this way is because the config system is meant to preserve blank lines in the configuration files when loading and re-saving them. Blank lines are inserted as empty comments into the internal data-structure and that's why the code unconditionally adds a newline for each one of them.

I see.  I didn't know that.

> Could you describe your problem a bit more?

Sure.  I think the best way to describe it is providing you the offending code.

For instance, every time a menu item that changes a setting of MININIM is clicked on, the function ui_save_setting is invoked to ensure MININIM's main configuration file reflects such change (as to preserve it for future invocations of the game) by adding a corresponding key/value pair.  The ui_save_setting function makes sure each key pair is unique by removing any previous entries (possibly multiple, disregarding case --- this is done by remove_config_entry function) and then adding it again with the correct value pair.  Both functions are listed below:

**ui_save_setting**

```C
bool
ui_save_setting (char *section, char *key, char *value)
{
  ALLEGRO_CONFIG *config = NULL;

  if (al_filename_exists (config_filename)) {
    config = al_load_config_file (config_filename);
    if (! config) goto error;
    const char *file_type_str =
      al_get_config_value (config, NULL, "FILE TYPE");
    if (! file_type_str
        || strcasecmp (file_type_str, "MININIM CONFIGURATION")) {
      al_destroy_config (config);
      config = al_create_config ();
    }
  } else if (value) {
    if (! al_make_directory (user_settings_dir)) goto error;
    config = al_create_config ();
  } else return false;

  if (! config) goto error;

  remove_config_entry (config, NULL, "FILE TYPE");
  al_set_config_value (config, NULL, "FILE TYPE", "MININIM CONFIGURATION");

  bool existing;
  if (key) {
    existing = remove_config_entry (config, section, key);
    if (value) al_set_config_value (config, section, key, value);
  } else existing = remove_config_section (config, section);

  if (! al_save_config_file (config_filename, config)) goto error;

  al_destroy_config (config);

  return existing;

 error:
  if (config) al_destroy_config (config);
  error (0, al_get_errno (), "can't save setting '[%s] %s=%s' to '%s'",
         section, key, value, config_filename);
  return false;
}
```

**remove_config_entry**

```C
bool
remove_config_entry (ALLEGRO_CONFIG *config, char *section, char *key)
{
  ALLEGRO_CONFIG_ENTRY *iterator;
  char const *entry = al_get_first_config_entry (config, section, &iterator);

  bool removed = false;
  while (entry) {
    if (! strcasecmp (entry, key)) {
      removed = al_remove_config_key (config, section, entry);
      entry = al_get_first_config_entry (config, section, &iterator);
    } else entry = al_get_next_config_entry (&iterator);
  }

  return removed;
}
```

This procedure, based solely on the two aforementioned functions and their Allegro primitives, triggers the odd behavior that originated this pull request.

SiegeLord:
Could you try out #731?

SiegeLord:
Closed in favor of https://github.com/liballeg/allegro5/pull/731.

Issue #460, opened 2011-05-15
[4.x]Make pack_fopen a weak symbol, add __allegro_pack_fopen


allebot:
Originally reported by: bero

Original Ticket: [alleg/feature-requests/42](https://sourceforge.net/p/alleg/feature-requests/42)

When porting AGS to Linux, I noticed they copied src/file.c from allegro in order to rename pack_fopen to __old_pack_fopen and override it with a new pack_fopen that handles looking inside another file (and then falls back to calling __old_pack_fopen).

Since this is a relatively dirty hack and doesn't work with static linking at all (and I can see valid reasons to do it), I'd propose to rename pack_fopen to __allegro_pack_fopen and adding a pack_fopen that just calls __allegro_pack_fopen, but is declared a weak symbol.

This way:
- "Normal" applications using allegro will keep working just fine
- AGS and others wanting to do the same thing can override pack_fopen without having to duplicate code
- Overridden pack_fopen functions can call the original (because it's available unconditionally through the __allegro_pack_fopen symbol)

ivan-mogilko:
I was recently doing investigation on this in the project that caused this feature request (Adventure Game Studio), and the actual problem, at least according to what I found so far, is that SOME of the Allegro objects do not (or did not) support creation from passed PACKFILE, only from passed filename.

Primary issue for our project is "load_midi". I checked in 4.4 branch and there is still no variant that would create MIDI object from passed PACKFILE. I believe it should not be a problem to add such function to Allegro API, because load_midi calls pack_fopen internally anyway.

ivan-mogilko:
Sorry for bumping this after such a long time, but I believe this issue has been adressed by #729 and #761, and is no longer a problem.

SiegeLord:
Seems so. Thanks @ivan-mogilko.

PR #729, opened 2017-02-11
4.4: implement load_midi_pf

ivan-mogilko:
This adds **load_midi_pf** function as a part of Allegro 4.4 API.

Resource loading functions in Allegro 4 usually come in pairs, one takes filename as parameter, another takes PACKFILE, which may be convenient if program uses custom resource storage. But midi does not provide PACKFILE variant.

I tried to keep code changes at minimum, basically splitting existing load_midi into two functions, using existing function pairs (such as **load_wav** & **load_wav_pf**) as an example.

This request is related to an old issue created several years ago: #460  I found that the root of the issue was that some objects could not be created using program's own PACKFILEs at that time. MIDI seem to be the only object left that still missing this possibility. If this pull request is accepted, I believe #460 could be closed.

### DIFF
--- a/include/allegro/midi.h
+++ b/include/allegro/midi.h
@@ -24,6 +24,10 @@
 #ifdef __cplusplus
    extern "C" {
 #endif
+
+struct PACKFILE;
+
+
                                        /* Theoretical maximums: */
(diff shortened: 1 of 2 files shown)


goobliata:
Great, thanks for thiis!

PR #738, opened 2017-03-11
Work around crashes on some Androids when an FBO is bound during egl_…

goobliata:
…SwapBuffers

### DIFF
--- a/src/android/android_display.c
+++ b/src/android/android_display.c
@@ -658,8 +658,15 @@ static void android_unset_current_display(ALLEGRO_DISPLAY *dpy)
 
 static void android_flip_display(ALLEGRO_DISPLAY *dpy)
 {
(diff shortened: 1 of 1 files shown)


goobliata:
For future note the device is a 1st gen Kindle Fire with Android 2.3.

elias-pschernig:
I don't understand what this does... does al_set_target_bitmap clear FBO bindings? If that's the case, could clearing FBOs have any possible side effects?

goobliata:
FBO is bound when you set the target to a bitmap. So on some devices if the target is a bitmap (and not the display), it will crash on flip. This has no side effects because it's set back right after the flip.

elias-pschernig:
Sounds good to me then!

goobliata:
Just Yet Another Stupid Workaround for Android! :)

PR #739, opened 2017-03-17
Add al_x_set_initial_icon.

goobliata:
This requires libxpm but is optional. This patch also sets the default
icon to Alex (if you have xpm.) This patch is required to set an icon
on Ubuntu Unity (and possibly others) without using .desktop files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -635,6 +635,13 @@ if(SUPPORT_X11)
         message(FATAL_ERROR "X11 support requires Xcursor library.")
     endif(CAN_XCURSOR)
 
+    check_library_exists(Xpm XpmCreatePixmapFromData "" CAN_XPM)
+    if(CAN_XPM)
+        set(ALLEGRO_XWINDOWS_WITH_XPM 1)
+        find_library(XPM_LIB "Xpm")
+        list(APPEND X11_LIBRARIES "${XPM_LIB}")
+    endif(CAN_XPM)
+
     check_include_file("X11/extensions/XInput2.h" CAN_XINPUT2)
     run_c_compile_test("
         #include <X11/extensions/XInput2.h>
--- a/docs/src/refman/platform.txt
+++ b/docs/src/refman/platform.txt
@@ -169,3 +169,12 @@ These functions are declared in the following header file:
 Retrieves the XID associated with the Allegro display.
 
 Since: 5.1.12
+
+### API: al_x_set_initial_icon
+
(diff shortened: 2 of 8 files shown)


elias-pschernig:
Oh, that sounds really useful! Travis CI has a compile error though about a missing XpmCreatePixmapFromData, not sure if that's false positive or something needs to be fixed. I'll try it out tonight.

goobliata:
Oops, forgot to test it with libxpm-dev not installed. Should work now.

PS: there may be an easier way to create a pixmap from an ALLEGRO_BITMAP than use libxpm, but I didn't feel like searching through X11 references etc. The method used is slow for large icons with lots of colours... I tested with a 256x256 icon with lots of colors and it takes a couple seconds, while the same image reduced to 256 colors (looks the same) loads instantly.

SiegeLord:
Is this something that could be useful cross-platform for the same reasons that `al_set_new_window_title` was made? Not something that needs to be done now, obviously, but perhaps sometime soon after.

goobliata:
The purpose of the patch initially was just to get an icon on Unity, as when you're distributing a game it's annoying to have to ask people to set up a .Desktop file (or use an installer.)

Initially I didn't think it was possible but I ran a a4 program and it had an Alex icon. So after trying that code and it not working, I eventually figured out that it had to be done before mapping the window.

So that's the main reason it's like that. On other platforms I suppose there may be a millisecond flash of default icon so it could be useful there too, though I never noticed it.

fatcerberus:
Does this work even if there *is* a `.desktop` file?  That'd be useful for miniSphere since I change the window icon based on the game that's running.

goobliata:
No it doesn't work when there is a .desktop file.

Any consensus on this? Is it good to apply or needs changes?

elias-pschernig:
I'm sorry, I was travelling both last and this weekend and never found any time for Allegro :( I agree with SiegeLord that it probably could be made to work on all platforms eventually. But I really would like to have a go at speeding up the xpm conversion, there is no reason even a 1024x1024 one should take more than a few ms. I can still do that later though so I'm ok with applying it.

goobliata:
Yeah, the xpm conversion is kind of a hack, it stores every unique colour in a vector so it'd work with any number of colours... it's searching the vector for matching colours that takes so long. You could speed it up just by duplicating colours which I'm sure xpm format wouldn't care about.

SiegeLord:
I'm fine with merging this as is.

Issue #749, opened 2017-04-20
Deprecation warning when starting app


tbass:
When I start an application (OSX), I receive a deprecation warning

```
2017-04-20 09:27:29.669488+0200 ex_haiku[4790:827995] SLPSGetCurrentProcess(): This call is deprecated and should not be called anymore.
2017-04-20 09:27:29.669526+0200 ex_haiku[4790:827995] set_foreground_operation_state(): This call is deprecated and should not be called anymore.
```

PR #750, opened 2017-04-20
Fixed deprecation warning when starting OSX console app

tbass:
This is the pull request for issue #749
Chnaged the deprecated way to the new TransformProcessType

### DIFF
--- a/src/macosx/system.m
+++ b/src/macosx/system.m
@@ -64,9 +64,9 @@ _AL_VECTOR _osx_threads = _AL_VECTOR_INITIALIZER(THREAD_AND_POOL *);
  */
 static void osx_tell_dock(void)
 {
-    ProcessSerialNumber psn = { 0, kCurrentProcess };
-    TransformProcessType(&psn, kProcessTransformToForegroundApplication);
-    [NSApplication sharedApplication];
+   ProcessSerialNumber psn = { 0, kCurrentProcess };
+   TransformProcessType(&psn, kProcessTransformToForegroundApplication);
+   [NSApplication sharedApplication];
 }
 
 
--- a/src/macosx/system.m
+++ b/src/macosx/system.m
@@ -66,7 +66,7 @@ static void osx_tell_dock(void)
 {
    ProcessSerialNumber psn = { 0, kCurrentProcess };
    TransformProcessType(&psn, kProcessTransformToForegroundApplication);
-   [NSApplication sharedApplication];
+   [[NSApplication sharedApplication] activateIgnoringOtherApps: YES];
 }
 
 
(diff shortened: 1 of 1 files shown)


goobliata:
Could you fix the indentation? We indent with 3 spaces per stop. Thanks.

goobliata:
Also does this actually fix any deprecation or just hide them? Do you get console output from the relevant examples?

tbass:
Fixed the indention
It really fixes the deprecation warning when starting a console application and uses the proposed Apple way using ```TransformProcessType``` introduced in OSX 10.3. 
Relevant output is in issue #749

goobliata:
And are you still getting console output?

goobliata:
I tried it myself. Indeed it fixes the deprecations warnings, but for me windows now pop up behind everything else. Not for you?

goobliata:
```
diff --git a/src/macosx/system.m b/src/macosx/system.m
index b441571..b8282f9 100644
--- a/src/macosx/system.m
+++ b/src/macosx/system.m
@@ -66,7 +66,7 @@ static void osx_tell_dock(void)
 {
    ProcessSerialNumber psn = { 0, kCurrentProcess };
    TransformProcessType(&psn, kProcessTransformToForegroundApplication);
-   [NSApplication sharedApplication];
+   [[NSApplication sharedApplication] activateIgnoringOtherApps:YES];
 }
```

With this patch on top of yours the windows come to the front. Can you confirm?

tbass:
Nope, I can not confirm. The ex_* examples I use still come up in the same manner both ways (with or without your addition). Also my own little test game pops up in front.

I think it won't hurt to add the ```activateIgnoringOtherApps:YES``` as it explicitly activates the app, even though, Apple does not recommend to use it (ref: http://apple.co/2ocKTbB)

fatcerberus:
For a game (the usual Allegro use case), I personally don't mind the potential focus-stealing.  I vote for keeping the flag.

goobliata:
If you could add my patch to the PR I will merge this, thanks!

tbass:
added your patch

src/macosx/system.m
```diff
@@ -66,7 +66,7 @@ static void osx_tell_dock(void)
 {
    ProcessSerialNumber psn = { 0, kCurrentProcess };
    TransformProcessType(&psn, kProcessTransformToForegroundApplication);
-   [NSApplication sharedApplication];
+   [[NSApplication sharedApplication] activateIgnoringOtherApps: YES];
```
fatcerberus: This is neither here nor there, but I just have to say that Objective-C is a very strange language. :-p

Issue #194, opened 2009-02-25
kcm_audio: use al_set_errno


allebot:
Originally reported by: wangp

Original Ticket: "alleg/bugs/173":https://sourceforge.net/p/alleg/bugs/173

Make kcm_audio use al_set_errno instead of its own thing.

tbass:
Hope, that this is the intended way that needed to be done. As far as I can see, the old method does nothing special besides adding one unused error code.

PR #753, opened 2017-04-22
Replaced the custom _al_set_error method 

tbass:
with the default ALLEGRO_ERROR call as it does not do anything else

Removed all traces of the _al_set_error functionality

refers to issue #194

### DIFF
--- a/addons/audio/allegro5/internal/aintern_audio.h
+++ b/addons/audio/allegro5/internal/aintern_audio.h
@@ -353,15 +353,6 @@ extern void _al_kcm_mixer_read(void *source, void **buf, unsigned int *samples,
    ALLEGRO_AUDIO_DEPTH buffer_depth, size_t dest_maxc);
 
 
-typedef enum {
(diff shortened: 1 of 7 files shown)


goobliata:
This could be used at some point to add error codes, so I will not accept this. Also the error codes are not shown anymore which might be helpfull.

SiegeLord:
I think the intended way to solve that was perhaps to do:

```
al_set_errno(ALLEGRO_GENERIC_ERROR);
ALLEGRO_ERROR("A driver already running");
```

Granted, `al_set_errno` isn't often used anyway. Alternatively maybe the right way would be to move that custom _al_set_error into the main library and rename it?

SiegeLord:
In terms of the first suggestion it could be as simple as sticking the `al_set_errno` inside `_al_set_errno`, hah.

goobliata:
yes it would make sense to use al_set_errno, just make sure the codes don't collide with existing ones.

PR #760, opened 2017-05-16
Fix OSX magic main with the LTO switch.

### LINKED ISSUE #758: Using clang 8.1 with -flto fails on macOS 10.12
Using -flto with clang 8.1 strips out _al_mangled_main() because it things the function is not used. Removing -flto as a linker flag fixes this, but the following change fixes the issue properly:

> 
> --- alosx.h.old	2016-09-25 03:32:06.000000000 +0200
> +++ alosx.h	2017-05-12 14:08:51.000000000 +0200
> @@ -53,7 +53,7 @@
>     #ifndef ALLEGRO_NO_MAGIC_MAIN
>        #define ALLEGRO_MAGIC_MAIN
>        #if __GNUC__ >= 4
> -         #define main __attribute__ ((visibility("default"))) _al_mangled_main
> +         #define main __attribute__ ((visibility("default"),used)) _al_mangled_main
>        #else
>           #define main _al_mangled_main
>        #endif
>

SiegeLord:
Using -flto with clang 8.1 strips out _al_mangled_main() because it things the
function is not used. Removing -flto as a linker flag fixes this, but the
following change fixes the issue properly.

Fixes #758

### DIFF
--- a/include/allegro5/platform/alosx.h
+++ b/include/allegro5/platform/alosx.h
@@ -53,7 +53,7 @@ ALLEGRO_PATH *_al_osx_get_path(int id);
    #ifndef ALLEGRO_NO_MAGIC_MAIN
       #define ALLEGRO_MAGIC_MAIN
       #if __GNUC__ >= 4
-         #define main __attribute__ ((visibility("default"))) _al_mangled_main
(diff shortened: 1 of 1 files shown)


SiegeLord:
Merged as 71a38151f3d499f5d1b93511759f7f3e8ae853d7.

PR #762, opened 2017-05-17
4.4: Make use of <stdint.h> in MSVS 2010 and higher

ivan-mogilko:
This is a small fix that lets compile Allegro-based programs in MSVS 2010 and higher (tested VS 2010, *15 and *17).

MSVS introduced stdint.h starting with VS2010, but if it is included in Allegro 4 program, this will cause conflicts with Allegro's astdint.h. Provided fix seem to deal with the issue.

### DIFF
--- a/include/allegro/platform/almsvc.h
+++ b/include/allegro/platform/almsvc.h
@@ -78,8 +78,12 @@
 #define INLINE       __inline
 
 #define LONG_LONG    __int64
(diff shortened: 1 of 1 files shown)


SiegeLord:
Merged as 71a38151f3d499f5d1b93511759f7f3e8ae853d7. Thanks!

PR #774, opened 2017-06-18
Only test cooperative lavel if device is known to be lost.

SiegeLord:
@goobliata We reverted a version of this patch a few years ago as it broke some of your games. Could you test this one on them?

### DIFF
--- a/include/allegro5/mouse.h
+++ b/include/allegro5/mouse.h
@@ -40,7 +40,7 @@ struct ALLEGRO_MOUSE_STATE
     * (z) Mouse wheel position (1D 'wheel'), or,
     * (w, z) Mouse wheel position (2D 'ball')
     * display - the display the mouse is on (coordinates are relative to this)
-    * pressure - the pressure appleid to the mouse (for stylus/tablet)
+    * pressure - the pressure applied to the mouse (for stylus/tablet)
     */
    int x;
    int y;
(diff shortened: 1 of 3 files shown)


goobliata:
One of my games crashes on reset with this, I'll have to investigate it.

SiegeLord:
Perfect, let me know if I can be of help.

goobliata:
I can reproduce the crash in ex_fs_window by adding ALLEGRO_PROGRAMMABLE_PIPELINE to the flags (at first I was trying to fix my game but after being unable to fix it, I tried to reproduce it in the example because it's also crashing in toggling fullscreen window in my game.)

goobliata:
Possible hint: it was crashing in al_draw_prim in my game, ex_fs_window also draws primitives, perhaps it's the primitive shaders.

goobliata:
I haven't looked into the cause of the crash yet. It might just be that we're not releasing everything when we toggle fullscreen window.

SiegeLord:
I got it crashing with ex_fs_window too, like you said; haven't had time to look at it.

SiegeLord:
Turns out we were leaking D3D shaders if we attached both pixel and vertex shader sources. The commit here fixes the crash with ex_fs_window. Could you test this again? Thanks!

goobliata:
Works in my game now as well.

PR #775, opened 2017-06-20
Improved path behavior in GTK native file dialog.

bkdoormaus:
Before, the path "/a/b/c/" would open with the directory at "/a/b/", instead of
the desired path.

I believe the new logic corrects this.

### DIFF
--- a/addons/native_dialog/gtk_filesel.c
+++ b/addons/native_dialog/gtk_filesel.c
@@ -52,23 +52,15 @@ static gboolean create_gtk_file_dialog(gpointer data)
    }
 
    if (fd->fc_initial_path) {
-      if (fd->flags & ALLEGRO_FILECHOOSER_FILE_MUST_EXIST || ! save) {
+      if (strcmp(al_get_path_filename(fd->fc_initial_path), "") != 0) {
          gtk_file_chooser_set_filename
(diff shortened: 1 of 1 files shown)


goobliata:
Looks better to me.

SiegeLord:
This feels like more of a bug with our file routines than the GTK dialog per-se. No objections from me though, the new behavior makes as much sense as the old one.

oitofelix:
Is this consistent with other platforms' native file dialogs?

bkdoormaus:
I don't know. I don't have access to a Windows machine to test at the moment and for the foreseeable future.

From a cursory look, the Windows file dialog box should have similar behavior. It has an extra check to see if the provided path is a directory, and if not, clears the filename portion of the path. I'm not sure if that's ideal. I also don't see where the Windows file dialog box assigns the filename portion, if provided.

Again, I don't have a Windows machine so I can't test it.

goobliata:
It works the same on Windows. A note should be added to the docs to add a trailing slash (Windows doesn't need it but macOS/Linux too) if you want a directory. macOS behaves almost the same except it doesn't allow setting the inital path to a file (it will then just show the directory the file is in.)

goobliata:
Thanks for the fix!

PR #776, opened 2017-06-23
Fixed corrupted regexps while generating python bindings

verderten:
Got errors while building


verderten:
Strangely, after re-checking unchanged version it seemed to work as expected 
Previously it was failing because of unknown '\e'

SiegeLord:
Hmm, even if it works, it seems that the intention was to have '\b' there.

PR #777, opened 2017-06-26
Fix "Since" field for al_x_set_initial_icon in docs

dos1:
5.2.2 has been already released and doesn't contain al_x_set_initial_icon function. 
It's now 5.2.3 in git, so let's set it as that.


SiegeLord:
Thanks!

PR #778, opened 2017-06-26
Android: fix LocalRef leak when opening files using APK interface

dos1:
LocalRef pool is limited in size to 512 elements, so code like
```c
for (int i=0; i<512; i++) {
  ALLEGRO_FILE *file = al_fopen(filename, "r");
  if (file) {
    al_fclose(file);
  }
}
```
was crashing the application. This also applied to functions like al_filename_exists, making it even easier to trigger.


goobliata:
Thank you for this!

PR #787, opened 2017-07-20
Fix al_set_clipping_rectangle() when out-of-bounds

### LINKED ISSUE #786: Vertex buffer creation randomly fails (OpenGL)
I found out that in miniSphere, `al_create_vertex_buffer()` often fails after a hundred or so calls.  This used to be fine because I had a fallback to keep vertices in software, so I didn't notice the failures.  I recently made it a hard error if VBO creation fails, and now I'm seeing games randomly failing with VO creation errors.

I have a GTX 1070 on my test machine, with 8 GB of VRAM, so I don't think I'm running out of memory.  We're talking very small buffers here, between 4-100 vertices.  Allegro is initialized in OpenGL mode; I haven't tested with D3D.

fatcerberus:
When using OpenGL and calling `al_set_clipping_rectangle()` specifying a rectangle that resides completely off-surface, the width and height of the scissor box are currently calculated as negative, causing `glScissor()` to post a `GL_INVALID_VALUE` error.  That error gets picked up later when an unrelated function checks `glGetError()`, causing mysterious failures.

This fixes the bug by explicitly clamping the scissor box to width/height 0.

Fixes #786.

### DIFF
--- a/src/bitmap.c
+++ b/src/bitmap.c
@@ -384,6 +384,12 @@ void al_set_clipping_rectangle(int x, int y, int width, int height)
    if (y + height > bitmap->h) {
       height = bitmap->h - y;
    }
+   if (width < 0) {
+      width = 0;
+   }
+   if (height < 0) {
+      height = 0;
+   }
 
    bitmap->cl = x;
    bitmap->ct = y;


SiegeLord:
Nicely done! Thanks for debugging this and the fix!

PR #789, opened 2017-07-23
Several fixes to DirectDraw Windows backend

evouga:
- the switch-in and switch-out handlers need mutexes to prevent race conditions with drawing when alt-tabbing into or out of fullscreen.
 - added message handlers to restore the palette if the system palette changed while the program was tabbed out
 - critical section needs to be initialized before installing the user window hook

### DIFF
--- a/src/win/wdispsw.c
+++ b/src/win/wdispsw.c
@@ -111,6 +111,8 @@ void _win_switch_in(void)
 
    _TRACE(PREFIX_I "switch in\n");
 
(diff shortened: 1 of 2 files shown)


SiegeLord:
Seems reasonable. Thanks!

evouga:
You are absolutely right, I missed applying a patch that adds
#include "wddraw.h"
which contains the missing definitions.

PR #791, opened 2017-07-25
Gracefully handle all cases in GTK file dialog.

bkdoormaus:
I tried the following cases. I don't think there's any other permutations possible.

For save dialogs:
* Path is directory: Shows directory
* Path is non-existing file: Sets the filename portion to the filename portion of the path, shows directory
* Path is an existing file: Selects file, shows directory

For open dialogs:
* Path is a directory: Shows last existing directory
* Path is a non-existing file: Shows directory, no filename set
* Path is an existing file: Shows directory, filename set

(Directory being the directory the file is in, or whatever portion of the path actually exists on the filesystem.)

There is one small thing: if you do something like "/usr/foo/bar/baz" for the save dialog, the directory will be "/usr" and the filename portion will be "baz". (Assuming directory "foo" doesn't exist). Is that behavior that acceptable?

### DIFF
--- a/addons/native_dialog/gtk_filesel.c
+++ b/addons/native_dialog/gtk_filesel.c
@@ -52,15 +52,45 @@ static gboolean create_gtk_file_dialog(gpointer data)
    }
 
    if (fd->fc_initial_path) {
-      if (strcmp(al_get_path_filename(fd->fc_initial_path), "") != 0) {
-         gtk_file_chooser_set_filename
-            (GTK_FILE_CHOOSER(window),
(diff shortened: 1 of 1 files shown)


SiegeLord:
That looks nice. The only permutation that I have a little bit of an issue with is the "Path is a non-existing file: Shows directory, no filename set" one. It seems like it'd be convenient to show the filename there as well, as a hint to the user what is expected.

I note however that maybe GTK doesn't even support that, as the current code doesn't check for the file existing, yet the behavior I want doesn't seem to happen.

SiegeLord:
Ok, looks like that's impossible. Merged as f52d73a as is. thanks!

PR #797, opened 2017-08-11
Missing #include

evouga:
Pull request #789 missed a patch that #includes a required header file. Very sorry about that.


SiegeLord:
We actually already have a commit to fix this: https://github.com/liballeg/allegro5/commit/c12b6be726f8959404734e52c36edfdd6edf2461

Thanks anyway!

PR #802, opened 2017-08-13
Android: use clang instead of gcc

dos1:
GCC in NDK is unfortunately deprecated since r11 and won't be updated beyond 4.9 (and is eventually going to be removed).

Clang has been introduced in r8c. The last NDK version that allows to target the lowest API level supported by Allegro (android-9) is r14b, so the compiler can be safely switched to Clang.


SiegeLord:
Merged as 15c1650cda1cf994a419e611c713796bc1246603. Thanks!

PR #805, opened 2017-08-15
Implement per-bitmap blenders

### LINKED ISSUE #783: Allow setting blend mode on a per-bitmap basis
Shaders, transformation matrices, clipping rectangles, etc. are all set per-bitmap, but blend modes are currently set per-thread.  This is somewhat confusing and perhaps a source of bugs.  Perhaps we could consider changing `al_set_blender()` to be set per-bitmap.

I realize this would be a breaking change so I'm not sure how practical it would be to fix, but figured I'd put this here anyway.

fatcerberus:
This adds the APIs listed below, plus accompanying documentation.

* `al_get_bitmap_blender(*op, *src, *dst)`
* `al_get_separate_bitmap_blender(*op, *src, *dst, *alpha_op, *alpha_src, *alpha_dst)`
* `al_get_bitmap_blend_color()`
* `al_set_bitmap_blender(op, src, dst)`
* `al_set_separate_bitmap_blender(op, src, dst, alpha_op, alpha_src, alpha_dst)`
* `al_set_bitmap_blend_color(color)`
* `al_reset_bitmap_blender()`

All functions operate on the current target bitmap.

Before `al_set_bitmap_blender()` is called for a given bitmap, its "use bitmap blender" flag is false and the thread-local blender is used instead.  `al_reset_bitmap_blender()` clears any blender set for the bitmap and returns it to using the thread-local blender.

Let me know if there are any issues with the implementation.

Fixes #783.


elias-pschernig:
To me names like al_get_bitmap_blender are a bit confusing. Currently all al_get_bitmap_* functions like al_get_bitmap_width operate on an ALLEGRO_BITMAP you pass in as first argument, not on the currently active target bitmap. A less confusing but more awkward name would be al_get_current_target_bitmap_blender. Maybe we can shorten that to al_get_current_blender or al_get_target_blender?

fatcerberus:
I actually debated having them take an `ALLEGRO_BITMAP` as first argument (there's no reason they can't, it's just setting a few ints under the hood), but decided to have them operate on the current target bitmap by analogy with `al_set_clipping_rectangle()`.  I'd be willing to change it if you think that'd be less confusing.  I don't want to make the names any more verbose than they already are if it can be avoided, names like `al_set_separate_bitmap_blender()` are already bordering on being too long.

SiegeLord:
I don't mind the `bitmap` terminology too much. Yes, it's definitely inconsistent, but I think it's the clearest option available to us.

fatcerberus:
@SiegeLord Are you okay with the current behavior of operating on the current target bitmap, or should I change it to take an explicit bitmap pointer?

SiegeLord:
I definitely prefer the current target bitmap, that's consistent with every other API we have. I think this aspect is the one we should definitely keep. I wouldn't worry about the names too strongly, I feel they're in some sense temporary until the far future when we can replace the current blender functions with these, so I'm mostly focused on clarity where possible.

fatcerberus:
I couldn't think of what to do for an example program that isn't already covered by the ones already present.  It's the same blender mechanism we've always had, just that its per-bitmap now.

I'll try to come up with some tests but the test harness confused me--there were hashes and stuff and the templates kind of baffled me too.  I'll study it in more detail later and try to decipher it all. :)

SiegeLord:
If the test system is giving you too much trouble, then I wouldn't worry about it... it's a little too obtuse for its own good, someone should replace it.

For an example, the typical way these things are demonstrated is that you have multiple bitmaps, set their blenders before the main loop and then switch between them during the main loop with the resultant differences in the blending mode happening automatically.

fatcerberus:
Alright, I've been busy with miniSphere the past month but I'll see if I can get this PR cleaned up in the next day or so so it can be merged.

fatcerberus:
The tricky part of making an example for this is that setting the bitmap blender only affects bitmaps used as render targets, i.e. the blend mode is used for things drawn *to* the bitmap and not for drawing the bitmap itself.  So it's not enough just to load a few bitmaps and draw them, I'll need to have it do some compositing or something.

It doesn't help either that I have no experience writing Allegro examples from scratch.  I have some free time now though, so I'll see what I can do.

fatcerberus:
No, problem, sorry I never got around to finishing this, other stuff got in the way. :frowning:

PR #810, opened 2017-08-18
Another attempt to get newer FreeType's linking.

### LINKED ISSUE #773: Adjust build system to support newer freetype
This is wrt png and bz2 dependencies.

SiegeLord:
This time we force the user to specify which optional FreeType dependencies we
should link in. This is supported by a build check that will fail if the user
screws up. Unfortunately it doesn't appear to be possible to reliably to detect
these things automatically.

Fixes #773


fatcerberus:
Since it's supported by a compile check, would it be possible to do multiple compile checks to determine which dependencies are needed?  That would probably slow down the configuration step a lot though.

SiegeLord:
This is what https://github.com/liballeg/allegro5/pull/794 did. It tried to compile without one of the dependencies each time. If it failed to compile, then we considered that dependency mandatory. The issue is that PNG and Zlib provided overlapping symbols (because PNG comes alongsize Zlib), so I couldn't get that system to disambiguate between PNG and Zlib requirement (i.e. it would erroneously conclude that PNG was required to link the TTF addon).

SiegeLord:
Tested this, and it seems to work okay everywhere in conjunction with #820, so meging it. It's not ideal, but it's better than it was before I think.

PR #811, opened 2017-08-18
android: use immersive mode for fullscreen to hide navigation buttons

### LINKED ISSUE #800: Support immersive mode ("real" fullscreen) on Android
Android supports (since API level 19, that is Android 4.4) a special mode called [immersive](https://developer.android.com/training/system-ui/immersive.html), which hides everything, even navigation buttons on devices without hardware ones, and gives the application access to whole screen. User can then slide out status and navigation bars with slide gesture.

I don't know how to proceed with implementing it in Allegro (should it get a new, platform-specific display flag, or should it be just ALLEGRO_WINDOW_FULLSCREEN, since it's actually just that, a real fullscreen mode? or maybe even just hardcoded default, like standard fullscreen right now? should the change of current behaviour be considered a breaking change then?), so I'll just leave here this code which works for me for you to use :)

```java
if (android.os.Build.VERSION.SDK_INT >= 19) {
   this.getWindow().getDecorView().setSystemUiVisibility(View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY | View.SYSTEM_UI_FLAG_HIDE_NAVIGATION);
}
```

at the end of `onCreate` in `android/allegro_activity/src/AllegroActivity.java`, with `import android.view.View;` added earlier.

The downside of this code is that it needs SDK 19 to build, but if that's any issue then I think that might be fixable using reflection. Doesn't change any runtime dependency though.

If adding it just like that is fine for you, I'll be happy to provide a pull request :)

dos1:
This bumps the build dependency to SDK 19.

Closes #800

### DIFF
--- a/README_android.txt
+++ b/README_android.txt
@@ -24,8 +24,8 @@ environment variable. If you installed the SDK in $HOME/android-sdk, add
 $HOME/android-sdk/tools to your PATH.
 
 Once extracted you must run "android" (or android.bat on Windows) and install
-a version of the SDK. Allegro needs version 13 (3.1) of the SDK or higher to
-build. It can run on version 9 and up still though, if you include the
+a version of the SDK. Allegro needs version 19 (4.4) of the SDK or higher to
+build. It can run on version 9 (2.3) and up still though, if you include the
 correct uses-sdk fields in your manifest. See:
 https://developer.android.com/guide/topics/manifest/uses-sdk-element.html
 
--- a/android/allegro_activity/src/AllegroActivity.java
+++ b/android/allegro_activity/src/AllegroActivity.java
@@ -14,6 +14,7 @@ import android.graphics.Rect;
 import android.util.Log;
 import android.view.Display;
 import android.view.SurfaceHolder;
+import android.view.View;
 import android.view.ViewGroup;
 import android.view.Window;
 import android.view.WindowManager;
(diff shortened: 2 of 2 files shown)


SiegeLord:
Alright, I finally got to testing this (I had a lot of trouble compiling Allegro for Android). This definitely looks super nice, but I'm not sure I like how this is unconditionally try. Do you think we could reuse, e.g. `ALLEGRO_FRAMELESS` flag to control this? I tried to implement something like this myself, but I'm not super familiar with how things are done on Android so I mostly just got "ex_draw_bitmap has stopped...".

dos1:
Sure, I can implement it. `ALLEGRO_FULLSCREEN` and `ALLEGRO_FULLSCREEN_WINDOW` are ignored on Android anyway, so maybe setting one/any of them to enable immersive mode would be fine? (with `ALLEGRO_FRAMELESS` I'd have to add an `#ifdef` to my code :D)

SiegeLord:
I'd rather not use `FULLSCREEN` vs `FULLSCREEN_WINDOW` distinction, as `FULLSCREEN` implies being able to set resolution (ideally, it should fail on mobile). Could also add a new flag for this purpose, since FRAMELESS isn't quite right anyway.

fatcerberus:
> Could also add a new flag for this purpose, since FRAMELESS isn't quite right anyway.

I don't know, I feel like FRAMELESS is exactly right here; on desktop platforms `ALLEGRO_FRAMELESS` creates a window without any OS-provided chrome; Android "immersive mode" hides the navigation buttons, i.e. OS-provided chrome.  So it seems like the right tool for the job to me.

SiegeLord:
That's an interesting way to think about it. I'd be happy with FRAMELESS being it then.

dos1:
OK, `ALLEGRO_FRAMELESS` it is then, I'll try to provide a patch soon.

SiegeLord:
Superseded by #1071.

Issue #781, opened 2017-07-05
Set standard file interface before loading /boot/config.txt


goobliata:
Easy fix but if you're using PHYSFS or another file interface on Raspberry Pi, creating the display loads /boot/config.txt but it can fail due to the non-standard interface interfering.

PR #812, opened 2017-08-18
raspberrypi: set default file interface when attempting to read /boot/config.txt

dos1:
Reading #781 helped me to realize why config file handling wasn't working in my games on Android, so as a tribute, I decided to fix it :D


SiegeLord:
Finally got around to booting my RPI, works like a charm! Thanks! Merged as 258c5bb82d2d8285ba25bf417786028a0fde8197.

PR #813, opened 2017-08-19
Ensure detach from buffers before deleting.

juvinious:
Need to detach from buffers before deleting otherwise openal will complain:

```bash
AL lib: (WW) FreeDevice: (0x227d7f0) Deleting 4 Buffer(s)
```
Sources should be deleted before buffer is deleted [line 447](https://github.com/liballeg/allegro5/blob/master/addons/audio/openal.c#L447)

Test **code**:
```cpp
#include <stdio.h>
#include <allegro5/allegro.h>
#include <allegro5/allegro_audio.h>
#include <allegro5/allegro_acodec.h>

int main(int argc, char ** argv){
    al_init();
    al_install_audio();
    al_init_acodec_addon();
    al_reserve_samples(8);
    return 0;
}
```


SiegeLord:
Thanks!

PR #817, opened 2017-08-28
Use unique ids to identify menu items internally.

### LINKED ISSUE #816: Reusing ids with different menus leads to confusion with respect to event sources.
This is due to Allegro using the id to do the reverse lookup to the menu item that generated the event.

Two action points here:

- [ ] document the status quo better
- [ ] try to fix it. On every platform except Windows we have complete control over the id type.

SiegeLord:
Fixes #816.

### DIFF
--- a/addons/native_dialog/allegro5/internal/aintern_native_dialog.h
+++ b/addons/native_dialog/allegro5/internal/aintern_native_dialog.h
@@ -64,10 +64,26 @@ struct ALLEGRO_MENU_ITEM
    ALLEGRO_MENU *popup;
    ALLEGRO_USTR *caption;
    ALLEGRO_BITMAP *icon;
-   
+
+   /* This id is unique. */
+   uint16_t unique_id;
+   /* This id is user provided, can be not unique. */
    uint16_t id;
+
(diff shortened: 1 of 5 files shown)


oitofelix:
Do you think this change might somehow break code relying on the previous non-unique behavior, or is it safe as a drop-in replacement?  I have a huge menu module for my game on which I've spent quite some time and I don't want to revisit it for now.

Ps: By the way, thank you for working on improving Allegro's menu interface!

fatcerberus:
@oitofelix I suspect most people would consider the previous behavior a bug, like you did.

SiegeLord:
This patch has two effects, making the `al_enable_menu_event_source` work for multiple menus, and also making the events' menu pointer point correctly to the new menu. The only way I can imagine this breaking code is if people somehow depended on that menu pointer not changing. It seems unlikely enough to me. I figure most people just compare the returned ids and not realize something's funky is going on under the hood.

SiegeLord:
Tested it on OSX, Windows and Linux... everything seems to work fine. Merging!

PR #819, opened 2017-08-31
add a gradle based Android build

elias-pschernig:
This allows building Allegro with newer SDKs where ant is not available any longer.

The way to add gradle to a project is to add their so called "gradle wrapper" into version control, which then on the first run will download the right version of itself. It's made by Google...


README_android.txt
```diff
@@ -7,9 +7,8 @@ This port should support Android 3.1 (Honeycomb) and above.
 Dependencies
 ============
 
-This port depends on having CMake, the Android SDK, and the Android NDK
-version r5b or better, JDK, and Apache ant. Python is required to build
-the examples.
+This port depends on having CMake, the Android SDK, the Android NDK and
+the Oracle JDK.
```
dos1: Oracle JDK? Why?

README_android.txt
```diff
@@ -19,26 +18,17 @@ Install the SDK
...
 
 Install the NDK
 ===============
 
 
-Extract the NDK and add it to your PATH. If you installed the NDK in
-$HOME/android-ndk, add $HOME/android-ndk to your PATH.
+The best way is again to use Android Studio. Create a new project with
+C++ support and it will ask you if you want to install the NDK and will
+then proceed to place it into ~/Android/ndk-bundle.
```
dos1: It's just a matter of downloading and extracting the archive... Thanks to previous instructions, I've easily and successfully managed to build Allegro for Android and integrate my cmake-based build system to automatically make APKs of my games, without installing the huge Android Studio at all. I don't feel like this is much of an upgrade :(

README_android.txt
elias-pschernig: All of Google's Android tools run on Java including gradle, so you do need Java. In theory any JDK should work, but for me at least Open JDK just crashes when running gradle while the Oracle one works.

README_android.txt
elias-pschernig: The normal SDK and NDK also work! Also note that this PR is not finished and the current instructions will not work at all.

README_android.txt
```diff
@@ -47,18 +37,26 @@ Make NDK standalone toolchain
 Next you need to setup a standalone NDK toolchain. Set an environment
 variable to point to the desired location of the Android toolchain:
 
-    export ANDROID_NDK_TOOLCHAIN_ROOT=$HOME/android-toolchain
+    export ANDROID_NDK_TOOLCHAIN_ROOT=$HOME/android-toolchain-x86_64
```
dos1: I think arm should be the default in instructions. Building APKs for x86(_64) is a niche case.

README_android.txt
dos1: Yes; I mean, I'm always using OpenJDK and staying away of Oracle's one, as it's not free, hence my question.

README_android.txt
dos1: Sure it works, but discovering how to make it work might be harder :) Just adding my two cents, as I was recently going through all the hassle of adding Android support to my Allegro based engine.

README_android.txt
elias-pschernig: I'll remove the "Oracle" - hopefully whatever crash I got is fixed in the meantime!

README_android.txt
elias-pschernig: The normal Allegro build builds an APK for each example and demo, without any Android Studio. I will add a note that you could just copy one of those as a template instead.

README_android.txt
elias-pschernig: I think I made a note somewhere of all the different architectures you can build for and everything is exactly the same for each. But yes, I can change the default. I was just thinking that people might not have an Android device and then are surprised how slow their game is in the emulator, while if they build with x86_64 the emulator will run at native speed.

beoran:
I see you dropped the python script that could be used to generate an android project. Does CMake do all of that now? The README_android.txt is helpful but I think that ideally all of that should be automated, one way or another, perhaps with CMake since you already need CMake to build Allegro anyway.

elias-pschernig:
Yes, it's all done by cmake now. That Python script was used internally by the old cmake configuration to compile the examples for Android (it was never used by users and the old readme does not even mention it).

Which part of README_android.txt do you think should be more automated? Many of the steps tell you to install the SDK and NDK and setup your cross-compiler toolchain and compile dependencies. That is not easy things to do. But also not something cmake can do for you. Compiling allegro itself is just "mkdir build; cd build; cmake; make; make install" like on most other platforms.

The section about using Allegro in Android Studio is completely optional, you can just copy any of the examples or demos as a template and replace with your own game's source code. Maybe I should make a more prominent note about that?

elias-pschernig:
I just swapped the last two sections, so now right after running "make" to compile Allegro you will be at the section telling you that you can just copy the template. And then if you like pain you can skip ahead and try and set up a new project from scratch instead :)

README_android.txt
```diff
@@ -19,26 +18,21 @@ Install the SDK
...
 
-Once extracted you must run "android" (or android.bat on Windows) and install
-a version of the SDK. Allegro needs version 13 (3.1) of the SDK or higher to
-build. It can run on version 9 and up still though, if you include the
-correct uses-sdk fields in your manifest. See:
-https://developer.android.com/guide/topics/manifest/uses-sdk-element.html
-
-Also make sure the JAVA_HOME environment variable is set to your JDK
-directory (not always the case.)
+Alternatively you can also download the command-line SDK tools.
```
SiegeLord: Probably should mention ALLEGRO_HOME enviroment variable here.

Also, I had to accept the licenses before it would let me build, maybe add that as an optional step:

~/Android/Sdk/tools/bin/sdkmanager --licenses

README_android.txt
```diff
@@ -47,18 +41,34 @@ Make NDK standalone toolchain
...
+amd x86_64 (the Android emulator).
 
-Assuming the NDK was extracted into $HOME/android-ndk run the following
+Assuming the NDK was extracted into ~/Android/Sdk/ndk-bundle run the following
 command:
 
-    $HOME/android-ndk/build/tools/make-standalone-toolchain.sh \
-        --platform=android-9 --install-dir=$ANDROID_NDK_TOOLCHAIN_ROOT
+    python ~/Android/Sdk/ndk-bundle/build/tools/make_standalone_toolchain.py \
+        --api=15 -install-dir=$ANDROID_NDK_TOOLCHAIN_ROOT --arch=arm
```
SiegeLord: --arch=arm doesn't work, as 'arm' is invalid according to the toolchain file. I had to do armabi. Should we change the readme or the toolchain file?

SiegeLord:
I tried this, and it was definitely easier to get it running. Unfortunately, all the examples I tried crashed for me. I did `make run_ex_prim`. Here's a potentially relevant bit of logs:

```
09-30 21:05:21.676 28989 28989 E AndroidRuntime: FATAL EXCEPTION: main
09-30 21:05:21.676 28989 28989 E AndroidRuntime: Process: org.liballeg.app, PID: 28989
09-30 21:05:21.676 28989 28989 E AndroidRuntime: java.lang.UnsatisfiedLinkError: dalvik.system.PathClassLoader[DexPathList[[zip file "/data/app/org.liballeg.app-2/base.apk"],nativeLibraryDirectories=[/data/app/org.liballeg.app-2/lib/arm, /data/app/org.liballeg.app-2/base.apk!/lib/armeabi, /vendor/lib, /system/lib]]] couldn't find "liballegro_ttf-debug.so"
09-30 21:05:21.676 28989 28989 E AndroidRuntime: 	at java.lang.Runtime.loadLibrary(Runtime.java:367)
09-30 21:05:21.676 28989 28989 E AndroidRuntime: 	at java.lang.System.loadLibrary(System.java:1076)
09-30 21:05:21.676 28989 28989 E AndroidRuntime: 	at org.liballeg.app.MainActivity.<clinit>(MainActivity.java:9)
09-30 21:05:21.676 28989 28989 E AndroidRuntime: 	at java.lang.Class.newInstance(Native Method)
09-30 21:05:21.676 28989 28989 E AndroidRuntime: 	at android.app.Instrumentation.newActivity(Instrumentation.java:1090)
09-30 21:05:21.676 28989 28989 E AndroidRuntime: 	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2327)
09-30 21:05:21.676 28989 28989 E AndroidRuntime: 	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2490)
09-30 21:05:21.676 28989 28989 E AndroidRuntime: 	at android.app.ActivityThread.-wrap11(ActivityThread.java)
09-30 21:05:21.676 28989 28989 E AndroidRuntime: 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1354)
09-30 21:05:21.676 28989 28989 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:102)
09-30 21:05:21.676 28989 28989 E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:148)
09-30 21:05:21.676 28989 28989 E AndroidRuntime: 	at android.app.ActivityThread.main(ActivityThread.java:5443)
09-30 21:05:21.676 28989 28989 E AndroidRuntime: 	at java.lang.reflect.Method.invoke(Native Method)
09-30 21:05:21.676 28989 28989 E AndroidRuntime: 	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:728)
09-30 21:05:21.676 28989 28989 E AndroidRuntime: 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:618)
09-30 21:05:21.693 21925 23205 W ActivityManager:   Force finishing activity org.liballeg.app/.MainActivity
```

Seems like it's looking for the ttf addon, but `ex_prim` doesn't depend on that (I don't have it, as I didn't bother with the freetype dependency yet).

README_android.txt
elias-pschernig: "arm" is right, see here: https://developer.android.com/ndk/guides/standalone_toolchain.html

Note that there are two actual architectures using the same "arm" toolchain: armeabi and armeabi-v7a.

What is the error you got?

elias-pschernig:
Oh, indeed - splitting the set of libraries according to what the example needs would be nice. Unfortunately required a lot of cmake trickery, but I'll try and add it back.

This also makes me realize I have yet to add back that special Android example (but I'll convert it to a normal example).

README_android.txt
elias-pschernig: I just tried here again and fixed a typo, -install-dir needs to be --install-dir. But "--arch=arm" seems to work for me.

SiegeLord: Whoops, my bad. I meant in the cmake command some lines below. It says `-DARM_TARGETS=arm`, but that doesn't work.

elias-pschernig:
I fixed the missing liballegro_ttf.so in a different way (I belive). I still make the above call in all examples to load the .so file, but just ignore when it is missing - that was much simpler than trying to have cmake output a different .java file.
Also, since we now build with gradle the location of assets changed, so had to fix the pass where to place the data/ folder as well. After that ex_prim and the skater demo both work. (Well, to the extent they can, given they do not handle things like screen orientation or touch input or backgrounding.)

android/gradle_project/app/src/main/java/org/liballeg/app/MainActivity.java
```diff
@@ -1,15 +1,39 @@
...
-        System.loadLibrary("allegro_acodec-debug");
-        System.loadLibrary("allegro_color-debug");
+        loadLibrary("allegro");
+        loadLibrary("allegro_primitives");
+        loadLibrary("allegro_image");
+        loadLibrary("allegro_font");
+        loadLibrary("allegro_ttf");
+        loadLibrary("allegro_audio");
+        loadLibrary("allegro_acodec");
+        loadLibrary("allegro_color");
```
dos1: memfile, physfs, and video addons are missing

PR #824, opened 2017-09-21
image: make compression/quality levels and gamma handling configurable

dos1:
With my 3200x1800 screen, Z_BEST_COMPRESSION meant that saving fullscreen
bitmap into PNG file took 15-20 seconds on my i5-4200U CPU.
With Z_DEFAULT_COMPRESSION it takes 3-5 seconds now and the resulting files
aren't much bigger (4.6M and 8.3M with DEFAULT vs. 4.5M and 8.2M with BEST
on the same test images).

Z_DEFAULT_COMPRESSION, per zlib's documentation, is a good compromise
between speed and size, while Z_BEST_COMPRESSION is extremely slow and
desirable only when trying to squeeze every bit out of the compressed file.
Seems like Z_DEFAULT_COMPRESSION would be a good default option to use by Allegro.

### DIFF
--- a/addons/image/iio.h
+++ b/addons/image/iio.h
@@ -28,11 +28,6 @@ typedef struct PalEntry {
  */
 extern double _al_png_screen_gamma;
 
-/* Choose zlib compression level for saving file.
- * Default is Z_BEST_COMPRESSION.
- */
-extern int _al_png_compression_level;
-
 
 
 #endif
--- a/addons/image/iio.h
+++ b/addons/image/iio.h
@@ -12,23 +12,5 @@ typedef struct PalEntry {
 } PalEntry;
 
 
-/* FIXME: Not sure if these should be made accessible. Hide them for now. */
-
-/* _al_png_screen_gamma is slightly overloaded (sorry):
- *
- * A value of 0.0 means: Don't do any gamma correction in load_png()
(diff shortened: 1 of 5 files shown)


elias-pschernig:
Default by default makes sense to me. But can we get a config option? Something like:

[image_addon]
png_compression_level=best

dos1:
Yes, we can!

elias-pschernig:
Awesome

dos1:
I went ahead and implemented similar config file entries for PNG gamma handling and JPEG quality setting :)

dos1:
Is there a documentation for allegro5.cfg entries anywhere outside of allegro5.cfg file itself?

beoran:
It hasn't been done consistently, but I feel this should be documented in https://github.com/liballeg/allegro5/blob/master/docs/src/refman/image.txt. Same for your webp pull request by the way.

SiegeLord:
The file is documented in http://liballeg.org/a5docs/trunk/system.html#al_get_system_config, but that's not very discoverable. It seems okay to mention this inside the image addon's documentation, as beoran suggests.

dos1:
Updated!

SiegeLord:
Merged, thanks for this!

fatcerberus:
This is nice, it will make my screenshot saving faster :)

PR #825, opened 2017-09-21
fix SDL platform build

dos1:
It's useful for testing GLES stuff on PC or checking the feasibility of porting Allegro into platforms without GLX. Although it's visibly experimental, it enabled me to experiment (hehe) a lot, so I thought that making it working out of box might be a good thing.

### DIFF
--- a/include/allegro5/platform/allegro_internal_sdl.h
+++ b/include/allegro5/platform/allegro_internal_sdl.h
@@ -23,7 +23,6 @@ typedef struct ALLEGRO_DISPLAY_SDL
    SDL_GLContext context;
 } ALLEGRO_DISPLAY_SDL;
 
-typedef struct ALLEGRO_SYSTEM_INTERFACE ALLEGRO_SYSTEM_INTERFACE;
 ALLEGRO_SYSTEM_INTERFACE *_al_sdl_system_driver(void);
 ALLEGRO_DISPLAY_INTERFACE *_al_sdl_display_driver(void);
 ALLEGRO_KEYBOARD_DRIVER *_al_sdl_keyboard_driver(void);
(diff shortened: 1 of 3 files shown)


include/allegro5/platform/allegro_internal_sdl.h
```diff
@@ -23,7 +23,6 @@ typedef struct ALLEGRO_DISPLAY_SDL
    SDL_GLContext context;
 } ALLEGRO_DISPLAY_SDL;
 
-typedef struct ALLEGRO_SYSTEM_INTERFACE ALLEGRO_SYSTEM_INTERFACE;
```
beoran: Why exactly did you remove this?

beoran:
Well, a better FindSdl2 seems really useful. Thanks for your efforts.

include/allegro5/platform/allegro_internal_sdl.h
dos1: This struct is already typedefed in `internal/aintern_system.h`; it's not really related to `allegro_internal_sdl.h` and it was breaking the build on some old GCC in one of my toolchains. It seemed to me that it was here by mistake and while newer compilers seem to be smart enough to ignore duplicated typedef, I thought that it's still cleaner to just take this out.

beoran: Ok, thanks for clarifying that for me. I agree it's probably better to remove that duplicate typedef.

SiegeLord:
What's the source of that FindSDL2.cmake file? Are we allowed to just copy it in?

dos1:
I have tracked it down to https://github.com/Kitware/VTK/blob/master/CMake/FindSDL2.cmake , which asked to substitute the short license info with full license text when distributing outside of cmake or other Kitware projects, which I did.

SiegeLord:
Merged as 044a918c287db98ff7f98523a5d1c5c43b333973, thanks!

PR #827, opened 2017-09-22
image: add WebP support with libwebp

dos1:
[WebP](https://developers.google.com/speed/webp/) is a VP8-based image format by Google. It supports both lossy and lossless encoding schemes and allows for alpha channel with both of them.
Also, lossless compression usually gives much smaller file sizes when compared to PNG.

I think supporting a image codec that can work as "JPEG with alpha" can be pretty useful for game development :)


addons/image/iio.c
```diff
@@ -87,6 +87,7 @@ bool al_init_image_addon(void)
          success |= al_register_bitmap_loader_f(extensions[i], _al_load_android_bitmap_f);
          //success |= al_register_bitmap_saver(extensions[i], _al_save_android_bitmap);
       }
+      success |= al_register_bitmap_identifier(".webp", _al_identify_webp);
```
elias-pschernig: Why is this done? (inside of ALLEGRO_CFG_IIO_HAVE_ANDROID)

dos1: Android supports WebP with system facilities and _al_identify_webp doesn't need libwebp, so I've added it there as well to make WebP identification work on Android too.

CMakeLists.txt
```diff
@@ -133,6 +133,7 @@ option(WANT_IMAGE "Enable image load/save addon" on)
 if (NOT IPHONE)
     option(WANT_IMAGE_JPG "Enable JPEG support in image addon" on)
     option(WANT_IMAGE_PNG "Enable PNG support in image addon" on)
+    option(WANT_IMAGE_WEBP "Enable WebP support in image addon" on)
```
beoran: Is libwebp not supported on iOS? Otherwise this should go outside of the iOS define.

addons/image/webp.c
```diff
@@ -0,0 +1,174 @@
...
+   lock = al_lock_bitmap(bmp, ALLEGRO_PIXEL_FORMAT_ABGR_8888_LE,
+      ALLEGRO_LOCK_WRITEONLY);
+
+   config.output.colorspace = premul ? MODE_rgbA : MODE_RGBA;
+   config.output.u.RGBA.rgba = (uint8_t*)lock->data;
+   config.output.u.RGBA.stride = lock->pitch;
+   config.output.u.RGBA.size = lock->pitch * config.input.height;
+   config.output.is_external_memory = 1;
+
+   if (WebPDecode(data, data_size, &config) != VP8_STATUS_OK) {
```
beoran: Now, I am convinced that libwebp is very easy to use compared to libpng...😏

addons/image/webp.c
```diff
@@ -0,0 +1,174 @@
...
+   lock = al_lock_bitmap(bmp, ALLEGRO_PIXEL_FORMAT_ABGR_8888_LE,
+      ALLEGRO_LOCK_READONLY);
+   if (!lock)
+      return false;
+
+   uint8_t *output = NULL;
+   size_t size;
+
+   const char* level = al_get_config_value(al_get_system_config(), "image", "webp_quality_level");
+   if (level && strcmp(level, "lossless") == 0) {
```
beoran: Since Allegro is a game library, where one might expect saved images to be pixel perfect,  I think lossless compression should be the default, and lossy compression an option.

addons/image/webp.c
```diff
@@ -0,0 +1,174 @@
...
+   ALLEGRO_ASSERT(bmp);
+
+   fp = al_fopen(filename, "wb");
+   if (!fp) {
+      ALLEGRO_ERROR("Unable to open file %s for writing\n", filename);
+      return false;
+   }
+
+   retsave = _al_save_webp_f(fp, bmp);
+   retclose = al_fclose(fp);
```
beoran: I would distinguish the two error cases with a different error message in stead of conflating them.

addons/image/webp.c
dos1: And that's "advanced API" - it could be even easier if not for premultiplied alpha :D 
https://developers.google.com/speed/webp/docs/api#simple_decoding_api

CMakeLists.txt
dos1: I don't really understand this construct to be honest and left it there as an uninformed guess. Why are JPG and PNG wrapped in "NOT IPHONE" conditional? I see that Allegro on iOS uses system provided facilities to load image files, but it does so on macOS, Android and Windows as well. What's so different about the iPhone?

addons/image/webp.c
dos1: This code is modelled after existing PNG and JPEG savers. Should it be changed in the other ones as well?

dos1:
Updated!
Docs are handled by #824 already.

addons/image/iio.c
elias-pschernig: Don't you still have to register .webp somehow for al_load_bitmap to know to use those system facilities? I don't doubt it works, I'm just curious how exactly...

dos1: Look at the loop, a few lines above. It was already there :)

elias-pschernig: Oh, interesting, I did not know that.

SiegeLord:
This doesn't link for me on Ubuntu:

```
../lib/liballegro_image.so.5.2.3: undefined reference to `WebPFree'
```

This is using Ubuntu's libwebp, version 0.4.4 (as far as I can tell). Could you also add a line to install this for our CI script? It's on line 16, in the `.travis.yml` file. Add a `libwebp-dev` there.

dos1:
Looks like WebPFree has been added in 0.5.0 - https://github.com/webmproject/libwebp/commit/822f113ebbe8810a1de0f11062faaf7edb0556e3 - to simplify bindings to managed languages.
However, it does nothing aside of calling free(), so I've replaced it with free call.

Travis config updated. Also, added tests :)

fatcerberus:
Hmm... in the future the function could end up doing more than just calling free(), so it would probably be better to do a version check using the preprocessor and always call WebPFree() in versions where it’s available.  Assuming that’s possible at least...

dos1:
That's what I wanted to do at the beginning, but then I found out that the reason why they introduced it was just for .NET bindings to be simpler.
Anyway, for the peace of mind, I've added the check :)

SiegeLord:
Alright, merged! Thanks a lot!

PR #828, opened 2017-09-25
allegro5.cfg: fix incorrect config file name referenced in the comments

dos1:
I've got burned by this once or twice already :P

### DIFF
--- a/docs/src/refman/system.txt
+++ b/docs/src/refman/system.txt
@@ -206,6 +206,8 @@ from a few different locations, in this order:
 
 - *Unix only:* $HOME/allegro5rc
 
+- *Unix only:* $HOME/.allegro5rc
+
 - allegro5.cfg next to the executable
 
 If multiple copies are found, then they are merged using [al_merge_config_into].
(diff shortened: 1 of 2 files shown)


elias-pschernig:
Ha! How did you find that out? Anyway, problem with your "fix" is that the documentation is actually correct, but there is a typo in the code. I.e. the file was named ~/.allegrorc in Allegro 4 and was meant to keep the leading dot with Allegro 5: ~/.allegro5rc (just like for example ~/.bashrc or ~/.vimrc or any ancient unix programs).

SiegeLord:
Let's just support both, since maybe there's some person relying on this.

dos1:
Haha. I found it out by wondering why my .allegro5rc file wasn't taken into effect. I've checked and turned out both documentation and the code says "allegro5rc" while the comments in the config file itself say ".allegro5rc". Yes, the version with dot is what I'd expect, I'll modify this PR :)

dos1:
Updated!

SiegeLord:
Merged as dd285c8d089e35ea2a94ca0e295c9806880e6901, thanks!

PR #829, opened 2017-09-25
docs: audio: note that callback set by al_set_mixer_postprocess_callback runs in separate thread

dos1:
Another thing that has already bitten me in the past :D


elias-pschernig:
I would prefer to use only the first sentence.

Maybe replace the second sentence with something like "Unless you set a target display or bitmap this means you cannot use any drawing functions." But I think that should go without saying.

And the last part, mentioning to "properly synchronize" data is kinda pointless - either they know how threads work and will do so already, or they will not understand what "properly synchronize" even means :)

dos1:
Updated!

SiegeLord:
Merged as dc95e93495ffa8342b95c59f8abaa519e0495cce, thanks!

PR #837, opened 2017-10-02
Enable blocked locking of S3TC formats by default.

dos1:
The patent has expired today! \o/


SiegeLord:
Nice. I was planning to remove this setting entirely, would you like to do that as well? It's just a few defines in the OpenGL code somewhere.

dos1:
Ah, couldn't be too easy, eh? :D 
There it goes :)

SiegeLord:
Woohoo!

PR #839, opened 2017-10-05
fix dumb_off_t recognition

SimonN:
`dumb_off_t` is not a macro, it is a typedef. The preprocessor can't
see `dumb_off_t` even in DUMB 2.0.0, but I assumed it could.

And let's use a typedef to get proper errors.


SiegeLord:
Thanks!

PR #841, opened 2017-10-07
fix warning

elias-pschernig:


SiegeLord:
That's pretty funny. I like how the code actually computed the correct value in a roundabout way.

SiegeLord:
Merged as https://github.com/liballeg/allegro5/commit/020f3fa394f77836c97da2fc6c857e178afbd4cf.

PR #848, opened 2017-10-19
run Python 2 scripts in misc/ with python2 and regenerate

dos1:
On more and more systems "python" actually means python3, which
doesn't work with those scripts yet.


dos1:
If somebody doesn't know it yet, you can append `?w=1` to github's diff URL and skip simple whitespace changes. Handy to look at that scanline_drawers.inc diff here :)
https://github.com/liballeg/allegro5/pull/848/files?w=1

fatcerberus:
`scanline_drawers.inc` is autogenerated is it not?

elias-pschernig:
Python2 is really outdated at this point - I'd just convert those scripts to Python3.

dos1:
@fatcerberus: Yes, it is; not on build though.
@elias-pschernig: I tried it by fixing print invocations, but running them with python3 produced different output than with python2, so I opted for a quick fix for now instead.

SiegeLord:
Thanks!

PR #854, opened 2017-10-28
Fix `ptr` assert in file.c

fatcerberus:
According to the C specification `malloc(0)` can return a null pointer.  It turns out that this actually happens in practice, notably on Ubuntu.  Therefore NULL is actually a valid buffer pointer, *if and only if* size == 0.  This adjusts the assertions in al_fread() and al_fwrite() to align.


SiegeLord:
Thanks!

PR #858, opened 2017-11-19
[font addon] add support for bmfont format

elias-pschernig:
A lot of libraries like libgdx  or Phaser use a bitmap font format (bmfont) slightly different and more feature-rich than Allegro's (multiple ranges per bitmap, kerning pairs). This is achieved by having an .xml file with some meta data in addition to the bitmaps.

For some reason I needed to load a bmfont font in Allegro and so this patch adds support for bmfont fonts.

(If this PR gets applied we could probably simplify font.c by just calling functions from bmfont.c because Allegro's bitmap fonts are just a subset of bmfont fonts.)


fatcerberus:
Never heard of this format, kerning and such is nice to have.

addons/font/bmfont.c
```diff
@@ -0,0 +1,498 @@
...
+static bool tag_is(BMFONT_DATA *data, char const *str) {
+   return strcmp(al_cstr(data->tag), str) == 0;
+}
+
+static bool attribute_is(BMFONT_DATA *data, char const *str) {
+   return strcmp(al_cstr(data->attribute), str) == 0;
+}
+
+static int get_int(char const *value) {
+   return strtol(value, NULL, 0);
```
SiegeLord: Why not force base 10?

addons/font/xml.c
```diff
@@ -0,0 +1,138 @@
...
+
+static void create_tag(XmlParser *x) {
+   scalar(x);
+}
+
+static void open_tag(XmlParser *x) {
+   x->state = Outside;
+}
+
+void _al_xml_parse(ALLEGRO_FILE *f,
```
SiegeLord: Is this a full XML parser or some subset of the language?

addons/font/bmfont.c
```diff
@@ -0,0 +1,498 @@
...
+   BMFONT_CHAR **characters;
+   BMFONT_RANGE *next;
+};
+
+typedef struct {
+   ALLEGRO_USTR *tag;
+   ALLEGRO_USTR *attribute;
+   int pages_count;
+   ALLEGRO_BITMAP **pages;
+   BMFONT_RANGE *range_first;
```
SiegeLord: Curious, why use a linked list?

addons/font/bmfont.c
```diff
@@ -0,0 +1,498 @@
...
+   BMFONT_CHAR *prev = find_codepoint(data, codepoint1);
+   BMFONT_CHAR *c = find_codepoint(data, codepoint2);
+   if (!c) {
+      if (!f->fallback) return 0;
+      return al_get_glyph_advance(f->fallback, codepoint1, codepoint2);
+   }
+   return c->xadvance + get_kerning(prev, codepoint2);
+}
+
+static bool get_glyph (const ALLEGRO_FONT *f, int prev_codepoint,
```
SiegeLord: Extraneous space after get_glyph

addons/font/bmfont.c
```diff
@@ -0,0 +1,498 @@
...
+
+struct BMFONT_RANGE {
+   int first;
+   int count;
+   BMFONT_CHAR **characters;
+   BMFONT_RANGE *next;
+};
+
+typedef struct {
+   ALLEGRO_USTR *tag;
```
SiegeLord: Maybe it'd be clearer if this mentioned that these were ephemeral for parsing... or maybe even there should be a dedicated structure for parsing.

addons/font/bmfont.c
```diff
@@ -0,0 +1,498 @@
...
+      range = range->next;
+   }
+   return NULL;
+}
+
+static void add_page(BMFONT_DATA *data, char const *filename) {
+   data->pages_count++;
+   data->pages = al_realloc(data->pages, data->pages_count *
+      sizeof *data->pages);
+   al_set_path_filename(data->path, filename);
```
SiegeLord: This doesn't do the right thing if 'filename' is an absolute path, right? Is that an impossibility? Maybe we should assert this?

addons/font/bmfont.c
elias-pschernig: Isn't that what Allegro's normal bitmap font uses? Thinking about it it might be possible to use a binary tree or something for faster lookup. My use case did not have a lot of separate ranges so a linked list was good enough.

addons/font/bmfont.c
elias-pschernig: The [specification](http://www.angelcode.com/products/bmfont/doc/file_format.html) calls it "filename". I interpreted that as no path allowed and the file **has** to be next to the .xml.

addons/font/xml.c
elias-pschernig: subset

elias-pschernig:
I've never seen one that is not XML, and it's the only format I required myself. It also should be easy converting from the other formats to XML. If someone needs one of the other formats it probably would not hurt adding support in that case.

addons/font/bmfont.c
elias-pschernig: I moved it to a dedicated structure.

addons/font/bmfont.c
```diff
@@ -0,0 +1,498 @@
...
+   ALLEGRO_BITMAP *page = al_load_bitmap_flags(
+      al_path_cstr(data->path, '/'), data->flags);
+   data->pages[data->pages_count - 1] = page;
+}
+
+static bool tag_is(BMFONT_DATA *data, char const *str) {
+   return strcmp(al_cstr(data->tag), str) == 0;
```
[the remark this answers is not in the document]
elias-pschernig: Then I would have to allocate a new ustr to compare to, wouldn't I?

addons/font/bmfont.c
elias-pschernig: I always use hexadecimal numbers myself - in this case it doesn't really matter I guess.

SiegeLord:
Do you have a bmfont handy that we can add to the data folder? Something small... I wonder if we can wrap one of the existing bitmap fonts?

elias-pschernig:
I wrapped a4_font.tga with an a4_font.fnt. To see it in action can do:

./ex_ttf data/a4_font.fnt

We could probably add a dedicated example or modify one of the examples already using a4_font.tga to instead load it as .fnt. In fact I think ex_vsync loads it wrong (characters after 127 will be mapped incorrectly to unicode 128+).

SiegeLord:
Alright, merged as https://github.com/liballeg/allegro5/commit/677b8a97c64f19c626b8fba1bce9899274bb5d4e. I had to fix the ids in the example font, as they were wrong past ASCII. Also, the `get_glyph_advance` wasn't correct, so I fixed that as well.

PR #864, opened 2017-12-06
add al_transform_coordinates_4d

### LINKED ISSUE #863: al_transform_coordinates_3d doesn't work with perspective matrices
I'm not sure how to fix this regarding the API compatibility, so I'll just open an issue instead. If you tell me what to do I can make a PR afterwards ;)

When using transformation matrix with fourth column different than "0 0 0 1" (like the one made by `al_perspective_transform`), `al_transform_coordinates_3d` doesn't take the fourth coordinate (w) into account, resulting in invalid data.

Either:
* `al_transform_coordinates_3d` should already return values transformed to Cartesian coordinates (divided by w)
* `al_transform_coordinates_3d` should be modified to also return the fourth value (w), so user can divide the others (x, y, z) by himself
* documentation should take a note about situations with non-one w, like perspective transforms, so users won't try to use this function when it won't work and will do the calculations by themselves

elias-pschernig:
Fixes #863

### DIFF
--- a/include/allegro5/transformations.h
+++ b/include/allegro5/transformations.h
@@ -34,6 +34,10 @@ AL_FUNC(void, al_scale_transform_3d, (ALLEGRO_TRANSFORM *trans, float sx, float 
 AL_FUNC(void, al_transform_coordinates, (const ALLEGRO_TRANSFORM* trans, float* x, float* y));
 AL_FUNC(void, al_transform_coordinates_3d, (const ALLEGRO_TRANSFORM *trans,
    float *x, float *y, float *z));
+   AL_FUNC(void, al_transform_coordinates_4d, (const ALLEGRO_TRANSFORM *trans,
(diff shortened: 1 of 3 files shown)


include/allegro5/transformations.h
```diff
@@ -34,6 +34,10 @@ AL_FUNC(void, al_scale_transform_3d, (ALLEGRO_TRANSFORM *trans, float sx, float
 AL_FUNC(void, al_transform_coordinates, (const ALLEGRO_TRANSFORM* trans, float* x, float* y));
 AL_FUNC(void, al_transform_coordinates_3d, (const ALLEGRO_TRANSFORM *trans,
    float *x, float *y, float *z));
+   AL_FUNC(void, al_transform_coordinates_4d, (const ALLEGRO_TRANSFORM *trans,
```
SiegeLord: Extra leading tab here

PR #871, opened 2017-12-17
Fix 3D (non-projection) transforms with al_hold_bitmap_drawing

dos1:
The z coordinate set by current transform was dropped when calculating
the vertex coords, so it couldn't reach the projective transform residing
at the GPU.

This isn't an issue when not held, since there is no other way in Allegro
to set the z coordinate than using transforms, and without holding,
both transforms are applied already on the GPU using all four coordinates.

Note that projection transforms are still unsupported.

### DIFF
--- a/include/allegro5/internal/aintern_opengl.h
+++ b/include/allegro5/internal/aintern_opengl.h
@@ -153,7 +153,7 @@ typedef struct ALLEGRO_OGL_EXTRAS
 
 typedef struct ALLEGRO_OGL_BITMAP_VERTEX
 {
-   float x, y;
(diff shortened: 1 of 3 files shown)


SiegeLord:
Before merging this, I'll want some benchmarks done to measure the effect of this. Not saying that you have to do this, but just something that will need to be done. I'll get to it eventually, naturally.

SiegeLord:
I tried benchmarking this, but observed no difference. Sounds like this is good to merge.

Note that an equivalent change needs to be made for Direct3D, but I can do that.

PR #879, opened 2018-01-20
ljoynu: fix build with older Android NDKs

dos1:
Commit 76f076e70de5e42eb2122b56e5367f9e3550bd9e removed those defines
to fix some warnings - probably about symbol redefinition. However,
on my setup with NDK r14b and build tools 26.0.1, those defines were
in fact needed and that commit made the build fail. Let's reintroduce
them, but this time guarded by ifndefs, so they don't fail builds on
older NDKs and don't generate warnings on newer ones.

### DIFF
--- a/src/linux/ljoynu.c
+++ b/src/linux/ljoynu.c
@@ -47,6 +47,14 @@
    #include <sys/inotify.h>
 #endif
 
+#ifndef KEY_CNT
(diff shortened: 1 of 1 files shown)


SiegeLord:
Works for me. Thanks for following up on this!

PR #885, opened 2018-02-06
X11: make compositor bypass configurable, and bypass only when fullscreen by default

### LINKED ISSUE #818: Make bypassing the compositor optional.
Currently on X, we bypass the compositor unconditionally:

https://github.com/liballeg/allegro5/blob/master/src/x/xdisplay.c#L249

while this makes sense for fullscreen games, this is a bit disruptive for windowed ones. It makes sense to me to make this configurable. The way this can be surfaced to users is via the system config, something like this:

```ini
[x11]
# Can be fullscreen_only, always, never
bypass_compositor = fullscreen_only
```

Alternatively, this could be a display option, but I don't know what this looks like on non-X platforms.

dos1:
Closes #818

### DIFF
--- a/allegro5.cfg
+++ b/allegro5.cfg
@@ -181,6 +181,10 @@ timestamps=1
 # Set to 0 to disable function names in log files.
 functions=1
 
+[x11]
+# Can be fullscreen_only, always, never
+bypass_compositor = fullscreen_only
+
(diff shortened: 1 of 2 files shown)


SiegeLord:
Nice, thanks!

PR #886, opened 2018-02-08
docs: take note that al_invert_transform and friends work only with 2D transforms

dos1:
It might be useful to add al_invert_transform_3d and al_invert_transform_4d sometime, but for now just document that al_invert_transform doesn't work with 3- and 4- dimensional transforms.

### DIFF
--- a/docs/src/refman/transformations.txt
+++ b/docs/src/refman/transformations.txt
@@ -202,6 +202,9 @@ If there is no target bitmap, this function returns NULL.
 This is similar to calling `al_invert_transform(al_get_current_transform())`
 but the result of this function is cached.
 
+> *Note*: Allegro's transformation inversion functions work correctly only 
+with 2D transformations.
+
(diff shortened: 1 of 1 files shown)


docs/src/refman/transformations.txt
```diff
@@ -202,6 +202,9 @@ If there is no target bitmap, this function returns NULL.
 This is similar to calling `al_invert_transform(al_get_current_transform())`
 but the result of this function is cached.
 
+> *Note*: Allegro's transformation inversion functions work correctly only 
+with 2D transformations (2x2 matrices).
```
SiegeLord: The 2x2 matrix comment isn't actually right, as 2D transformations are 2x3 as we have the translation there as well. I'd just remove this parenthetical.

dos1:
Oh, right! Changed.

SiegeLord:
Thanks!

PR #890, opened 2018-02-26
Add deps/ to .gitignore

fatcerberus:
The CMake build is set up to pull dependencies from `deps/`.  Because of that `git diff` gets polluted with all the includes there making it basically useless.  Since we never want to commit these anyway, I figured we might want to add the whole directory to `.gitignore`.


SiegeLord:
Thanks!

PR #897, opened 2018-03-24
fix allegro_opengl.h comment

byteblob:
I know it's just the comment, but that codepath is not the macos one. I've tried in windows and it's taken there, so I've changed the comment :p


SiegeLord:
Thanks! We won't be merging this as #826 will restructure that header a little bit anyway, and do a somewhat alternate change for these comments.

PR #901, opened 2018-04-13
improve linux joystick detection, thanks to imp

elias-pschernig:
Before only devices named /dev/input/event0 through /dev/input/event31
were detected. But a device can also have a name like
/dev/input/even260 and Allegro would not see it in that case (unless
with an entry in allegro.cfg or with hotplugging if the user
disconnected and reconnected it).


SiegeLord:
Merged as 56f087ab966203fe093f440ec8ea4f4765d65e9c. I never released that this was different from that other pull request! I tested this, and it appeared to keep working.

PR #902, opened 2018-04-14
improve linux joystick detection

allefant:
This correctly identifies X/Y/Z axes as well as common button types.
More work needs to be done but this also adds enough log information
that someone could just send us an allegro.log and we could improve from
that.

My old xbox 360 controller has 100% of axes and buttons correctly
identified with his.

Before the patch all the buttons were just named B and the axes were
mapped completely wrong, mapping triggers as axes and swapping X and Y
axes of the sticks.

### DIFF
--- a/examples/ex_joystick_events.c
+++ b/examples/ex_joystick_events.c
@@ -115,9 +115,18 @@ static void draw_joystick_button(ALLEGRO_JOYSTICK *joy, int button, bool down)
 
    if (joy) {
       const char *name = al_get_joystick_button_name(joy, button);
-      if (strlen(name) < 4) {
-         al_draw_text(font, fg, x + 13, y + 8, ALLEGRO_ALIGN_CENTRE, name);
+      char name2[4];
+      if (strlen(name) >= 4) {
+         name2[0] = name[0];
+         name2[1] = name[1];
+         name2[2] = name[2];
+         name2[3] = 0;
       }
+      else {
+         strcpy(name2, name);
+      }
+      al_draw_text(font, fg, x + 13, y + 8, ALLEGRO_ALIGN_CENTRE, name2);
+      
    }
 }
 
--- a/src/linux/ljoynu.c
+++ b/src/linux/ljoynu.c
@@ -181,6 +181,39 @@ static bool is_single_axis_throttle(int i)
 
 
 
+static bool is_x_axis(int i)
+{
+   return i == ABS_X
+       || i == ABS_RX
+       || i == ABS_HAT0X
+       || i == ABS_HAT1X
+       || i == ABS_HAT2X
+       || i == ABS_HAT3X
+       || i == ABS_TILT_X;
+}
+
+
+
+static bool is_y_axis(int i)
+{
+   return i == ABS_Y
+       || i == ABS_RY
+       || i == ABS_HAT0Y
+       || i == ABS_HAT1Y
+       || i == ABS_HAT2Y
+       || i == ABS_HAT3Y
+       || i == ABS_TILT_Y;
+}
+
+
+static bool is_z_axis(int i)
+{
+   return i == ABS_Z
+       || i == ABS_RZ;
+}
+
+
+
 static bool is_hat_axis(int i)
 {
    return (i >= ABS_HAT0X && i <= ABS_HAT3Y);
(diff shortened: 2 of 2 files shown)


fatcerberus:
Does this work properly with non-Xbox gamepads?  I know we've had issues with Xbox controllers in the past where attempting to fix them breaks things with other devices.

examples/ex_joystick_events.c
```diff
@@ -115,9 +115,18 @@ static void draw_joystick_button(ALLEGRO_JOYSTICK *joy, int button, bool down)
 
    if (joy) {
       const char *name = al_get_joystick_button_name(joy, button);
-      if (strlen(name) < 4) {
-         al_draw_text(font, fg, x + 13, y + 8, ALLEGRO_ALIGN_CENTRE, name);
+      char name2[4];
+      if (strlen(name) >= 4) {
```
igormorgado: is this not the case for a strncpy?

strncpy(name2, name, 3);

igormorgado:
about the axis auto discovery, I think (cannot test all joysticks in the world) that will probably break something, SDL went to the path to manual configuration instead auto discovery. They probably had a good reason for that.

fatcerberus:
> Still - do you have any more information about those issues you had in the past?

They weren't my issues--I just remember there being a good amount of discussion in the past w.r.t. incorrect behavior for Xbox controllers; I'm also pretty sure I remember a fix being reverted because it turned out to break other things.

examples/ex_joystick_events.c
allefant: I don't think so - strncpy will not copy the final 0.

allefant:
Let's see if someone can test it with one or two other joysticks. If something currently has 4 axes and designates them all as X axis, it would still show up as 2 sticks with X/Y axis each right now, since the current code completely ignores the designation. My chance would make it show up as 4 sticks with a single X axis each. So yes, I can see cases where something would end up worse than before if evdev returns wrong information to begin with. (With my only test device evdev returns correct information.)

I looked at the SDL axis/button database and it is very easy to use: https://github.com/gabomdq/SDL_GameControllerDB/blob/master/gamecontrollerdb.txt

And it also means the mapping would be identical under Windows and OSX so it would be a big advantage. However it seems SDL is actually using this only in their GameController API - their Joystick API works the same as ours and does not use that database at all, just a direct mapping. So we probably would have to investigate this a bit more and maybe also use an additional API for it.

beoran:
I found this Linux kernel doc on joypads that may be helpful. The problem
is that what the linux driver reports and what the button, trigger or axis
really is may not always be easy to map. Nevertheless this doc suggests a
way of doing it which we don't implement. Perhaps we should try what the
doc suggests.

https://www.kernel.org/doc/html/latest/input/gamepad.html

allefant:
Are we still not doing this with the PR? (except for the part about legacy drivers which is in fact keeping a custom database for those yourself)

beoran:
No, not quite. For game pads there seems to be a definite mapping for the
axes of the sticks and the triggers that neither the old code nor this pr
implements.

elias-pschernig:
I re-read it, I feel that's exactly what the PR does, maybe minus some smaller things. But the main problem is it seems two of SiegeLord's joystick simply do not follow the new Linux kernel guidelines at all, likely using some kind of legacy driver from before those gamepad specifications were written. The only way to make those work seems to be (as that document admits) to keep a database of device uuids and hardcode a mapping for them.

beoran:
Well, Allegro already used Allegro.cfg for configuring and dealing with all
sorts of hardware problems. I guess introducing some more settings for
Linux Joysticks is pretty much inevitable if we want all devices to work
well.

SiegeLord:
I'd argue that none of my controllers are mapped correctly. The correct mapping for F310 (and the XBox controller it is emulating) is 11 butons, 2 1D axes for the triggers and 2 2D axes for the analog sticks. I think we should hold off on merging this code until we got a remapping setup in place.

fatcerberus:
> 2 1D axes for the triggers

I note for the record that Windows itself seems to treat Xbox controllers as having a single 1D axis which is the sum of both triggers.  Kind of weird.  No idea if this is the same for other platforms.

SiegeLord:
Hmm... well, for me on Windows Allegro does treat F310's triggers as 1D sticks (interestingly, it treats the D-pad as separate 4 buttons, and doesn't detect the center button). Notable, F310 in DInput mode is not detected at all!

Issue #907, opened 2018-05-23
ex_physfs uses deprecated PHYSFS_addToSearchPath


snoopdouglas:
I've just rebuilt an Allegro project with Ubuntu 18.04 and have noted the following warning:

```
cl15_1  | boot.c: In function 'init':
cl15_1  | boot.c:79:3: warning: 'PHYSFS_addToSearchPath' is deprecated [-Wdeprecated-declarations]
cl15_1  |    PHYSFS_addToSearchPath(CL15_ASSETS_ARCHIVE, 0);
cl15_1  |    ^~~~~~~~~~~~~~~~~~~~~~
cl15_1  | In file included from main.h:10:0,
cl15_1  |                  from boot.h:4,
cl15_1  |                  from boot.c:1:
cl15_1  | /usr/include/physfs.h:851:17: note: declared here
cl15_1  |  PHYSFS_DECL int PHYSFS_addToSearchPath(const char *newDir, int appendToPath)
cl15_1  |                  ^~~~~~~~~~~~~~~~~~~~~~
```

The PhysicsFS package in Ubuntu 18.04 has been bumped to 3.0.1, which suggests [`PHYSFS_mount`](https://www.icculus.org/physfs/docs/html/physfs_8h.html#a8eb320e9af03dcdb4c05bbff3ea604d4) instead, so the example should be updated to use it.

SiegeLord:
Now fixed.

PR #908, opened 2018-05-23
update ex_physfs to use PHYSFS_mount

snoopdouglas:
was using PHYSFS_addToSearchPath (deprecated in v3)

#907


SiegeLord:
Thanks!

PR #910, opened 2018-06-06
opengl: don't force POT bitmaps on GLES

dos1:
Android devices that don't like NPOT FBOs are probably long gone and incompatible
with current versions of Allegro anyway.

If you need it, POT bitmaps can be always forced using Allegro configuration by
masking GL_ARB_texture_non_power_of_two and GL_OES_texture_npot extensions.

### DIFF
--- a/src/opengl/ogl_bitmap.c
+++ b/src/opengl/ogl_bitmap.c
@@ -1033,8 +1033,7 @@ ALLEGRO_BITMAP *_al_ogl_create_bitmap(ALLEGRO_DISPLAY *d, int w, int h,
       }
    }
 
-   /* Android included because some devices require POT FBOs */
-   if (IS_OPENGLES || !d->extra_settings.settings[ALLEGRO_SUPPORT_NPOT_BITMAP]) {
+   if (!d->extra_settings.settings[ALLEGRO_SUPPORT_NPOT_BITMAP]) {
       true_w = pot(true_w);
       true_h = pot(true_h);
    }


fatcerberus:
I hadn’t realized NPOT texture support was this pervasive, good to know.  I wonder if the same can be said for non-square textures...

dos1:
Well, even if they weren't so common, it's still only about the devices that don't like NPOT FBOs even though they export GL_OES_texture_npot extension, as otherwise Allegro will still make them POT :)

dos1:
I've got some issues on WebGL with this patch: `Non-power-of-two textures must have a wrap mode of CLAMP_TO_EDGE.`

Looks like there are some places where Allegro fails to set it, needs to be checked.

dos1:
I've looked into it. Actually, there's a bug in a completely different place: when using GLES, Allegro constructs extension list by using desktop GL data and determines support by comparing GLES context version to GL versions specified as ones that included particular extensions, which is obviously wrong.

This causes issues with Allegro thinking that NPOT textures are supported when they in fact are not (these days it's probably happening only under WebGL 1 though :P). I guess this was also the actual reason for putting that `IS_OPENGLES` check in the first place.

So, a GLES version of `allegro5/opengl/GLext/gl_ext_list.h` needs to be prepared and used in GLES builds. I think I'll put that into #826 and merge those two PRs.

SiegeLord:
What's the status of this with respect the WebGL comment you made?

dos1:
I haven't worked on that yet. However, I'd probably merge this anyway, as fixing false-positives when detecting NPOT support seems way less valuable these days than fixing big amounts of VRAM being wasted on any other GLES device (and on WebGL 2 it works just fine); but Allegro's extension handling still needs to be fixed some day, as it may subtly break some other things in similar way.

I certainly won't mind if this particular thing (providing a GLES version of `include/allegro5/opengl/GLext/gl_ext_list.h` file) will be done by anyone else though ;)

SiegeLord:
Makes sense, I filed a separate issue for the extension detection being broken and we can merge this in the interim.

PR #913, opened 2018-06-22
Add ability to query DPI of an attached screen

neuro-sys:
I needed to query DPI for at least Windows, Linux and MacOS platforms for my own stuff, here's the one for Linux, and MacOS and Windows. Although I'm not all too familiar with Allegro codebase, so I'm happy to improve this branch if you provide directions.

### DIFF
--- a/docs/src/refman/monitor.txt
+++ b/docs/src/refman/monitor.txt
@@ -60,6 +60,12 @@ Returns `true` on success, `false` on failure.
 
 See also: [ALLEGRO_MONITOR_INFO], [al_get_num_video_adapters]
 
+## API: al_get_monitor_dpi
+
+Get the dots per inch of a monitor attached to the display adapter.
+
+See also: [al_get_num_video_adapters]
+
 ## API: al_get_num_video_adapters
 
 Get the number of video "adapters" attached to the computer. Each
--- a/include/allegro5/monitor.h
+++ b/include/allegro5/monitor.h
@@ -24,6 +24,7 @@ enum {
 
 AL_FUNC(int, al_get_num_video_adapters, (void));
 AL_FUNC(bool, al_get_monitor_info, (int adapter, ALLEGRO_MONITOR_INFO *info));
+AL_FUNC(int, al_get_monitor_dpi, (int adapter));
 
 
 #ifdef __cplusplus
--- a/src/monitor.c
+++ b/src/monitor.c
@@ -48,5 +48,19 @@ bool al_get_monitor_info(int adapter, ALLEGRO_MONITOR_INFO *info)
    return false;
 }
 
+/* Function: al_get_monitor_dpi
+ */
+int al_get_monitor_dpi(int adapter)
+{
+   ALLEGRO_SYSTEM *system = al_get_system_driver();
+
+   if (adapter < al_get_num_video_adapters()) {
+       if (system && system->vt && system->vt->get_monitor_dpi) {
+           return system->vt->get_monitor_dpi(adapter);
+       }
+   }
+
+   return 0;
+}
 
 /* vim: set sts=3 sw=3 et: */
--- a/examples/ex_monitorinfo.c
+++ b/examples/ex_monitorinfo.c
@@ -25,7 +25,8 @@ int main(int argc, char **argv)
    for (i = 0; i < num_adapters; i++) {
       al_get_monitor_info(i, &info);
       log_printf("Adapter %d: ", i);
-      log_printf("(%d, %d) - (%d, %d)\n", info.x1, info.y1, info.x2, info.y2);
+      int dpi = al_get_monitor_dpi(i);
+      log_printf("(%d, %d) - (%d, %d) - dpi: %d\n", info.x1, info.y1, info.x2, info.y2, dpi);
       al_set_new_display_adapter(i);
       log_printf("   Available fullscreen display modes:\n");
       for (j = 0; j < al_get_num_display_modes(); j++) {
--- a/include/allegro5/internal/aintern.h
+++ b/include/allegro5/internal/aintern.h
@@ -47,6 +47,8 @@ AL_FUNC(int, _al_rand, (void));
 
 AL_FUNC(int, _al_stricmp, (const char *s1, const char *s2));
 
+#define _AL_INCHES_PER_MM 0.039370
+
 #ifdef __cplusplus
    }
 #endif
--- a/include/allegro5/internal/aintern_system.h
+++ b/include/allegro5/internal/aintern_system.h
@@ -33,6 +33,7 @@ struct ALLEGRO_SYSTEM_INTERFACE
    void (*shutdown_system)(void);
    int (*get_num_video_adapters)(void);
    bool (*get_monitor_info)(int adapter, ALLEGRO_MONITOR_INFO *info);
+   int (*get_monitor_dpi)(int adapter);
    ALLEGRO_MOUSE_CURSOR *(*create_mouse_cursor)(ALLEGRO_BITMAP *bmp, int x_focus, int y_focus);
    void (*destroy_mouse_cursor)(ALLEGRO_MOUSE_CURSOR *cursor);
    bool (*get_cursor_position)(int *ret_x, int *ret_y);
--- a/src/macosx/system.m
+++ b/src/macosx/system.m
@@ -365,17 +365,23 @@ static int osx_get_num_video_adapters(void)
  */
 static bool osx_get_monitor_info(int adapter, ALLEGRO_MONITOR_INFO* info)
 {
-/*
    int count = osx_get_num_video_adapters();
    if (adapter < count) {
       NSScreen *screen = [[NSScreen screens] objectAtIndex: adapter];
       NSRect rc = [screen frame];
+      rc = [screen convertRectToBacking: rc];
       info->x1 = (int) rc.origin.x;
       info->x2 = (int) (rc.origin.x + rc.size.width);
       info->y1 = (int) rc.origin.y;
       info->y2 = (int) (rc.origin.y + rc.size.height);
+      ALLEGRO_INFO("Display %d has coordinates (%d, %d) - (%d, %d)\n",
+                   adapter, info->x1, info->y1, info->x2, info->y2);
+      return true;
    }
-*/
+   else {
+      return false;
+   }
+/*
    CGDisplayCount count;
    // Assume no more than 16 monitors connected
    static const int max_displays = 16;
@@ -394,6 +400,7 @@ static bool osx_get_monitor_info(int adapter, ALLEGRO_MONITOR_INFO* info)
    else {
       return false;
    }
+*/
 }
 
 /* osx_get_monitor_dpi:
--- a/src/macosx/system.m
+++ b/src/macosx/system.m
@@ -396,6 +396,37 @@ static bool osx_get_monitor_info(int adapter, ALLEGRO_MONITOR_INFO* info)
    }
 }
 
+/* osx_get_monitor_dpi:
+ * Return the dots per inch value of one monitor
+ */
+static int osx_get_monitor_dpi(int adapter)
+{
+   static const int max_displays = 16;
+   CGDirectDisplayID displays[max_displays];
+   CGDisplayCount count;
+   ALLEGRO_MONITOR_INFO info;
+   CGError err;
+   CGSize size;
+   int dpi_hori;
+   int dpi_vert;
+
+   if (!osx_get_monitor_info(adapter, &info)) {
+      return 0;
+   }
+
+   err = CGGetActiveDisplayList(max_displays, displays, &count);
+   if (err != kCGErrorSuccess || adapter < 0 || adapter > (int) count) {
+      return 0;
+   }
+
+   size = CGDisplayScreenSize(displays[adapter]);
+
+   dpi_hori = (info.x2 - info.x1) / (_AL_INCHES_PER_MM * size.width);
+   dpi_vert = (info.y2 - info.y1) / (_AL_INCHES_PER_MM * size.height);
+
+   return sqrt(dpi_hori * dpi_vert);
+}
+
 /* osx_inhibit_screensaver:
  * Stops the screen dimming/screen saver activation if inhibit is true
  * otherwise re-enable normal behaviour. The last call takes force (i.e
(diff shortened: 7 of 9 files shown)


include/allegro5/monitor.h
```diff
@@ -16,6 +16,7 @@ typedef struct ALLEGRO_MONITOR_INFO
    int y1;
    int x2;
    int y2;
+   int dpi;
```
elias-pschernig: maybe this should be float?

src/x/xfullscreen.c
```diff
@@ -922,6 +922,13 @@ bool _al_xglx_get_monitor_info(ALLEGRO_SYSTEM_XGLX *s, int adapter, ALLEGRO_MONI
       info->y1 = 0;
       info->x2 = DisplayWidth(s->x11display, DefaultScreen(s->x11display));
       info->y2 = DisplayHeight(s->x11display, DefaultScreen(s->x11display));
+      int x2_mm = DisplayWidthMM(s->x11display, DefaultScreen(s->x11display));
+      int y2_mm = DisplayHeightMM(s->x11display, DefaultScreen(s->x11display));
+#define INCHES_PER_MM 0.039370
+      int dpi_hori = (info->x2 - info->x1) / (INCHES_PER_MM * x2_mm);
+      int dpi_vert = (info->y2 - info->y1) / (INCHES_PER_MM * y2_mm);
+#undef INCHES_PER_MM
+      info->dpi = sqrt(dpi_hori * dpi_vert);
```
elias-pschernig: maybe we should report both horizontal and vertical dpi like X11?

fatcerberus:
I’m happy to see this PR, I need this functionality too as I dynamically scale my display window based on desktop resolution but then it ends up too large on, e.g. 4K displays.

src/monitor.c
```diff
@@ -44,7 +44,7 @@ bool al_get_monitor_info(int adapter, ALLEGRO_MONITOR_INFO *info)
       }
    }
 
-   info->x1 = info->y1 = info->x2 = info->y2 = INT_MAX;
+   info->x1 = info->y1 = info->x2 = info->y2 = info->dpi = INT_MAX;
```
elias-pschernig: Can we set info->dpi to 0 before the vt->get_monitor_info call? That way platforms who do not report DPI would report it as 0 instead of leaving it completely undefined.

include/allegro5/monitor.h
neuro-sys: Although I think it makes more sense to represent the number of physical dots as an integer, I just checked xrandr source code for reference, and it uses `double` there. So, I could as well change this.

src/x/xfullscreen.c
neuro-sys: I was thinking just returning display dimensions, and let the users calculate DPI themselves. But then it's also more convenient to have it ready.

X11 doesn't return DPI per se, but display dimensions in millimeters.

fatcerberus: How would the OS know how big the monitor is?

src/monitor.c
neuro-sys: Yes, it makes more sense to me too. Seeing this line here made me think a negative value means an invalid or failed query by convention.

src/x/xfullscreen.c
neuro-sys: That's a good question! I believe there's some driver magic going on under the hood to guess or find out the display dimensions. I remember Xorg used to let you configure them manually, so maybe they are not always so reliable after all.

neuro-sys: I believe it'd help to read this: https://www.x.org/wiki/Development/Documentation/HowVideoCardsWork/

neuro-sys: Asking around, it turns out there is,

https://en.wikipedia.org/wiki/Extended_Display_Identification_Data

And in Linux kernel there's DRM and EDID with some drivers implementing it. I'd say for external displays it may not be reliable or in most cases possible at all, but laptops, embedded displays should be fine.

src/monitor.c
neuro-sys: Oh, I just understood what you meant, missed that it returns earlier, and it will be garbage. Right, fixing it now, thanks.

neuro-sys:
I have added MacOS one too.

neuro-sys:
I added Windows one too here: https://github.com/neuro-sys/allegro5/pull/1/files

However as said in the description it couples the windows build with at least Windows 8.1 as it's using [GetDpiForMonitor](https://msdn.microsoft.com/en-us/library/windows/desktop/dn280510(v=vs.85).aspx) available only on:

> Windows 8.1 [desktop apps only]

So I'm excluding that from this branch for now.

docs/src/refman/monitor.txt
```diff
@@ -23,6 +24,7 @@ typedef struct ALLEGRO_MONITOR_INFO
    int y1;
    int x2;
    int y2;
+   int dpi;
```
SiegeLord: We can't actually add stuff to this struct, for backwards compatibility reasons (e.g. imagine a game compiled for 5.2.4 then running on 5.2.5 with this argument added; the library will write to this byte offset while the binary only allocated enough for the 4 fields).

I think the best option here is to just make a new function.

src/x/xrandr.c
```diff
@@ -655,6 +657,12 @@ static bool xrandr_get_monitor_info(ALLEGRO_SYSTEM_XGLX *s, int adapter, ALLEGRO
    mi->y1 = crtc->y;
    mi->x2 = crtc->x + crtc->width;
    mi->y2 = crtc->y + crtc->height;
+#define INCHES_PER_MM 0.039370
```
SiegeLord: Could you just add this (as _AL_INCHES_PER_MM) to allegro/internal/aintern.h so we don't need to re-define it everywhere?

neuro-sys:
@SiegeLord Thanks a lot for the input! I did the changes you requested which makes a whole lot sense now that I've got a little better understanding of the codebase. Just one thing that seemed more appropriate is to put the API in monitor subsystem instead of display. The reason for that is DPI seems to be relevant in the context of a physical monitor. Another reason is, similar to `al_get_monitor_info`, it seems more appropriate to query the DPI per video adapter.

In that case, one would first need to query `al_get_num_video_adapters`, and per video adapter, query the DPI of a monitor attached to the given adapter.

neuro-sys:
I have another unfinished branch in my local that tries to use display API (instead of monitor.h) and infer adapter from the display instance, so that the library user wouldn't have to pick the adapter themselves. Although, I am a little busy lately, if it works out well I could replace this branch with that one.

SiegeLord:
No, I like this one more, it's a little more composable (e.g. if we added al_get_display_adapter, then we'd be able to mimic what you're writing now). I've almost finished testing this change, so it should be merged soon.

SiegeLord:
Alright, I tested this on the big 3 systems. Everything seems to work okay, except in one respect. On OSX, it is possible to force the retina display mode on a regular display. The effect of this is that we pretend the display is physically larger than it is, so everything gets scaled up. Now the DPI returned by this method actually goes down, as if it's computed from points, and not pixels.

Notably, this works correctly on Win10, where I can also scale up everything and this yields a proportional increase in DPI you'd expect.

neuro-sys:
Thanks for testing it out. I don't have a MacBook with me at the moment, but I'll check it out as soon as I get it. What I am curious is the value of DPI on Mac before and after the so scalled retina mode is enabled. From the readings I have had on the web, it sounds similar to Windows' DPI unawareness, which simply bitmap-stretches every pixel by 2 or 4. Just like we enable DPI awareness on Windows (so that no bitmap-stretching occurs), I will look into if something similar is available on Macs when retina mode is displayed, and how other DPI aware apps handle it with and without retina mode is enabled.

neuro-sys:
I got my macbook, and now with a test application I can see what you mean.

However, it actually works correctly. The issue is in the concept of scaling in MacOS and Windows differs inversely.

### Windows

In Windows, when you set scaling, the resolution does not change, and Windows applies a scaling factor to the DPI which is then reported back to the application. The effect of this is that the native resolution does not change, and the DPI is increased or decreased depending on the user's scaling setting, and DPI aware applications scale themselves accordingly (non DPI aware applications are scaled by Windows via bitmap stretching which we disable).

### OSX
Whereas in OSX, when you set scaling, unlike in Windows, the resolution is changed, and the screen dimensions are the same. This results in DPI to change as expected based on the new reported screen resolution. Note that since the "scaling" is done through changing the display resolution in OSX, effective DPI value is calculated inversely as opposed to Windows (in Windows DPI is reported to us anyways). One other background detail about OSX scaling is that the resolutions (or "display modes" as it's called) can be virtual, in which case bitmap stretching is applied to simulate a native resolution, as far as I have understood. Also native cocoa widgets and fonts are still rendered at the physical resolution regardless of the virtual resolution so that no blurring occurs, but this too is based on empirical observation since I couldn't find clear documentation on this.

#### Property `NSHighResolutionCapable`
Setting `NSHighResolutionCapable` to true or false, I have observed no difference on a bitmap application `ex_bitmap`. My impression is that this property is enabled by default and cannot be disabled anymore, since based on the documentation it is used for disabling "Open in Low Resolution" property of an application, which simply activates bitmap stretching (as in non-DPI aware on Windows) which we don't want anyways. So applications are always DPI aware (i.e. HighResolutionCapable).

All in all, It's a conceptual confusion around the definitions of resolution and scaling with regards to Windows and MacOS.

I would say leaving it as it is would have the correct behaviour from a technical standpoint, but might confuse MacOS users when they change their OS scaling (which simply changes the resolution, whether physically or virtually), however the app would scale correctly based on their selected resolution. Alas, OSX uses the phrase "Looks like _widthxheight_", and `CGDisplayPixelsHigh/CGDisplayPixelsWide` and `CGDisplayBounds` both returns the said _widthxheight_. So it does not "look like" but it "is" the said resolution (regardless of the cocoa tricks to render crispy fonts and widgets using native resolutions).

I compared the behaviour of both Windows and MacOS by changing the resolution in both systems (and not changing scaling since MacOS doesn't have the same concept) in an allegro DPI scaled test application, and the application scales correctly in both systems.

Any suggestions are still welcome.

SiegeLord:
Thanks for looking into it. Let me first explain what Allegro does today with respect to hi-DPI displays on OSX. Since Allegro has a pixel-level API, I tried to make it that it keeps working that way on OSX as well. I.e. if you request a 800x600 display, then you'll get an 800x600 pixel display. The way this is accomplished is that we grab the scaling factor (e.g. see here: https://github.com/liballeg/allegro5/blob/master/src/macosx/osxgl.m#L1023-L1030). So, I wonder if we can do the same technique here.

In terms of `NSHighResolutionCapable`, I wonder how you're setting it for ex_bitmap. In my experience `NSHighResolutionCapable` only works in bundles, while apps outside the bundle (like Allegro examples) are always treated as if `NSHighResolutionCapable` was set to true.

SiegeLord:
Ok,. I tried to implement what I had in mind in my branch here: https://github.com/SiegeLord/allegro5/tree/osx_dpi_fix. Could you try this out on your macbook and see if it works properly? The intention is for the DPI to return the 'true' pixels-per-inch on real retina displays, and 'simulated' pixels-per-inch for scaled displays.

neuro-sys:
@SiegeLord Yes, I tested it on a _MacBook Pro (Retina, 13-inch, Early 2015)_, and scaling everything with DPI as a factor, the allegro application remains the same size in all different display modes (Larger Text and More Space, etc). Although, the newly returned DPI values seemed rather large, but could be due to the Retina display. I wonder if the same DPI scaling logic in an application would cause things to scale in different amounts depending on the platform (i.e. on OSX versus say, Linux).

Here is a screenshot comparing two different display modes:

osx_dpi

Here is what I have added to the `ex_bitmap.c` example:

```c
    int dpi = al_get_monitor_dpi(0);

    zoom = dpi / 96.0f;
```

and elsewhere for window dimensions

```c
    int w = 320 * zoom;
    int h = 240 * zoom;

    display = al_create_display(w, h);
```

Assuming a default of `96` of DPI for development (such as my other non HiDPI desktop screen).

neuro-sys:
> In terms of NSHighResolutionCapable, I wonder how you're setting it for ex_bitmap. In my experience NSHighResolutionCapable only works in bundles, while apps outside the bundle (like Allegro examples) are always treated as if NSHighResolutionCapable was set to true.

I have tested it by bundling `ex_bitmap` in a bundle by following the instructions [here](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html#//apple_ref/doc/uid/10000123i-CH101-SW19).

I made sure that bundle's `Info.plist` is taken into account by changing, say, window title. Here is the structure:

```
$ tree ex_bitmap.app/
ex_bitmap.app/
└── Contents
    ├── Info.plist
    ├── MacOS
    │   └── ex_bitmap
    └── Resources
        └── data
            └── mysha.pcx
```

Where `Info.plist` is:

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>CFBundleDevelopmentRegion</key>
	<string>English</string>
	<key>CFBundleExecutable</key>
	<string>ex_bitmap</string>
	<key>CFBundleGetInfoString</key>
	<string>5.1, Copyright 2010 Allegro Developers</string>
	<key>CFBundleIconFile</key>
	<string></string>
	<key>CFBundleIdentifier</key>
	<string>org.liballeg.ex_bitmap</string>
	<key>CFBundleInfoDictionaryVersion</key>
	<string>6.0</string>
	<key>CFBundleLongVersionString</key>
	<string>Allegro Ex Bitmap</string>
	<key>CFBundleName</key>
	<string></string>
	<key>CFBundlePackageType</key>
	<string>APPL</string>
	<key>CFBundleShortVersionString</key>
	<string>5.1</string>
	<key>CFBundleSignature</key>
	<string>????</string>
	<key>CFBundleVersion</key>
	<string></string>
	<key>CSResourcesFileMapped</key>
	<true/>
	<key>NSHumanReadableCopyright</key>
	<string>Copyright 2010 Allegro Developers</string>
        <key>NSHighResolutionCapable</key>
        <true/>
</dict>
</plist>
```

neuro-sys:
> Could you try removing the multiplication here and trying again? SiegeLord@349392e#diff-fa95f3e76664b8255cc4c98127a91f9eR424

I removed the scaling factor and printed the value to stdout, and took a screenshot for comparison. It prints 2.0f for both display modes.

osx_dpi2

> DPI values to be similar across platforms for the same hardware

With the tests I've done on all platforms, it seemed "reasonable" so far, now that after removing the scaling factor from the function you pointed. To get it exactly right will require some more thinking on my part for the time being.

SiegeLord:
Ok, that's good enough for now. Thanks a lot for your contribution!

PR #915, opened 2018-06-28
Fixed compilation for older C

arves100:
This commit fixes compilation of Allegro5 library for compilers that dosen't support C99 for Windows builds without any dependency.
Tested with Visual Sutdio 2008.

### DIFF
--- a/src/opengl/ogl_fbo.c
+++ b/src/opengl/ogl_fbo.c
@@ -174,7 +174,7 @@ static void attach_depth_buffer(ALLEGRO_FBO_INFO *info)
       ALLEGRO_DISPLAY *display = _al_get_bitmap_display(info->owner);
       int w = al_get_bitmap_width(info->owner);
       int h = al_get_bitmap_height(info->owner);
-	  int samples = 0;
+      int samples = 0;
 	  bool extension_supported;
 	  GLint e;
 
--- a/src/opengl/ogl_fbo.c
+++ b/src/opengl/ogl_fbo.c
@@ -175,8 +175,8 @@ static void attach_depth_buffer(ALLEGRO_FBO_INFO *info)
       int w = al_get_bitmap_width(info->owner);
       int h = al_get_bitmap_height(info->owner);
       int samples = 0;
-	  bool extension_supported;
-	  GLint e;
+      bool extension_supported;
+      GLint e;
 
       if (bits == 24) gldepth = GL_DEPTH_COMPONENT24;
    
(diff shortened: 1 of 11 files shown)


src/opengl/ogl_fbo.c
```diff
@@ -174,15 +174,18 @@ static void attach_depth_buffer(ALLEGRO_FBO_INFO *info)
       ALLEGRO_DISPLAY *display = _al_get_bitmap_display(info->owner);
       int w = al_get_bitmap_width(info->owner);
       int h = al_get_bitmap_height(info->owner);
+	  int samples = 0;
```
dos1: Looks like there's some whitespace inconsistency.

src/win/wgl_disp.c
```diff
@@ -967,7 +968,9 @@ static bool create_display_internals(ALLEGRO_DISPLAY_WGL *wgl_disp)
    if ((disp->flags & ALLEGRO_OPENGL_3_0) || major != 0) {
       if (major == 0)
          major = 3;
-      bool fc = (disp->flags & ALLEGRO_OPENGL_FORWARD_COMPATIBLE) != 0;
+	  if ((disp->flags & ALLEGRO_OPENGL_FORWARD_COMPATIBLE) != 0)
```
allefant: that != 0 there is a bit unnecessary if we are introducing a new if() statement here...

src/opengl/ogl_fbo.c
arves100: Fixed in the next commit. Thanks.

src/win/wgl_disp.c
arves100: You're right, fixed in the next commit. Thanks.

arves100:
By now, allegro, color, dialog, font, image, memfile, primitives and ttf should compile.
Image without libwebp support.

elias-pschernig:
Just out of curiosity, why are using such an ancient compiler? In a way it would be nice to move forward and remove support for outdated compilers from Allegro over time (for example require stdint support) - so I'm a bit conflicted about all the changes now to support something so old.

fatcerberus:
I agree with Elias, VS2008 is 10 years old at this point.  I thought our documentation already stated we required C99 support anyway?

dos1:
I think it may be because older Visual Studio use stdlib that's included in Windows, while newer ones need redistributables. However, I'm also conflicted about such stuff, as it definitely doesn't make the code more legible. It may be nice to have such branch floating around somewhere for such strange use cases, but I'm not sure if it should be merged in after all.

arves100:
I haven't checked the documentation if it requires C99.
If so, I could keep this changes into my account in case someone needs it.
Unfortunaly it dosen't help the code begin more redible.
I still use vs2008 because it's the fastest in compilation and IDE usage (my pentium cpu don't like newer vs).
While there are alternatives like CodeBlocks, I don't like none of the open ide, while Vscode it's slow and it's missing several features for C/C++ development.
Regarding stdint, Allegro already have astdint that defined the missing types (uint16_t,...), so I guess there's no problem.

arves100:
I think this PR would have to be rebased eventually, I remember that I was able to get it on vs2008 (so that was w2k+)
I'll close this PR as it doesn't really make sense to keep it open (I forgot to have kept it open)

PR #916, opened 2018-06-29
Added macro definitions for compilers that dosen't have stdint

arves100:
Compiler: Visual Studio 2008
I received an error about UINT16_MAX not defined, this compiler dosen't support stdint.h.
This commit should add the all the macros used by this custom types.
I haven't tested it with other compilers than VS2008, I think it could, at least, warn about some macro redefinitions.

### DIFF
--- a/include/allegro5/platform/astdint.h
+++ b/include/allegro5/platform/astdint.h
@@ -32,12 +32,24 @@
    #ifndef ALLEGRO_GUESS_INTTYPES_OK
       #warning Guessing the definitions of fixed-width integer types.
    #endif
+   #include <limits.h>
+   
    #define int8_t       signed char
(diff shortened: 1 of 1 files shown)


SiegeLord:
Thanks!

dos1:
I believe it should be guarded with some `#ifdef`s.

arves100:
I belive that compilers that does not have stdint support should not have
that headers, at least with vs2008. However that should be tested with
other compilers.

Il ven 3 ago 2018, 00:45 Sebastian Krzyszkowiak <notifications@github.com>
ha scritto:

dos1:
On second thought, yeah, it should be fine as it is, sorry :)

PR #917, opened 2018-06-29
Added support of HarfBuzz in FreeType

arves100:
I've added HarfBuzz compilation support for FreeType, since the current CMakeLists dosen't support it, it pops up some linking error.
HarfBuzz MUST be compiled with HB_HAVE_FREETYPE otherwise the linking won't succed.


elias-pschernig:
I somehow think cmake's Freetype detection should include all the libraries required to link against freetype. I can see you only did what was already done for other similar dependencies so if we can't come up with a better solution that looks reasonable - but I have this feeling there should be a better way.

fatcerberus:
I recall there were attempts in the past to improve FreeType dependency detection; what we have now was deemed the best we could do.  /cc @SiegeLord

SiegeLord:
Ok, merging this as is. I'll try to get HarfBuzz included into the Windows binaries for the next release. Thanks!

PR #918, opened 2018-06-29
Removed stdint.h in allegro/image/bmp.c

arves100:
My compiler (Visual studio 2008) dosen't contain stdint.h, so i remove the inclusion.
Didn't allegro already include astdint.h, making the inclusion pointless?


SiegeLord:
Thanks!

PR #929, opened 2018-09-10
SDL: fix segfaults in mutex handling

dos1:
This is fixed by just removing the code as SDL mutexes are already recursive.

The removed code segfaulted when locking the mutex from another thread than the one that created it (SDL_TLSGet returned NULL, because nothing set it for that thread before).


SiegeLord:
Thanks, didn't test this but your reasoning makes sense.

PR #931, opened 2018-10-03
Fix memory leak in al_install_joystick under Linux

dos1:


SiegeLord:
Thanks!

@fatcerberus Do you have anything in particular we could improve? Or should we just look at tinydir for inspiration?

fatcerberus:
Ideally I would prefer not to have to call al_destroy_fs_entry() for each entry - but I don't think this is possible to change without breaking downstream code, sadly. 😢   Often I want to skip entries using `continue` (when filtering by wildcard, e.g.) and it's easy to create leaks that way by forgetting to free the fs_entry structure.  In fact that's the exact cause of the leak fixed here.

A smaller complaint I have is the need to create an ALLEGRO_FS_ENTRY for the directory being traversed seems like an unnecessary extra step--would be great just to be able to pass in a directory name directly to `al_open_directory`

fatcerberus:
I note additionally that dynamically allocating a structure for each file or directory listed may also be a performance concern when traversing large directories--malloc can be surprisingly expensive.

SiegeLord:
Could just add a new function instead of `al_read_directory` that doesn't allocate a new entry every time. I think it's possible.

PR #932, opened 2018-10-06
fix python3 string return bug

elias-pschernig:
It seems the python wrapper never allowed using functions returning a string under Python3. This patch will fix calling any functions returning a string from Python3 and will have them return a normal Python string.


SiegeLord:
Friends don't let friends use python 3 :snake:

PR #933, opened 2018-10-06
fix and clarify shader docs

elias-pschernig:
The user attributes are named al_user_attr_# and not user_attr_# as
claimed in the documentation. Also clarify how you can use those
attributes.


dos1:
Oh, good one. I remember figuring it out by myself, wasn't particularly easy :)

PR #938, opened 2018-10-12
Allow modules to loop properly when the user hasn't specified loop po…

NewCreature:
…ints. I think this might have been broken in commit 3919ac2c2377dd06afbed0cd2df1bf2498bd9ddc. Previously, if you loaded a module audio stream and played it back with ALLEGRO_PLAYMODE_LOOP and hadn't specified a loop yourself, the module would loop properly to the loop points set within the module itself. I think this is the behavior that anyone using modular music would expect within their applications. Also, since the looping is now handled by the audio add-on instead of individual codec handlers, I have removed the looping code and fixed the rewind code to rewind to the user-specified loop point instead of the beginning.

### DIFF
--- a/addons/acodec/modaudio.c
+++ b/addons/acodec/modaudio.c
@@ -163,7 +163,7 @@ static size_t modaudio_stream_update(ALLEGRO_AUDIO_STREAM *stream, void *data,
 
    /* the mod files are stereo and 16-bit */
    const int sample_size = 4;
-   size_t written;
+   size_t written = 0;
    size_t i;
 
    DUMB_IT_SIGRENDERER *it_sig = lib.duh_get_it_sigrenderer(df->sig);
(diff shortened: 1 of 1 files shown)


fatcerberus:
It hasn’t been my observation that module playback is broken—the loop info in the module file itself was still honored last I checked.

NewCreature:
I just noticed it recently while updating a project of mine. I think it might possibly only manifest if you use DUMB 2.0.x. That's what I'm using. I haven't looked into it thoroughly. I found that when the module reaches the loop point, it doesn't fill the buffer all the way, causing the streaming code to rewind the stream, thus ignoring the module's own loop data.

SiegeLord:
@NewCreature You don't happen to have a module that has internal looping? I used to have one when I was testing this last (pre-DUMB 2.0), but I can't find it now.

NewCreature:
[neon.xm.zip](https://github.com/liballeg/allegro5/files/2475980/neon.xm.zip)
This is the song I first noticed it on. After noticing it here, I tested some other songs and they all fail to loop properly. In this particular song, when it reaches the end it is supposed to loop back to order 1, but instead goes back to the beginning (order 0). All the other looping songs I tested gave the same results.

SiegeLord:
Ok, I tested this and indeed it's a DUMB 2.0-only issue (didn't investigate why). I checked that with this patch DUMB 0.93 and 2.0 work the same way and properly loop, so it all looks good to me.

Thanks a lot!

PR #939, opened 2018-10-13
acodec/wav: fix swapping for big endian

dougmencken:
Current code just moves one byte left without swapping http://coliru.stacked-crooked.com/a/87be408dc4c14d6b because it doesn’t increment by two inside `while (p < end) { uint8_t *const q = p + 1; const uint8_t tmp = *p; *p = *q; *q = tmp; p = q; }`, it swaps 0th byte with 1st then instead of swapping 2nd and 3rd it swaps 1st and 2nd

This is how does fixed code work http://coliru.stacked-crooked.com/a/196d759192422012

Plus, aqueue driver needs to be fixed to play samples on big endian without multiplying gain by 0.089(9) https://developer.apple.com/documentation/coreaudio/klinearpcmformatflagisbigendian

### DIFF
--- a/addons/acodec/wav.c
+++ b/addons/acodec/wav.c
@@ -175,16 +175,13 @@ static size_t wav_read(WAVFILE *wavfile, void *data, size_t samples)
     */
 #ifdef ALLEGRO_BIG_ENDIAN
    if (wavfile->bits == 16) {
-      uint8_t *p = data;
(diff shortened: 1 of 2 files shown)


SiegeLord:
Don't really have a way to test the OSX changes, but everything seems reasonable. Thanks!

PR #943, opened 2018-10-16
add al_transpose_transform

elias-pschernig:
I just needed this functionality and was thinking it almost would be useful to have in Allegro itself - even if it is a bit hard to explain...

### DIFF
--- a/docs/src/refman/transformations.txt
+++ b/docs/src/refman/transformations.txt
@@ -223,6 +223,20 @@ with 2D transformations.
 
 See also: [al_check_inverse]
 
+## API: al_transpose_transform
+
+Transposes the matrix of the given transform. This can be used for
+inversing a rotation transform. For example:
+
+~~~~c
+al_build_camera_transform(camera, 0, 0, 0, x, y, z, xu, yu, zu)
+al_copy_transform(inverse, camera)
+al_transpose_transform(camera)
+// Now "inverse" will be a transformation rotating in the opposite
+// direction from "camera". Note that this only works if the camera
+// position is 0/0/0 as in the example.
+~~~~
+
 ## API: al_check_inverse
 
 Checks if the transformation has an inverse using the supplied tolerance.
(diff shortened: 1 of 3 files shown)


fatcerberus:
Interestingly `glUniformMatrix*fv()` takes a `transpose` parameter, so this is maybe not as out-of-place as it seems.

elias-pschernig:
Something like:

temp = trans->m[i][j]; trans->m[i][j] = trans->m[j][i]; trans->m[j][i] = temp;

But that still would do 16 assignments to temp and 16 from temp, just instead of allocating 16 floats on the stack would just re-use a single one. I really can't tell if that's a in performance-wise or would just not matter at all. And it makes for very ugly code :P

fatcerberus:
Yeah, you're probably right - you'd just end up doing a bunch of individual swaps and probably end up thrashing the CPU cache in the process... making a temp copy of 16 floats upfront is hardly an issue.

SiegeLord:
Looks good to me.

allefant:
I don't really know what transposing a general matrix would do (with translation and scaling and so on...), but *if* a matrix is just a rotation matrix it just so happens that this trick works.

For example here's a matrix which rotates around the z axis by angle a:

cos a  | -sin a  | 0  | 0
--- | --- | --- | ---
__sin a__  | __cos a__ | 0 | 0
0 | 0 | 1 | 0
0 | 0 | 0 | 1

Now transpose the entire matrix:

cos a  | sin a  | 0  | 0
--- | --- | --- | ---
__-sin a__  | __cos a__ | 0 | 0
0 | 0 | 1 | 0
0 | 0 | 0 | 1

And realize how cos(-a) = cos(a) and sin(-a) = -sin(a) and therefore this is identical to:

cos -a  | -sin -a  | 0  | 0
--- | --- | --- | ---
__sin -a__  | __cos -a__ | 0 | 0
0 | 0 | 1 | 0
0 | 0 | 0 | 1

Which is exactly the rotation to undo the original rotation. And it also works for rotations around arbitrary axes.

SiegeLord:
On a deeper level, this works out because rotation matrices are [orthogonal](https://en.wikipedia.org/wiki/Orthogonal_matrix), which means that their transposes are their inverses. Other orthogonal matrices like that are reflections. Translation and scaling are right out.

fatcerberus:
Ah, that makes sense, thanks.  On an interesting side note (interesting to me at least), technically reflections are rotations too, you just have to rotate *through* (**not** about) the next higher dimensional axis.  A 2D object rotated 180 degrees through Z-space obtains its mirror image across its axis of rotation, same deal in 3D if you rotate through W-space.  You could turn a left shoe into a right shoe if you could only turn it around in the fourth dimension!

Issue #944, opened 2018-10-16
ALSA non-deterministic crash: alsa_update_mmap: Assertion `frames' failed.


tehsausage:
The `frames` variable is possibly set to 0 by the `snd_pcm_mmap_begin()` call here:

https://github.com/liballeg/allegro5/blob/acce1d20b195fb12f4ade77595a7fcad0d863cf8/addons/audio/alsa.c#L341

`alsa_voice_is_ready()` is called before to try confirm that the stream is ready but it doesn't seem to be 100% reliable.

This triggers an assertion failure equivalent to `frames != 0` a few lines down.

Example trace of a crash:

```c
static void *alsa_update_mmap(ALLEGRO_THREAD *self, void *arg)
{

// ...

while (!al_get_thread_should_stop(self)) {

// ...

ret = alsa_voice_is_ready(alsa_voice); // ret > 0, so loop continues

// ...

frames = alsa_voice->frag_len; // frames=940
ret = snd_pcm_mmap_begin(alsa_voice->pcm_handle, &areas, &offset, &frames); //frames=940

// ...

ASSERT(frames); // OK

// ...

} // loop repeats

// ...

ret = alsa_voice_is_ready(alsa_voice); // ret > 0, so loop continues

// ...

frames = alsa_voice->frag_len; // frames=940
ret = snd_pcm_mmap_begin(alsa_voice->pcm_handle, &areas, &offset, &frames); //frames=0

// ...

ASSERT(frames); // Assertion fails
```

I believe the triggering condition is that the loop executes faster than ALSA can replenish a frame, and the failure of `alsa_voice_is_ready` could be down to an unusual ALSA setup (the device used passes through multiple plugins).

Reproducing asound.conf, probably not minimal. Reproduction rate is about 5% on the startup of any of the 3 demo games:

```
pcm.noeq_raw {
	type dmix
	ipc_key 1024
	ipc_perm 0666
	slave {
		pcm "hw:PCH"
		rate 48000
		channels 2
		periods 4
		period_size 1024
		buffer_size 4096
	}
}

pcm.aout {
	type plug
	slave.pcm {
		type equal
		library "/usr/lib/ladspa/sc4m_1916.so"
		module "sc4m"
		slave.pcm {
			type plug
			slave.pcm "noeq_raw"
		}
	}
}

pcm.loopout {
	type dmix
	ipc_key 1025
	ipc_perm 0666
	slave {
		pcm "hw:Loopback,0,0"
		rate 48000
[...]

PR #945, opened 2018-10-20
ALSA/PulseAudio crash and memory leak fixes

tehsausage:
Fixes a large memory leak in al_install_audio() if PulseAudio is compiled in but not available, and a smaller memory leak in ALSA after uninstalling audio.

Also comes with a small fix for what appears to be an obvious use-after-free / double-free bug for the "pa_mainloop_iterate failed" case in PulseAudio initialization.

I'm not sure if the ALSA fix (issue #944) is the best it could be -- it just waits and re-tries as the existing under-run handler does -- but it should be a pretty rare case anyway.

### DIFF
--- a/addons/audio/alsa.c
+++ b/addons/audio/alsa.c
@@ -144,6 +144,7 @@ static void alsa_close(void)
    
    alsa_device = NULL;
    
+   snd_output_close(snd_output);
    snd_config_update_free_global();
 }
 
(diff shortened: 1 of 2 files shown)


SiegeLord:
Wow, that's a lot of nastiness right there. Thanks a lot for your debugging and fixes!

PR #952, opened 2018-10-28
SDL: hidpi support

dos1:
This is actually untested, as only after doing that I found out that HIGHDPI support in SDL2 is implemented only for macOS, iOS and Emscripten :D Nevertheless, it's there and should make it match the hidpi behavior of native Allegro on macOS.


dos1:
Started to try implementing it in SDL and I think this patch may actually need something more to it, so maybe don't merge it yet :D

PR #953, opened 2018-10-28
SDL: implement touch subsystem

dos1:
SDL2 only returns absolute positions of touches on the input device. This means we have to fake them in order to make them related to the display, which is what Allegro returns in its API. This is likely to break in lots of cases, but should work well with non-rotated fullscreen or Wayland windows.

### DIFF
--- a/src/sdl/sdl_touch.c
+++ b/src/sdl/sdl_touch.c
@@ -1,0 +1,254 @@
+#include "allegro5/allegro.h"
+#include "allegro5/internal/aintern_system.h"
+#include "allegro5/platform/allegro_internal_sdl.h"
+
+ALLEGRO_DEBUG_CHANNEL("SDL")
+
+typedef struct ALLEGRO_TOUCH_INPUT_SDL
+{
+   ALLEGRO_TOUCH_INPUT touch_input;
+   ALLEGRO_TOUCH_INPUT_STATE state;
+   ALLEGRO_DISPLAY *display;
+   int touches;
+} ALLEGRO_TOUCH_INPUT_SDL;
+
+static ALLEGRO_TOUCH_INPUT_DRIVER *vt;
+static ALLEGRO_TOUCH_INPUT_SDL *touch_input;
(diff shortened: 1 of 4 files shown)


src/sdl/sdl_touch.c
```diff
@@ -0,0 +1,252 @@
...
+      void **v = (void **)_al_vector_ref(&s->displays, 0);
+      d = *v;
+   }
+   if (!d) {
+      return;
+   }
+
+   int touchId = e->tfinger.fingerId;
+
+   touch_input->state.touches[touchId].x = e->tfinger.x;
```
SiegeLord: Shouldn't you multiply-in the display dimensions here for the al_get_touch_state to work correctly?

src/sdl/sdl_touch.c
```diff
@@ -0,0 +1,252 @@
...
+      touch_input->state.touches[touchId].id = -1;
+   } else {
+      return;
+   }
+
+   // SDL2 only returns absolute positions of touches on the input device. This means we have to fake them
+   // in order to make them related to the display, which is what Allegro returns in its API.
+   // This is likely to break in lots of cases, but should work well with non-rotated fullscreen or Wayland windows.
+   generate_touch_input_event(type, e->tfinger.timestamp, touchId,
+                              touch_input->state.touches[touchId].x * al_get_display_width(d),
```
SiegeLord: I guess this means that it's fullscreen only atm (possibly mobile only?)

src/sdl/sdl_touch.c
dos1: Oh, yeah, likely - haven't tested that.

src/sdl/sdl_touch.c
dos1: There is now: https://github.com/SDL-mirror/SDL/commit/e4a845faec1803b3e5c92e2f671f7651cfba4238 :D

dos1:
Updated!

SiegeLord:
Looks good to me! Thanks!

PR #955, opened 2018-11-01
SDL: use the context created by renderer instead of creating a new one

dos1:
This fixes issues with fullscreen toggling, as SDL was making its own
context current again behind our back.

Creating this commit took waaaaay more time than it seems from the diff :D


SiegeLord:
Nice, thanks.

PR #956, opened 2018-11-02
Bunch of SDL stuff

dos1:


src/sdl/sdl_display.c
```diff
@@ -303,6 +309,71 @@ static bool sdl_resize_display(ALLEGRO_DISPLAY *display, int width, int height)
...
+   amask = 0xff000000;
+#endif
+
+   ALLEGRO_LOCKED_REGION *lock = al_lock_bitmap(bitmaps[0], ALLEGRO_PIXEL_FORMAT_ABGR_8888, ALLEGRO_LOCK_READONLY);
+   if (lock) {
+      int i = 0;
+      for (int y = 0; y < h; y++) {
+         for (int x = 0; x < w; x++) {
+            ALLEGRO_COLOR c = al_get_pixel(bitmaps[0], x, y);
+            al_unmap_rgba(c, data+i, data+i+1, data+i+2, data+i+3);
```
SiegeLord: This is hardly speed critical, but you could in principle memcpy here rather than do a get/unmap.

dos1: Correct me if I'm wrong, but I think this would have to take possibility of negative pitch into account, which Allegro can return but SDL doesn't seem to handle in SDL_CreateRGBSurfaceFrom.

SiegeLord: Yes, you'd have to copy row-by row, but that's still significantly faster. Again, this is just an FYI, it doesn't matter here I think.

SiegeLord:
Everything looks fine, thanks!

PR #967, opened 2018-12-07
opengl: fix compilation with Mesa 18.2.5 and later

dos1:
Mesa headers have been updated and changed some defines that Allegro
is hackily relying on.

https://gitlab.freedesktop.org/mesa/mesa/commit/f7d42ee7d319256608ad60778f6787c140badada


rvega:
Confirmed this patch fixes compilation issues with latest mesa in Arch Linux.

SiegeLord:
Thanks!

PR #970, opened 2018-12-12
[4.4] Fix building with Mesa 18.2.5 and later

gusnan:
Based on the patch on Allegro5,
https://github.com/liballeg/allegro5/commit/a40d30e21802ecf5c9382cf34af9b01bd3781e47


SiegeLord:
Thanks!

PR #971, opened 2018-12-12
[4.4] Remove f* variants of fixed math functions

gusnan:
Removing functions that conflicts with later glibc implementations,
people should use fix* variants instead.
(Simply use fixadd instead of fadd).

### DIFF
--- a/include/allegro/alcompat.h
+++ b/include/allegro/alcompat.h
@@ -40,24 +40,6 @@
       #endif
    #endif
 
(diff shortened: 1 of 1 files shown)


gusnan:
For clarity, without this patch, this:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=915196

happens.

SiegeLord:
Works for me. Thanks!

PR #972, opened 2018-12-13
4.4 release fix

rofl0r:
continuing from https://github.com/liballeg/allegro5/pull/962

this PR fixes and modernizes the script to create release tarballs

here is the allegro-4.4.3.tar.xz tarball generated using it: https://0x0.st/sDCp.xz

it seems to be correct afaict. i admit that i didnt check the generated HTML documentation in a browser though.

a minor issue i found is that the extracted sources contain this:
```diff
+++ allegro-4.4.3/include/allegro/base.h        2018-12-13 15:04:02.000000000 +0000
-#define ALLEGRO_WIP_VERSION      2
-#define ALLEGRO_VERSION_STR      "4.4.2"
+#define ALLEGRO_WIP_VERSION      3
+#define ALLEGRO_VERSION_STR      "4.4.3 (SVN)"
```
this would need to be fixed pre release (removal of ` (SVN)` from the string.

the script compiles and utilizes a program to create dat files, which in turn require an installed liballegro4.
therefore i think it's a good idea to use a dedicated VM/container to build the release, so it's not getting mistakenly mixed up with an eventually installed allegro5.


rofl0r:
@SiegeLord any comments ?

SiegeLord:
Apologies, I haven't had the time to look at it. One thing that'll help me with this is a little note why you did what you did: it seems you had to move a bunch of lines around? What about the existing script didn't work for you (when I tried it, it seemed to produce approximately correct output, and your script is also approximately correct in some different way?). I'll do my best to get to this sometime next week.

rofl0r:
Well basically i just read the script to figure out what it's supposed to do, and where potential issues could happen, then rearranged the code-blocks in a way those issues couldn't happen anymore and added some sanity checks while removing the blocks that are basically just error-prone cruft.
I figured that would be way more productive than trying to guess/debug what goes wrong while running the original script. and indeed, it just took me like 5 iterations to get it right.

SiegeLord:
Alright, I got around to testing this. I think the output looks pretty good, save for the `.dist` directory ending up in the archive for me.

There's also something broken with the HTML documentation generation, but I don't think it's related to your changes.

SiegeLord:
Yeah, please. I probably can figure it out eventually, but it'll probably be faster if you do it.

rofl0r:
fixed ^

rofl0r:
many thanks! please keep us posted when you got around to make a release.

SiegeLord:
4.4.3 is now released, https://www.allegro.cc/forums/thread/617731/1041144.

PR #977, opened 2018-12-23
Fix lost precision warning

pedro-w:
This wasn't causing errors but has been bugging me for a while.
````
In file included from /usr/local/include/allegro5/allegro.h:45:
In file included from /usr/local/include/allegro5/fmaths.h:42:
/usr/local/include/allegro5/inline/fmaths.inl:134:25: warning: implicit conversion loses integer precision: 'int64_t' (aka 'long long') to 'int' [-Wshorten-64-to-32]
         int res = lres >> 16;
             ~~~   ~~~~~^~~~~
````
Also, it's not guaranteed that `int` and `al_fixed` (a.k.a `int32_t`) are the same type.

### DIFF
--- a/include/allegro5/inline/fmaths.inl
+++ b/include/allegro5/inline/fmaths.inl
@@ -131,8 +131,7 @@ AL_INLINE(al_fixed, al_fixsub, (al_fixed x, al_fixed y),
          return 0x80000000;
       }
       else {
-         int res = lres >> 16;
-         return res;
+         return (al_fixed) (lres >> 16);
       }
    })
 #endif	    /* al_fixmul() C implementations */

### CODE REVIEW

File: include/allegro5/inline/fmaths.inl [hunk 1]
```diff
@@ -131,8 +131,7 @@ AL_INLINE(al_fixed, al_fixsub, (al_fixed x, al_fixed y),
          return 0x80000000;
       }
       else {
-         int res = lres >> 16;
-         return res;
+         return (al_fixed) (lres >> 16);
```
fatcerberus: Should this be instead `(al_fixed)lres >> 16`?  I know you only did this to fix a warning but the difference is your version does the bitshift in the `int` domain and then casts the result, versus converting `lres` to whatever type al_fixed is and *then* doing the shift.  I'm not sure it matters here but something worth taking into consideration.

pedro-w: `al_fixed` is a typedef for `int32_t`. Here, `lres` is an `int64_t`. To multiply two `al_fixed` you need to multiply them 'as if' they were integers and then divide by 65536, equivalently shift right by 16.
So if you multiply, then cast to 32 bit, then right shift 16, the top 16 bits of your new `al_fixed` will be all 0 or all 1, i.e useless. You need to shift the full 64 bits, *then* cast to 32 bits.

pedro-w: Maybe we could add some tests for fixed numbers (though they are not used much nowadays as I understand it)

fatcerberus:
I actually wasn’t aware there were any platforms where `int` was 64-bit; I had thought it was 32-but everywhere.  Apparently I was mistaken...

fatcerberus:
Yeah, I understand the purpose of the change.  I got confused because I noticed the `int res` bit in the diff and assumed `int` was 64-bit in your case and that’s what triggered the warning.  I should have viewed the whole file to realize `lres` was explicitly declared as int64.  Once I saw that, the warning made perfect sense.

SiegeLord:
Looks reasonable. Thanks!

PR #980, opened 2018-12-28
cmake: fix HarfBuzz detection script and clean up

### LINKED ISSUE #979: cmake: finding HarfBuzz is broken
```
CMake Error at cmake/FindHarfBuzz.cmake:61 (include):
  include could not find load file:

    /deps/allegro5-dos/cmake/SelectLibraryConfigurations.cmake
Call Stack (most recent call first):
  addons/CMakeLists.txt:97 (find_package)


-- Could NOT find HarfBuzz (missing:  HARFBUZZ_LIBRARY)
```

https://github.com/liballeg/allegro5/blob/a40d30e21802ecf5c9382cf34af9b01bd3781e47/cmake/FindHarfBuzz.cmake#L61

There's no `cmake/SelectLibraryConfigurations.cmake` file in the repository.

dos1:
For starters,
  include(${CMAKE_CURRENT_LIST_DIR}/SelectLibraryConfigurations.cmake)
is wrong and should be changed to
  include(SelectLibraryConfigurations)

However, judging from both the docs and the actual output, this function
shouldn't be used at all if earlier detection doesn't set _RELEASE and
_DEBUG variants of its variables. Otherwise, a path already found gets
replaced with "HARFBUZZ_LIBRARY-NOTFOUND".

Fixes #979

### DIFF
--- a/cmake/FindHarfBuzz.cmake
+++ b/cmake/FindHarfBuzz.cmake
@@ -34,10 +34,6 @@
 set(HARFBUZZ_FIND_ARGS
   HINTS
     ENV HARFBUZZ_DIR
-  PATHS
-    ENV GTKMM_BASEPATH
-    [HKEY_CURRENT_USER\\SOFTWARE\\gtkmm\\2.4;Path]
-    [HKEY_LOCAL_MACHINE\\SOFTWARE\\gtkmm\\2.4;Path]
 )
 
 find_path(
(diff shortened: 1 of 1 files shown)


SiegeLord:
I guess I was leaving properly testing this for when I made the windows package >.>... and I'll still do so, hah.

Thanks a lot, this looks a lot more reasonable. Not sure what `select_library_configurations` is all about, but if it works without it for you, then I'm satisfied.

PR #982, opened 2019-01-10
SDL: implement al_(show|hide)_mouse_cursor

dos1:


SiegeLord:
Wth is that API >_<. Almost as bad as `ALLEGRO_ONE`... Thanks!

fatcerberus:
> `ALLEGRO_ONE`

The thing about these is, for whatever reason they seem to be a necessary evil.  When I designed the BlendOp API for miniSphere I ultimately ended up with a blending API that looks very similar to Allegro/OpenGL/etc.

It would at least be nice if there were a second CONST_COLOR slot but even modern OpenGL appears to only support one...

PR #983, opened 2019-01-13
bitmap_io: try to identify bitmap format for unknown extensions

dos1:
Allows to load identifiable bitmaps stored with custom extension.

### DIFF
--- a/src/bitmap_io.c
+++ b/src/bitmap_io.c
@@ -193,13 +193,11 @@ ALLEGRO_BITMAP *al_load_bitmap_flags(const char *filename, int flags)
    Handler *h;
    ALLEGRO_BITMAP *ret;
 
-   ext = strrchr(filename, '.');
+   ext = al_identify_bitmap(filename);
    if (!ext) {
-      ext = al_identify_bitmap(filename);
+      ext = strrchr(filename, '.');
       if (!ext) {
-         ALLEGRO_ERROR("Bitmap %s has no extension and filetype "
-            "identification failed - not even trying to load it.\n",
-            filename);
+         ALLEGRO_ERROR("Could not identify bitmap %s!", filename);
          return NULL;
       }
    }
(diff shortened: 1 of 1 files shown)


fatcerberus:
Does this allow you to pass NULL for `ext` to get the auto-detection?  If not, perhaps it should.

dos1:
Hmm, thinking about it now, I'd actually prefer to always try to identify and only rely on the extension when the bitmap can't be identified.

What do you think?

fatcerberus:
I’d be behind that change 100%.  Currently in miniSphere I have to do the filetype detection myself, to support files saved with the wrong extension.  It would be great to just let Allegro do the detection for me.

fatcerberus:
As you can see, I'm basically reimplementing what Allegro could in theory already do for me, but unfortunately when loading from memfiles (`al_load_bitmap_f`) I apparently have to supply the extension (or at least, passing NULL for `file_ext` wasn't legal last time I checked the docs):
https://github.com/fatcerberus/minisphere/blob/master/src/minisphere/image.c#L165-L176

Changing to use autodetection first and the supplied extension only as a fallback like you suggest would allow me to get rid of the above code altogether. :smiley:

SiegeLord:
I'm all for identifying first. It's not clear to me when you'd ever want to fall back to the extension, honestly, unless somehow the identification code is incomplete. I suppose it's better be safe than sorry, however.

fatcerberus:
> It's not clear to me when you'd ever want to fall back to the extension, honestly

Yeah, I agree, if a file type is supported then the autodetection logic should be able to identify it.  The fallback would basically just be if the code was buggy or if a format was extended in a way that the detection failed (but the file is otherwise compatible).  Also the `ident` parameter can’t be removed without breaking source compatibility, so the caller may as well provide it. :wink:

In theory there might also be an unseekable file so you don’t have the luxury of reading ahead, but that’s just a hypothetical and unlikely to happen in any real-world scenario.

dos1:
Some native loaders currently define extensions without providing identification function, but otherwise yeah, identification alone should be enough.

dos1:
(ideally the native loader should probably be invoked as a fallback for every extension, since it might support something that's not defined in Allegro, but that's a different story)

SiegeLord:
Excellent. Thanks!

PR #984, opened 2019-01-13
xwin.c: increase X_MAX_EVENTS to 50

rofl0r:
this should fix mouse lag on Linux with hi-res devices.
( see https://github.com/aseprite/aseprite/issues/283 )


SiegeLord:
Thanks!

PR #986, opened 2019-01-20
cmake: fix static linking ttf addon dependencies when there's circular dependency between freetype and harfbuzz

dos1:
Otherwise it ends up with unresolved symbols.

### DIFF
--- a/addons/CMakeLists.txt
+++ b/addons/CMakeLists.txt
@@ -112,7 +112,9 @@ if(SUPPORT_FONT AND WANT_TTF)
             endif()
 			
             if (FREETYPE_HARFBUZZ AND HARFBUZZ_FOUND)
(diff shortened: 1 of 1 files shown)


SiegeLord:
Weird... but I guess that's harmless even if sometimes unnecessary. Thanks!

fatcerberus:
Yeah, the linker only uses the libraries on the command line to resolve *currently unresolved* symbols; if there’s a circular dependency then linking in the second library will add new unresolved symbols (i.e. the linker won’t look at things earlier in the command line), and you have to name the earlier library again.

In my experience MSVC is the only compiler that doesn’t exhibit this behavior.

dos1:
I think linker parameters can be also surrounded with `-(` and `-)` in order to specify a group of libraries to deal with that, but I have no idea how to get CMake to do that. Specifying the library second time seems to be a commonly practiced solution too, so I just went with that :)

PR #988, opened 2019-01-21
Restore FBO/GLContext after we're done with the current context.

### LINKED ISSUE #981: bitmap locking is broken
Look at this simple example:
```c
#include <allegro5/allegro5.h>

int main(void) {
    al_init();
    al_create_display(100, 100);
    ALLEGRO_BITMAP *test = al_create_bitmap(100, 100);
    for (int i = 0; i < 17; i++) {
        al_set_target_bitmap(NULL);
        ALLEGRO_LOCKED_REGION *lock = al_lock_bitmap(test,
            al_get_bitmap_format(test), ALLEGRO_LOCK_READONLY);
        al_unlock_bitmap(test);
    }
}
```
If you look at the generated ALLEGRO_LOG, you will see 

opengl   E           ogl_lock.c:143  _al_ogl_lock_region_new          [   0.28232] glPixelStorei(GL_PACK_ALIGNMENT, 4) failed (GL_STACK_OVERFLOW).

Inside of al_lock_bitmap, Allegro has the following call sequence:

// in _al_ogl_lock_region_new
glPushClientAttrib(GL_CLIENT_PIXEL_STORE_BIT);
ogl_lock_region_nonbb_readwrite(...);
glPopClientAttrib();

However ogl_lock_region_nobb_readwrite will destroy the current OpenGL context if the target bitmap is NULL (I do not know why): https://github.com/liballeg/allegro5/blob/master/src/opengl/ogl_lock.c#L273

As a consequence, glPopClientAttrib() is called after the context was removed from the thread, getting ignored. This means glPushClientAttrib gets pushed to the stack but never removed - for me the stack only has 16 entries so the 17th time OpenGL responds with a GL_STACK_OVERFLOW error.

The obvious fix would be to simply comment out the above line, it makes everything work for me - and I do not see why the locking code should ever remove the OpenGL context. Another more conservative fix would be to still remove the context, but only *after* glPopClientAttrib() has been called.

Waiting first if someone else has some more insight into this part of the OpenGL code.

SiegeLord:
Previously, we would restore it before we were done with the current
context, causing issues as described in #981.

Fixes #981.


SiegeLord:
@elias-pschernig What do you think of this fix?

allefant:
> @elias-pschernig What do you think of this fix?

Looks good to me, and it does fix my test case. My game also still works with the PR applied (but I changed it since to also set the target bitmap before locking I think). I did not test the OpenGL ES version (and my original bug was not possible there because the ES version does not restore the client attributes, unlike the desktop GL version).

PR #994, opened 2019-02-03
SDL: set window title via al_get_new_window_title()

dos1:


SiegeLord:
Is that `title` field used for anything? It doesn't appear to be written to (or now read) anywhere...

dos1:
Right, updated :)

PR #997, opened 2019-02-04
Fix: Race Condition & Add Static Runtime Option

ZoriaRPG:
Changelog:
Fix: The keyboard handler has a race condition that can cause Allegro programs to stop accepting any keyboard input (usually manifests itself after the program has been running a considerable length of time). Replaced the old attempt at protecting the critical section with a mutex.
( evouga, 12th August, 2017 )

Feat: Added option to compile Allegro with a static runtime library.  (ZC needs this.)
( evouga, 19th August, 2017 )

FixL Added a missing #include directive.
( evouga, 10th August, 2017 )

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,6 +111,10 @@ option(WANT_EMBED
 set(FRAMEWORK_INSTALL_PREFIX "/Library/Frameworks" CACHE STRING
     "Directory in which to install Mac OS X frameworks")
 
+if(MSVC)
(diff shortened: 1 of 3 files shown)


SiegeLord:
Thanks, this looks good! The only thing I've changed after merging this is to toggle the STATIC_RUNTIME option to off by default, you can always turn it back on as necessary on the command line.

Merged as https://github.com/liballeg/allegro5/commit/51d822d959e2d44ea8dc28e5e34fc95f9d209a6a.

PR #999, opened 2019-02-05
SDL: hidpi support

dos1:
This time done right :D

I'm not yet sure if the case of pixel ratio changing at runtime will be handled properly (it should be, but you never know until you see it :P ), but considering that at this time SDL supports hidpi only on macOS, iOS and Emscripten, it can be tackled in a later PR.

### DIFF
--- a/src/sdl/sdl_display.c
+++ b/src/sdl/sdl_display.c
@@ -16,6 +16,15 @@ int _al_win_determine_adapter(void);
 
 static ALLEGRO_DISPLAY_INTERFACE *vt;
 
+float _al_sdl_get_display_pixel_ratio(ALLEGRO_DISPLAY *display)
+{
+   ALLEGRO_DISPLAY_SDL *sdl = (void *)display;
+   int window_width, drawable_width, h;
+   SDL_GetWindowSize(sdl->window, &window_width, &h);
+   SDL_GetRendererOutputSize(sdl->renderer, &drawable_width, &h);
+   return drawable_width / (float)window_width;
+}
+
 ALLEGRO_DISPLAY *_al_sdl_find_display(uint32_t window_id) {
    unsigned int i;
    ALLEGRO_SYSTEM *s = al_get_system_driver();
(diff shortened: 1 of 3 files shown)


src/sdl/sdl_mouse.c
```diff
@@ -14,6 +14,15 @@ typedef struct ALLEGRO_MOUSE_SDL
 static ALLEGRO_MOUSE_DRIVER *vt;
 static ALLEGRO_MOUSE_SDL *mouse;
 
+static float sdl_get_display_pixel_ratio(ALLEGRO_DISPLAY *display)
```
SiegeLord: Maybe make this internal (i.e. put a declaration in include/allegro5/platform/allegro_internal_sdl.h) ? You use this in a few more place than just in this file.

src/sdl/sdl_display.c
```diff
@@ -144,14 +144,27 @@ static ALLEGRO_DISPLAY *sdl_create_display_locked(int w, int h)
...
    sdl->renderer = SDL_CreateRenderer(sdl->window, -1, flags);
    sdl->context = SDL_GL_GetCurrentContext();
+
+   SDL_GetRendererOutputSize(sdl->renderer, &d->w, &d->h);
+
+   // there's no way to query pixel ratio before creating the window, so we
+   // have to compensate afterwards
+   if (d->flags & ALLEGRO_RESIZABLE &&
+        !(d->flags & ALLEGRO_FULLSCREEN || d->flags & ALLEGRO_FULLSCREEN_WINDOW)) {
+      int window_width, window_height;
```
SiegeLord: Would you mind adding an ALLEGRO_DEBUG saying that it's resizing, and also ALLEGRO_WARN if it wasn't resizable but the ration is not 1.?

dos1: If it wasn't resizable, then the hidpi support wouldn't be enabled at all (most likely resulting in automatic resize by system compositor with Allegro/SDL not even being aware of that): https://github.com/liballeg/allegro5/pull/999/files#diff-49d564f1704de963274032237e706cc9R105-R106

If we want to allow hidpi support on non-resizable Allegro windows, I guess a new API would have to be created on Allegro side to handle that.

dos1:
Updated!

SiegeLord:
Thanks!

PR #1000, opened 2019-02-05
Experimental emscripten support with SDL platform

dos1:
Just a bare-bones stuff to get the thing running with the SDL backend. Roughly based on some old @elias-pschernig's patch, but greatly simplified thanks to the GLES rework.

I compile it with:
`-DWANT_ALLOW_SSE=OFF -DVORBIS_COMPILES=yes -DTTF_COMPILES=yes -DWANT_EXAMPLES=no -DWANT_DEMO=no -DWANT_TESTS=no -DWANT_MONOLITH=yes -DWANT_OPENAL=no -DSHARED=no -DEMSCRIPTEN_GENERATE_BITCODE_STATIC_LIBRARIES=yes`

then manually link (so I don't have to do it on the game side) with:
`emcc liballegro_monolith-static.bc libSDL2main.a libSDL2.a libvorbisfile.a libvorbis.a libopusfile.a libopus.a libphysfs.a libtheoradec.a libFLAC.a libogg.a libdumb.a libfreetype.a libharfbuzz.a libfreetype.a libjpeg.a libwebp.a libpng.a libz.a libbz2.a -o allegro5.so`

and it seems to work :)

(I compiled all dependencies manually, without using emscripten's port system)

Note: There's no support for audio and video streams for now, as those depend on threading.


SiegeLord:
I'm confused a little bit. I don't see anything about disabling threads in this patch, which to me means that SDL implements them somehow on emscripten. Is that not true?

dos1:
Emscripten implements the pthreads API, but it doesn't actually do anything unless experimental pthreads support via webworkers is enabled (which doesn't actually work in current browsers thanks to Meltdown/Spectre mitigations).

dos1:
(also, there's no way to not be confused with emscripten, you just have to accept it ;))

fatcerberus:
I thought web workers themselves weren’t disabled, it was just the SharedArrayBuffer that allows workers to communicate directly (instead of the weird postMessage API).

SiegeLord:
I noticed (for the first time) the heartbeat system, I wonder if we can implement threads in general with it. This'll will probably need an interface system that SDL/other backends can fallback on emscriopten. Then, we won't have to do the timer callback thing manually like you did.

dos1:
@fatcerberus Yup, I was referring to emscripten's pthreads support, which requires SharedArrayBuffer.
@SiegeLord I don't think you can really generalize it this way (an infinite loop inside a thread is going to break everything; any long running loop is going to give you troubles in the browser too), but for short-lived, repeating callbacks it should be good enough. That's how I think audio and video streams can be eventually tackled.

dos1:
(oh, after rereading that, you probably meant reimplementing existing threads that Allegro uses for its stuff with this new interface... yeah, should work :) )

SiegeLord:
Or `al_create_callback_thread` for the public-facing version of it.

dos1:
Makes sense, but I'd avoid the word "thread" there. Maybe "task"?

GObject seems to call such callbacks idle functions: https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html#g-idle-add

CMakeLists.txt
```diff
@@ -334,6 +334,11 @@ if(ALLEGRO_RASPBERRYPI)
    set(ALLEGRO_EXCLUDE_GLX 1)
 endif(ALLEGRO_RASPBERRYPI)
 
+if(EMSCRIPTEN)
+    set(GL_AUTO_BUILD_TYPE "gles2+")
+    set(ALLEGRO_SDL on)
```
SiegeLord: I'm a little uncomfortable about turning this on silently. Could we force the user to turn on SDL manually and check here that it is set?

dos1: Makes sense :+1:

dos1:
Updated!

SiegeLord:
Thanks!

PR #1001, opened 2019-02-05
SDL: fix compilation with SDL older than 2.0.9

dos1:
SDL_CreateThreadWithStackSize is a very new API addition.


SiegeLord:
...we need to add SDL to Travis :P.

Thanks!

PR #1002, opened 2019-02-07
Add SDL to the travis script.

SiegeLord:


.travis.yml
```diff
@@ -29,7 +30,7 @@ before_script:
 script:
     - |
       if [ `uname` = "Linux" ]; then
-         cmake .. -DCMAKE_BUILD_TYPE=Debug -DWANT_SHADERS_GL=$WANT_SHADERS_GL -DWANT_CURL_EXAMPLE=on
+         cmake .. -DCMAKE_BUILD_TYPE=Debug -DWANT_SHADERS_GL=$WANT_SHADERS_GL -DWANT_SDL=$WANT_SDL -DWANT_CURL_EXAMPLE=on
```
dos1: It's ALLEGRO_SDL, not WANT_SDL :D

SiegeLord: :man_facepalming:

SiegeLord:
Current issues:
- No dialog addon, so can't do the install test on Linux
- Doesn't build on OSX

dos1:
I've looked into SDL on macOS.

* the first issue is that the build system sets ALLEGRO_MACOSX and appends macOS files to the build
* after fixing that, build fails on including GL headers, because the correct ones are behind ALLEGRO_MACOSX ifdef...

SiegeLord:
Ah, going above and beyond I see. Don't worry about it too much, I just wanted to note down what issues I hit so I remembered what was up if I returned to it. Thanks for investigating things regardless!

dos1:
Well, it just made me curious. I have a pretty convenient cross-compilation environment already setup, wouldn't bother otherwise :D

PR #1004, opened 2019-02-10
Fix vsync on OSX 10.14+.

### LINKED ISSUE #978: Vsync is broken on macOS 10.14 Mojave
Apparently Apple broke vsync with OpenGL on Mojave.  See for example:
* https://github.com/glfw/glfw/issues/1337
* https://discourse.libsdl.org/t/sdl2-no-renderer-vsync-on-macos-mojave/25169
* (a zillion other GitHub issues on various repos that I don't feel like rounding up 😛)

The bug was reported against miniSphere by @rhuanjl; upon further iteration we discovered that neither setting `ALLEGRO_VSYNC` on the display, nor manually calling `al_wait_for_vsync()` immediately before flipping has any effect.  Then I took to Google and found out that this is apparently a widespread issue.

In SDL's case, the issue was able to be worked around by compiling with Xcode 9 instead of 10, but that doesn't seem to fix it for us.  So I figured I would report it here in case anyone else has any ideas.

SiegeLord:
This follows the changeset from SDL and GLFW:

https://hg.libsdl.org/SDL/rev/73f3ca85ac0e
https://github.com/glfw/glfw/issues/1337

But with a few modifications based on my observations:

- <s>I think I do a better job at always waiting for a refresh</s>
- I keep the old way around because in my testing the new way (while limiting
  frames) did not prevent screen tearing.

Fixes #978

### DIFF
--- a/src/macosx/osxgl.h
+++ b/src/macosx/osxgl.h
@@ -36,6 +36,11 @@ typedef struct ALLEGRO_DISPLAY_OSX_WIN {
    BOOL in_fullscreen;
    BOOL single_buffer;
    CGDisplayModeRef original_mode;
+   /* For new (10.14+) vsyncing. */
+   CVDisplayLinkRef display_link;
+   ALLEGRO_MUTEX *flip_mutex;
+   ALLEGRO_COND *flip_cond;
+   int num_flips;
 } ALLEGRO_DISPLAY_OSX_WIN;
 
 /* This is our version of ALLEGRO_MOUSE_CURSOR */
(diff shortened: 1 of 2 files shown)


SiegeLord:
@fatcerberus I'm reasonably sure that this is correct, but is this something you could test?

fatcerberus:
I had a MacBook Air for all of about a week before returning it so I can’t personally test this, but I’ll ring in @rhuanjl who maintains the macOS release of miniSphere and reported the original problem.

PR #1006, opened 2019-02-12
SDL: don't create SDL_Renderer

dos1:
As we do all GL drawing by ourselves, SDL_Renderer is completely unnecessary.

It actually even triggers a bug in current SDL master :D


SiegeLord:
Thanks!

PR #1007, opened 2019-02-12
SDL: mouse: ignore simulated touch events

dos1:


SiegeLord:
Interesting... thanks!

PR #1008, opened 2019-02-12
SDL: fix window resize with hidpi

dos1:
To fully work in all cases it will need a fix on SDL side as well, since SDL on Emscripten doesn't report window resize event when only the pixel ratio scale has changed and the logical window size remained the same (which is the case when moving the window to another screen or changing the zoom ratio in browser).


SiegeLord:
Thanks!

PR #1009, opened 2019-02-13
SDL: don't assume that mouse events always come with associated window

dos1:


SiegeLord:
That's... interesting as well. I guess it's the best you can do. Thanks!

dos1:
FYI, I got crashes on mouse up events when you moved the cursor outside of the window while you hold the button.

SiegeLord:
Yeah, I understood the gist of it. My concern was that once you do that, it'll start giving you nonsensical values since the scale is wrong.

dos1:
Ah, that's right. I guess we just have to live with the fact that some details like that will be lost in the translation due to differences in how those APIs are built. We could track state and employ some heuristic, but I'm not sure it's worth it. Can always be added later when it turns out there's a real need for that ¯\_(ツ)_/¯ With this patch I simply wanted to avoid a hard crash.

(similar situation happens with multiple displays - without tracking state, you have no idea which display does such event refer to, so any reported coordinates are non-sense)

dos1:
You know what? The single-display case is actually trivial, so I just implemented it :D 
#1011

PR #1010, opened 2019-02-15
Add al_get_system_id function

dos1:
Some systems, like SDL or Raspberry Pi, can be employed on a platform where a more generic system can also work.
In such case there's no way to differentiate between systems, as the usual way to handle things like that, preprocessor macros, by definition can't work at runtime.

BTW. I'll be more than happy to change those names if you come up with better ideas :D If you prefer strings to enum, that's fine too. I have added `ALLEGRO_SYSTEM_ID_UNKNOWN` as I see some support for custom systems in the code.

### DIFF
--- a/src/sdl/sdl_system.c
+++ b/src/sdl/sdl_system.c
@@ -315,7 +315,7 @@ ALLEGRO_SYSTEM_INTERFACE *_al_sdl_system_driver(void)
       return vt;
 
    vt = al_calloc(1, sizeof *vt);
-   vt->id = AL_ID('S', 'D', 'L', '2');
+   vt->id = ALLEGRO_SYSTEM_ID_SDL;
    vt->initialize = sdl_initialize;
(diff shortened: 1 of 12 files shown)


src/sdl/sdl_system.c
```diff
@@ -315,7 +315,7 @@ ALLEGRO_SYSTEM_INTERFACE *_al_sdl_system_driver(void)
       return vt;
 
    vt = al_calloc(1, sizeof *vt);
-   vt->id = AL_ID('S', 'D', 'L', '2');
```
SiegeLord: I'm kind of partial towards this macro, could we use it for everything (i.e. use it as the values in the enum, I'm fine with the enum existing).

SiegeLord:
I think I like this. If you can finish in a day or two it can make it into 5.2.5!

dos1:
Updated!

SiegeLord:
Thanks, this is great!

PR #1011, opened 2019-02-16
SDL: match window-less mouse events to a display if there's only one

dos1:
When SDL returns an event with no associated window (like a mouse button
release outside of the window), we assume 1 as a scaling factor, which
might be wrong.

When there's only one display, figuring out which display the event
coordinates refer to is trivial, so this implements it. Given that
multidisplay usage is rare, and multidisplay usage with SDL backend
probably non-existent, this should cover vast majority of cases.


SiegeLord:
Nice, thanks!

PR #1013, opened 2019-02-19
fix python wrapper generation after the al_get_system_id patch

allefant:
It would fail on things like ALLEGRO_SYSTEM_ID_XGLX otherwise which is AL_ID('X', 'G', 'L', 'X').


SiegeLord:
Wow, what a catch! Another thing to add to our travis build...

PR #1015, opened 2019-02-21
Fix C4334 warning on VS2017

pedro-w:
This is just a minor change to fix a warning for the windows code only. It comes from casting 32-bit shift to 64 bit.


SiegeLord:
I liked the solution used by #1019 more (it didn't use a macro), but thanks anyway!

pedro-w:
Coincidence! I suppose.

PR #1016, opened 2019-02-21
Use non-deprecated function to get Windows version

pedro-w:
A minor fix to the code to avoid a deprecated function. Helper functions are given in VersionHelpers.h but this header is not available in SDKs before Windows 8. If the header is not present, the CMake configuration supplies an alternative (basically copied equivalent code from VersionHelpers.h)

### DIFF
--- a/src/win/d3d_disp.cpp
+++ b/src/win/d3d_disp.cpp
@@ -39,6 +39,21 @@
 
 #include "d3d.h"
 
+// Based on the code in <VersionHelpers.h> but that header is not
+// available on Windows 7 SDK and earlier.
+static BOOL is_windows_vista_or_greater() {
+   OSVERSIONINFOEX info;
+   ULONGLONG mask = 0;
+   VER_SET_CONDITION(mask, VER_MAJORVERSION, VER_GREATER_EQUAL);
+   VER_SET_CONDITION(mask, VER_MINORVERSION, VER_GREATER_EQUAL);
+   VER_SET_CONDITION(mask, VER_SERVICEPACKMAJOR, VER_GREATER_EQUAL);
+   memset(&info, 0, sizeof(info));
+   info.dwOSVersionInfoSize = sizeof(OSVERSIONINFOEX);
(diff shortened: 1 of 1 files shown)


fatcerberus:
My understanding of this is that *all* facilities to get the OS version are de facto deprecated; you’re just not supposed to do it.  Even the Win8+ functions won’t report any version past 8.0 (even on 8.1 or 10) unless the executable is compiled with the correct GUID in its manifest.

That said, it looks like the check here only needs to ask “Vista or later?” so that should still work fine.

pedro-w:
Agreed. I've fixed a compiler warning by replacing a deprecated function with a non-deprecated one. The advice from Microsoft is to test for features, not OS version. Unfortunately it's not clear to me what purpose these tests were originally implemented for, and I don't have any pre-Vista system to check.
Do we still support XP or earlier anyway?

fatcerberus:
I’m going to say yes, since the official binaries are compiled with the “WinXP” MSVC toolchain.

src/win/d3d_disp.cpp
```diff
@@ -39,6 +39,24 @@
 
 #include "d3d.h"
 
+#ifdef ALLEGRO_HAVE_VERSIONHELPERS_H
+#include <VersionHelpers.h>
+#else
+// Based on the code in <VersionHelpers.h> but that header is not 
+// available on Windows 7 SDK and earlier.
+static BOOL IsWindowsVistaOrGreater() {
```
SiegeLord: It feels like it'd be simpler to just always use this implementation and not worry about VersionHelpers.h. What do you think? If we do that, could you rename it to `is_windows_vista_or_greater` to match our style?

pedro-w: That makes sense, I will do it. The other advantage is that the changes to CMakeLists.txt and alplatf.h.cmake are no longer needed.

SiegeLord:
Alright, this looks good. Thanks!

PR #1017, opened 2019-02-21
Windows - use UNICODE mechanism throughout

### LINKED ISSUE #397: windows: compile with -DUNICODE -D_UNICODE
Originally reported by: wangp

Original Ticket: [alleg/bugs/376](https://sourceforge.net/p/alleg/bugs/376)

The Windows port should use the "unicode" versions of the Win32 API.

pedro-w:
This is quite an extensive change to the Windows code only. It allows Allegro to be compiled in wide-char (i.e. UTF-16) mode by defining UNICODE and _UNICODE in the config, or in Ansi mode if they are not defined. This affects mostly the Joystick and Native Dialog code.
In several places there was an implicit assumption that Ansi strings could be freely interchanged with `char*` in Allegro. This is not the case because Allegro assumes a `char*` to be UTF-8 (or the 7-bit ASCII subset) and Ansi strings can contain characters in the range 0x80-0xFF which are decoded according to the system code page.
This lead to https://www.allegro.cc/forums/thread/617652 and https://www.allegro.cc/forums/thread/617697
There is also code in fshooks which uses  the wide-char functions exclusively and this has not been changed.
Co-incidentally it fixes a bug in the native 'text log' where line breaks did not show up (LF -> CRLF conversion)

Fixes #397

### DIFF
--- a/addons/native_dialog/win_dialog.c
+++ b/addons/native_dialog/win_dialog.c
@@ -245,7 +245,6 @@ bool _al_show_native_file_dialog(ALLEGRO_DISPLAY *display,
          wpath = _twin_ustr_to_tchar(path);
       }
       else {
-         //al_ustr_encode_utf16(path, buf, sizeof(buf));
          /* Extract the directory from the path. */
          initial_dir_path = al_clone_path(fd->fc_initial_path);
          if (initial_dir_path) {
--- a/addons/native_dialog/win_dialog.c
+++ b/addons/native_dialog/win_dialog.c
@@ -14,6 +14,7 @@
 #include "allegro5/internal/aintern_native_dialog.h"
 
 #include "allegro5/platform/aintwin.h"
+#include "allegro5/internal/aintern_wunicode.h"
 
 #include "allegro5/allegro_windows.h"
 
@@ -36,7 +37,7 @@ static int wlog_count = 0;
 static void *wlog_rich_edit_module = 0;
 
 /* Name of the edit control. Depend on system resources. */
-static wchar_t* wlog_edit_control = L"EDIT";
+static TCHAR* wlog_edit_control = TEXT("EDIT");
 
 /* True if output support unicode */
 static bool wlog_unicode = false;
@@ -46,6 +47,31 @@ static ALLEGRO_MUTEX* global_mutex;
 static ALLEGRO_COND* wm_size_cond;
 static bool got_wm_size_event = false;
 
+/* For Unicode support, define some helper functions
+ * which work with either UTF-16 or ANSI code pages
+ * depending on whether UNICODE is defined or not.
+ */ 
+
+/* tcreate_path:
+ * Create path from TCHARs
+ */
+static ALLEGRO_PATH* _tcreate_path(const TCHAR* ts) 
+{
+   char* tmp = _twin_tchar_to_utf8(ts);
+   ALLEGRO_PATH* path = al_create_path(tmp);
+   al_free(tmp);
+   return path;
+}
+/* tcreate_path:
+ * Create directory path from TCHARs
+ */
+static ALLEGRO_PATH* _tcreate_path_for_directory(const TCHAR* ts) 
+{
+   char* tmp = _twin_tchar_to_utf8(ts);
+   ALLEGRO_PATH* path = al_create_path_for_directory(tmp);
+   al_free(tmp);
+   return path;
+}
 
 bool _al_init_native_dialog_addon(void)
 {
@@ -67,27 +93,33 @@ void _al_shutdown_native_dialog_addon(void)
    wm_size_cond = NULL;
 }
 
+
 static bool select_folder(ALLEGRO_DISPLAY_WIN *win_display,
    ALLEGRO_NATIVE_DIALOG *fd)
 {
    BROWSEINFO folderinfo;
    LPCITEMIDLIST pidl;
-   char buf[MAX_PATH] = "";
-   char dbuf[MAX_PATH] = "";
+   /* Selected path */
+   TCHAR buf[MAX_PATH] = TEXT("");
+   /* Display name */
+   TCHAR dbuf[MAX_PATH] = TEXT("");
 
    folderinfo.hwndOwner = win_display ? win_display->window : NULL;
    folderinfo.pidlRoot = NULL;
    folderinfo.pszDisplayName = dbuf;
-   folderinfo.lpszTitle = al_cstr(fd->title);
+   folderinfo.lpszTitle = _twin_ustr_to_tchar(fd->title);
    folderinfo.ulFlags = 0;
    folderinfo.lpfn = NULL;
 
    pidl = SHBrowseForFolder(&folderinfo);
+
+   al_free((void*) folderinfo.lpszTitle);
+
    if (pidl) {
       SHGetPathFromIDList(pidl, buf);
       fd->fc_path_count = 1;
       fd->fc_paths = al_malloc(sizeof(void *));
-      fd->fc_paths[0] = al_create_path(buf);
+      fd->fc_paths[0] = _tcreate_path(buf);
       return true;
    }
    return false;
@@ -143,7 +175,7 @@ static ALLEGRO_USTR *create_filter_string(const ALLEGRO_USTR *patterns)
    return filter;
 }
 
-static int skip_nul_terminated_string(char *s)
+static int skip_nul_terminated_string(TCHAR *s)
 {
    int i = 0;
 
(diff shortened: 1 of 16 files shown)


pedro-w:
If this looks OK, a subsequent change can just drop Ansi and use wide-char functions exclusively.

SiegeLord:
This looks pretty extensive! Could you explain what the advantage of making this support both ANSI and UNICODE rather than just UNICODE? Also, how do you select between ANSI and UNICODE with this change? Which one is the default?

pedro-w:
The dual UNICODE/ANSI concept is spread through a lot of the WIndows system headers. Generally this just means a conditional `#define`, e.g. `CreateWindow` is either `CreateWindowA` or `CreateWindowW` but there are some convenience macros in the DirectX code which are more complicated.
I tried initially to use explicit Unicode functions everywhere but it made things a bit messy and I wasn't 100% sure I had accounted for all functions (usually the compiler warns but sometimes we have an explicit cast and there is no warning)
Overall I didn't feel it was adding much code to do the ANSI too.
As it stands ANSI will be the default and Unicode is an opt-in but I would recommend using CMake to always have Unicode selected by setting a compiler flag (`/DUNICODE` `/D_UNICODE`)

pedro-w:
I haven't time just now but maybe at the weekend I can try unicode-only again and see how it looks.

fatcerberus:
I believe recent MSVC versions (since 2015 at least, possibly earlier than that) will define `UNICODE` by default in new projects.  There’s really no reason to support the ANSI functions anymore unless we need to support stuff like Windows 98/Me...

pedro-w:
I don't think the compiler itself defines UNICODE ([source](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2017)) but I think Visual Studio does set that option in the MSBuild file for new projects. But if that MSBuild file comes from CMake I suspect it doesn't.
My experience was that it seemed more straightforward to use the mechanisms that the Windows headers provide (`TCHAR` and so on) than to use the wide char functions explicitly.
I agree there's no *need* for ANSI  - I can't believe there are any Allegro users building for WinME out there? - and I do recommend we build Allegro in Unicode mode going forward (currently it's mixed, some parts are ANSI, some not). In any case the mode used to build Allegro itself shouldn't affect the Allegro API. (that statement probably needs checking!)
It may be the most pragmatic option is to set UNICODE via the CMake file, remove the ANSI functions and have something like

```
#ifndef UNICODE
#error Allegro must be compiled in UNICODE mode
#endif
```
in the windows internal code somewhere.

fatcerberus:
> I don't think the compiler itself defines UNICODE (source) but I think Visual Studio does set that option in the MSBuild file for new projects.

Yeah, that's what I meant by "by default in new projects" - I meant "MSVC" as in the IDE, not the compiler :smile:

FWIW Microsoft themselves seem to consider `TCHAR` to be basically deprecated and recommend just using the Unicode stuff (`WCHAR`, `W` APIs, etc.) nowadays (emphasis mine):
> The TEXT and TCHAR macros are less useful today, *because all applications should use Unicode.* 
 However, you might see them in older code and in some of the MSDN code examples.
https://docs.microsoft.com/en-us/windows/desktop/learnwin32/working-with-strings#tchars

pedro-w:
That's a useful link. 
I am not at a PC with the Windows SDK installed so I can't find the exact bit that was painful in the DirectX. We are using their type macros (say for sake of argument it was `LPCDXTHING` to mean a constant pointer to a `DxThing`) which is standard practice for Windows. But `DxThing` doesn't exist, only `DxThingA` and `DxThingW` so I was having to change `LPCDXTHING` to `const DxThingW*` and `LPLPDXTHING` to `DxThingW**` etc. etc. and it seemed error prone, because in COM it's quite common to cast to/from a void pointer so the compiler couldn't help. Does that make sense?

addons/native_dialog/win_dialog.c
```diff
@@ -207,21 +242,22 @@ bool _al_show_native_file_dialog(ALLEGRO_DISPLAY *display,
       }
 
       if (is_dir) {
-         ofn.lpstrInitialDir = path;
+         wpath = _twin_ustr_to_tchar(path);
       }
       else {
-         strncpy(buf, path, BUFSIZE - 1);
+         //al_ustr_encode_utf16(path, buf, sizeof(buf));
```
SiegeLord: What's up with this?

include/allegro5/platform/aintwin.h
```diff
@@ -283,6 +283,11 @@ void _al_win_touch_input_handle_end(int id, size_t timestamp, float x, float y,
 void _al_win_touch_input_handle_move(int id, size_t timestamp, float x, float y, bool primary, ALLEGRO_DISPLAY_WIN *win_disp);
 void _al_win_touch_input_handle_cancel(int id, size_t timestamp, float x, float y, bool primary, ALLEGRO_DISPLAY_WIN *win_disp);
 
+/* Helpers for getting Windows system errors 
+ * May be called from addons 
+ */
+AL_FUNC(const char*, _al_win_error, (DWORD));
```
SiegeLord: Doesn't look like these are thread safe (on the account of using static buffers). Is that a concern?

src/win/wgl_disp.c
```diff
@@ -135,27 +107,27 @@ static HGLRC init_temp_context(HWND wnd)
    pf = ChoosePixelFormat(dc, &pfd);
    if (!pf) {
       ALLEGRO_ERROR("Unable to chose a temporary pixel format. %s\n",
-                     get_error_desc(GetLastError()));
+		  _al_win_last_error());
```
SiegeLord: These got hard tabs introduced, should be just spaces.

src/win/wjoydxnu.cpp
```diff
@@ -681,20 +686,20 @@ static void fill_joystick_info_using_caps_and_names(ALLEGRO_JOYSTICK_DIRECTX *jo
...
-      info->stick[2].name = ADD_STRING("[D-pad]");
+      info->stick[0].name = ADD_STRING(0, "[L-stick] or D-pad");
+      info->stick[1].name = ADD_STRING(0, "[R-stick]");
+      info->stick[2].name = ADD_STRING(0, "[D-pad]");
    }
 
 #undef N_AXIS
 #undef N_STICK
-#undef OR
+#undef ADD_STRING
```
SiegeLord: No need, it's undeff-d a line below.

pedro-w:
Thanks for reviewing, @SiegeLord, I will address these issues but it might be a couple of days.

include/allegro5/platform/aintwin.h
pedro-w: They weren't thread-safe before, if I understood correctly. However if we're changing the code we might as well make it robust - I don't suppose these functions need to be fast.
I don't know off hand, but I bet you do -  what does the rest of Allegro do where it needs to avoid a static buffer?
* Allocate memory
* User supplies a buffer
* Something else?

src/win/wjoydxnu.cpp
pedro-w: PEBKAC!!

include/allegro5/platform/aintwin.h
SiegeLord: Ah, okay, if they weren't threadsafe before then we're not making things worse. In fact, this is better now because this logic is centralized, and it is possible to fix it without too much hunting.

I think I'd prefer allocating memory. This is not something you need to do though for this pull request to go in.

addons/native_dialog/win_dialog.c
pedro-w: Deleted the commented-out line

include/allegro5/platform/aintwin.h
pedro-w: OK.

src/win/wjoydxnu.cpp
pedro-w: I deleted the extra line.

pedro-w:
> it should be fine to mix and match a UNICODE Allegro + ANSI user app, right?

Yes, it should all go via UTF-8.

SiegeLord:
Merged as https://github.com/liballeg/allegro5/commit/597bcf834ada4663e51350ed31a978e487b8968f with some minor changes.  Notably, I added a definition of UNICODE/_UNICODE in CMakeLists.txt.

Thanks a lot, @pedro-w , this was amazing work. The code was of very high quality and it was quite tricky in a few places.  I appreciate it a lot, this was great.

In the future we should probably spend some time to delete the ANSI code, but that should be easy given how you structured your code.

PR #1018, opened 2019-02-21
Fix for OpenGL version selection on Windows

EdgarReynaldo:
This patch does a few things : 

1) Respect ALLEGRO_OPENGL_MAJOR and MINOR_VERSION on Windows
2) Fix for extension reporting based on inaccurate flag
3) Store the opengl version in the display options for retrieval by the user

### DIFF
--- a/src/win/wgl_disp.c
+++ b/src/win/wgl_disp.c
@@ -1002,6 +1002,13 @@ static bool create_display_internals(ALLEGRO_DISPLAY_WGL *wgl_disp)
       disp->extra_settings.settings[ALLEGRO_COMPATIBLE_DISPLAY] = 0;
    }
 
+   /* Fill in the display settings for opengl major and minor versions...*/
+   int* s = disp->extra_settings.settings;
+   s[ALLEGRO_OPENGL_MAJOR_VERSION] = 
+      (disp->ogl_extras->ogl_info.version >> 24) & 0xFF;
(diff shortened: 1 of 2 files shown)


src/win/wgl_disp.c
```diff
@@ -959,10 +959,8 @@ static bool create_display_internals(ALLEGRO_DISPLAY_WGL *wgl_disp)
       return false;
    }
 
-   major = _al_get_suggested_display_option(disp,
```
allefant: What is the reason for this change? Since a display is passed to the function it feels much safer using that display and not the global flags which (maybe?) could have changed in the meantime.

EdgarReynaldo: _al_get_suggested_display_option is broken and does not refer to the settings requested by the user. Threading was not taken into consideration, as there are other problems with multi threaded display creation.

EdgarReynaldo: al_get_new_X setting is used commonly throughout the display creation code, so it would need to be a bigger change if you wanted that

allefant: I'm not sure _al_get_suggested_display_option is broken - but I think it contains the detected settings, not the to-be-set ones, so just the wrong function to use there. So yeah, good catch! What problems do you see with multi-threaded display creation? Either way, your patch would not affect that so looks good to me.

EdgarReynaldo:
This pull request is superceded by https://github.com/liballeg/allegro5/pull/1021

PR #1019, opened 2019-02-21
Fixed bit shift

EdgarReynaldo:
Fixed broken bit shift to respect values greater than 31


fatcerberus:
Isn’t this the same fix as #1015?

allefant:
(uint64_t) is more consistent with how the rest of the file does it, so this one should probably be applied. It already was fixed in 69730350609fef711c6e719a923198c875a04586 but then someone came in c1fee9734c202532675f6eceb2c18516160bd51c and broke it again <_<

SiegeLord:
Thanks!

Issue #892, opened 2018-03-04
Display options inconsistently set on platforms


SiegeLord:
This affects pretty much every platform, and unfortunately likely needs to be fixed in the platform specific code. The idea is that you should get meaningful values for `al_get_display_option`, even if that option does nothing for that platform.

See https://github.com/liballeg/allegro5/issues/887 for an example issue.

Related this is also documenting what all those display options default to.

fatcerberus:
I would probably fix this by storing the user-provided flag mask in the core as-is, and then allowing the individual platform components to interpret the mask as they see fit.  When the user requests the current flags, just return the cached value from the core.

PR #1022, opened 2019-02-23
SDL: report ALLEGRO_OPENGL_MAJOR/MINOR_VERSION in display options

dos1:
This makes it consistent with #1021


SiegeLord:
I'll take it, but like my comments in #1021 suggest, I think this is just papering over the core issue in #892.

dos1:
Yup, as I said in #1021 - nothing else seems to set it right now :/

PR #1025, opened 2019-02-23
Report OpenGL version in display options on Android, iOS and Raspberry Pi

dos1:
The only ones left now are macosx (I got lost in all its various context initializations) and gp2xwiz, with this lovely piece of code inside:
```
   // FIXME
   // We don't have this extra_settings stuff set up right
   //if (display->extra_settings.settings[ALLEGRO_COMPATIBLE_DISPLAY])
      setup_gl(display);
```
:D


SiegeLord:
Thanks!

PR #1031, opened 2019-03-28
Add a .editorconfig file

dos1:
Lots of editors and IDEs support the EditorConfig format, so adding it to the repository allows automatic format configuration regardless of used editor.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+trim_trailing_whitespace = true
+indent_style = space
+indent_size = 3
+
+[CMakeLists.txt]
+indent_size = 4
+
+[*.cmake]
+indent_size = 4


.editorconfig
```diff
@@ -0,0 +1,15 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+trim_trailing_whitespace = true
```
SiegeLord: Hmm, right now the source code is in an inconsistent state about this... so if someone edits a file with trailing whitespace, they'll generate unrelated changes. I suppose we should just bite the bullet and do a wholesale whitespace fix (there's also hard tabs here and there), but until we do that I'm somewhat uncomfortable about this being the default.

guilt: I agree. I saw this when adding an example, and I guess any file modified/added would start getting fixes, which makes sense.

SiegeLord: Actually, let's just go for it as is, we can alter this if it becomes annoying.

PR #1036, opened 2019-04-06
Enable debug window for DEBUGMODE on Windows (XP and Higher).

guilt:
This patch enables easier debugging. It is easpecially useful for -mwindows / -SUBSYSTEM:windows applications.

We do not need to resort to allegro.log in Windows, so this patch does exactly that.

DebugWindow

### DIFF
--- a/src/debug.c
+++ b/src/debug.c
@@ -180,7 +180,7 @@ static void open_trace_file(void)
          // stderr will be redirected to xcode's debug console though, so
          // it's as good to use as the NSLog stuff.
          trace_info.trace_file = stderr;
-#elif defined(ALLEGRO_WINDOWS)
+#elif defined(ALLEGRO_ANDROID)
          trace_info.trace_file = NULL;
 #else
          trace_info.trace_file = fopen("allegro.log", "w");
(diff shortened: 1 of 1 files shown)


src/debug.c
```diff
@@ -326,9 +328,12 @@ void _al_trace_suffix(const char *msg, ...)
             static_trace_buffer);
       }
       #endif
+      #ifdef ALLEGRO_WINDOWS
+         (void) OutputDebugString((LPCTSTR) static_trace_buffer);
+      #endif
       static_trace_buffer[0] = '\0';
    }
-   else if (trace_info.trace_file) {
+   if (trace_info.trace_file) {
```
SiegeLord: Could you make this be `if (trace_info.trace_file && !_al_user_trace_handler) {`? The idea being that if there is a user trace handler, we don't want to write to the file. The `else` used to take care of it, but now since we need to do the above unconditionally,

guilt: Yeah, it's sort of weird when you install an handler in between and we still have the file open. Then people will be like: where's the allegro.log copy of it - or why is it empty? :)

SiegeLord:
This seems good, thanks!

PR #1037, opened 2019-04-06
MiniMP3 support (Sample mode only for now)

pmprog:
Referencing MiniMP3 (https://github.com/lieff/minimp3), currently only plays MP3s as "samples"

### DIFF
--- a/addons/acodec/mp3.c
+++ b/addons/acodec/mp3.c
@@ -1,0 +1,114 @@
+/*
+ * Allegro5 MP3 reader.
+ * Requires MiniMP3 from https://github.com/lieff/minimp3
+ * author: Mark Watkin (pmprog) 2019
+ */
+
+
+#include "allegro5/allegro.h"
+#include "allegro5/allegro_acodec.h"
+#include "allegro5/allegro_audio.h"
+#include "allegro5/internal/aintern.h"
+#include "allegro5/internal/aintern_audio.h"
+#include "allegro5/internal/aintern_exitfunc.h"
+#include "allegro5/internal/aintern_system.h"
+#include "acodec.h"
+#include "helper.h"
+
+#define MINIMP3_IMPLEMENTATION
+#include <minimp3.h>
+#include <minimp3_ex.h>
+
+ALLEGRO_DEBUG_CHANNEL("acodec")
+
+static bool mp3_libinit = false;
+static mp3dec_t mp3d;
+
+void mp3_initminimp3();
+
+ALLEGRO_SAMPLE *_al_load_mp3(const char *filename)
+{
+   ALLEGRO_FILE *f;
+   ALLEGRO_SAMPLE *spl;
+   ASSERT(filename);
+
+   f = al_fopen(filename, "rb");
+   if (!f)
+      return NULL;
+
+   spl = _al_load_mp3_f(f);
+
+   al_fclose(f);
+
+   return spl;
+}
+
+ALLEGRO_SAMPLE *_al_load_mp3_f(ALLEGRO_FILE *f)
+{
+	mp3_initminimp3(); // Make sure library is initialised
+	
+	mp3dec_file_info_t info;
+	ALLEGRO_SAMPLE *spl = NULL;
+
+	// Read our file size
+	int64_t filesize = al_fsize(f);
+	if (filesize == -1)
+		return NULL;
+
+	// Allocate buffer and read all the file
+	size_t fsize = (size_t)al_fsize(f);
+	uint8_t* mp3data = (uint8_t*)al_malloc(fsize);
+	size_t readbytes = al_fread(f, mp3data, fsize);
+	if (readbytes != fsize)
+	{
+		al_free(mp3data);
+		return NULL;
+	}
+
+	// Decode the file contents, and copy to a new buffer
+	mp3dec_load_buf(&mp3d, mp3data, filesize, &info, NULL, NULL);
+	uint8_t* pcmdata = (uint8_t*)al_malloc(info.samples * sizeof(int16_t));
+	memcpy(pcmdata, info.buffer, info.samples * sizeof(int16_t));
+
+	// Free file copy, and buffer (which is copied to pcmdata
+	al_free(mp3data);
+	al_free(info.buffer);
+
+	// Create sample from info variable
+	spl = al_create_sample(pcmdata, info.samples, info.hz, _al_word_size_to_depth_conf(2), _al_count_to_channel_conf(info.channels), true);
+	
+   return spl;
+}
+
+ALLEGRO_AUDIO_STREAM *_al_load_mp3_audio_stream(const char *filename, size_t buffer_count, unsigned int samples)
+{
+   ALLEGRO_FILE *f;
+   ALLEGRO_AUDIO_STREAM *stream;
+   ASSERT(filename);
+
+   f = al_fopen(filename, "rb");
+   if (!f)
+      return NULL;
+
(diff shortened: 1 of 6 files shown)


addons/acodec/mp3.c
```diff
@@ -0,0 +1,114 @@
...
+   spl = _al_load_mp3_f(f);
+
+   al_fclose(f);
+
+   return spl;
+}
+
+ALLEGRO_SAMPLE *_al_load_mp3_f(ALLEGRO_FILE *f)
+{
+	mp3_initminimp3(); // Make sure library is initialised
```
SiegeLord: Would you mind changing tabs to 3 spaces?

addons/acodec/mp3.c
```diff
@@ -0,0 +1,114 @@
...
+	// Read our file size
+	int64_t filesize = al_fsize(f);
+	if (filesize == -1)
+		return NULL;
+
+	// Allocate buffer and read all the file
+	size_t fsize = (size_t)al_fsize(f);
+	uint8_t* mp3data = (uint8_t*)al_malloc(fsize);
+	size_t readbytes = al_fread(f, mp3data, fsize);
+	if (readbytes != fsize)
```
SiegeLord: The braces should be attached to the conditional:
```
if (...) {

}
```

addons/acodec/mp3.c
```diff
@@ -0,0 +1,114 @@
...
+
+ALLEGRO_AUDIO_STREAM *_al_load_mp3_audio_stream_f(ALLEGRO_FILE* f, size_t buffer_count, unsigned int samples)
+{
+	mp3_initminimp3(); // Make sure library is initialised
+	return NULL;
+}
+
+void mp3_initminimp3()
+{
+	if(!mp3_libinit)
```
SiegeLord: Another brace that needs moving.

addons/acodec/mp3.c
```diff
@@ -0,0 +1,114 @@
...
+	{
+		mp3dec_init(&mp3d);
+		mp3_libinit = true;
+	}
+}
```
SiegeLord: Would you mind adding a newline after this brace?

addons/acodec/mp3.c
```diff
@@ -0,0 +1,114 @@
...
+#define MINIMP3_IMPLEMENTATION
+#include <minimp3.h>
+#include <minimp3_ex.h>
+
+ALLEGRO_DEBUG_CHANNEL("acodec")
+
+static bool mp3_libinit = false;
+static mp3dec_t mp3d;
+
+void mp3_initminimp3();
```
SiegeLord: This should be `mp3_initminimp3(void)`. If you omit the `void` in C it means that the argument list is unspecified (unlike in C++).

addons/acodec/mp3.c
```diff
@@ -0,0 +1,114 @@
...
+{
+	mp3_initminimp3(); // Make sure library is initialised
+	
+	mp3dec_file_info_t info;
+	ALLEGRO_SAMPLE *spl = NULL;
+
+	// Read our file size
+	int64_t filesize = al_fsize(f);
+	if (filesize == -1)
+		return NULL;
```
SiegeLord: Could you add a `ALLEGRO_WARN("Could not determine file size.")`

addons/acodec/mp3.c
```diff
@@ -0,0 +1,114 @@
...
+
+	// Allocate buffer and read all the file
+	size_t fsize = (size_t)al_fsize(f);
+	uint8_t* mp3data = (uint8_t*)al_malloc(fsize);
+	size_t readbytes = al_fread(f, mp3data, fsize);
+	if (readbytes != fsize)
+	{
+		al_free(mp3data);
+		return NULL;
```
SiegeLord: Could you add a `ALLEGRO_WARN("Failed to read file into memory.")`

addons/acodec/mp3.c
```diff
@@ -0,0 +1,114 @@
...
+	mp3dec_file_info_t info;
+	ALLEGRO_SAMPLE *spl = NULL;
+
+	// Read our file size
+	int64_t filesize = al_fsize(f);
+	if (filesize == -1)
+		return NULL;
+
+	// Allocate buffer and read all the file
+	size_t fsize = (size_t)al_fsize(f);
```
SiegeLord: You already have your filesize measured a few lines above.

addons/acodec/mp3.c
```diff
@@ -0,0 +1,114 @@
...
+	uint8_t* mp3data = (uint8_t*)al_malloc(fsize);
+	size_t readbytes = al_fread(f, mp3data, fsize);
+	if (readbytes != fsize)
+	{
+		al_free(mp3data);
+		return NULL;
+	}
+
+	// Decode the file contents, and copy to a new buffer
+	mp3dec_load_buf(&mp3d, mp3data, filesize, &info, NULL, NULL);
```
SiegeLord: What happens when this fails to decode the file? Should there be an error check + return NULL here somewhere?

addons/acodec/mp3.c
```diff
@@ -0,0 +1,114 @@
...
+	uint8_t* pcmdata = (uint8_t*)al_malloc(info.samples * sizeof(int16_t));
+	memcpy(pcmdata, info.buffer, info.samples * sizeof(int16_t));
```
SiegeLord: Why do we need to do a memcpy here. Can we not pass `info.buffer` directly to `al_create_samples`?

addons/acodec/mp3.c
```diff
@@ -0,0 +1,114 @@
...
+}
+
+ALLEGRO_AUDIO_STREAM *_al_load_mp3_audio_stream(const char *filename, size_t buffer_count, unsigned int samples)
+{
+   ALLEGRO_FILE *f;
+   ALLEGRO_AUDIO_STREAM *stream;
+   ASSERT(filename);
+
+   f = al_fopen(filename, "rb");
+   if (!f)
```
SiegeLord: Could you add `ALLEGRO_WARN("Could not open file '%s'.", filename)`

addons/acodec/mp3.c
pmprog: I think I did this for a reason... but for the life of me, I can't remember why now, so I'll change it and try it out

addons/acodec/mp3.c
pmprog: Unfortunately, mp3dec_load_buf does not have a return value, so I don't really know how to tell if it's worked or not

https://github.com/lieff/minimp3/blob/e9df0760e94044caded36a55d70ab4152134adc5/minimp3_ex.h#L46

addons/acodec/mp3.c
fatcerberus: I'm assuming that `info.buffer` would be managed by the `mp3dec` API.  Will Allegro make a copy of the data itself?

pmprog: It's not, because a few lines down you could see that I manually freed info.buffer after copying. Maybe that's what i thought when i originally wrote the code though. Kind of wish I'd made more notes :-)

addons/acodec/mp3.c
SiegeLord: I took a look at what it does, I think you can check whether info.buffer is NULL or not. Could you add code to do that?

SiegeLord:
Merged as https://github.com/liballeg/allegro5/commit/05009f72ff88736af896858b53525e22f6e6d0f3, with a few fixes. Thanks a lot, it works great!

PR #1039, opened 2019-04-06
Add FreeImage support in image addon.

guilt:
I realized that there are many image formats that offer plentiful advantages over the ones we support, hence it may be a good idea to start adding support for loading them into Allegro.

Currently, we load as 8-8-8-8 bitmaps but we should be able to use this library to load higher bits per channel images, such as the ones with HDR, if we support it.

I did not add save image support because for me, loading some images was the initial goal of writing this.

I added an example which is a semi-copy-pasta of ex_bitmap.c which uses ALLEGRO_FILE * routines to test the _bitmap_f loaders.


addons/image/freeimage.c
```diff
@@ -0,0 +1,168 @@
...
+      ALLEGRO_LOCKED_REGION *a_lock = al_lock_bitmap(bitmap,
+         ALLEGRO_PIXEL_FORMAT_ARGB_8888, ALLEGRO_LOCK_WRITEONLY);
+      if (a_lock) {
+         unsigned char *out = (unsigned char *)a_lock->data;
+         int out_inc = a_lock->pitch - (width*4);
+         for(int j=height - 1; j > -1; --j) {
+            for(int i=0; i < width; ++i) {
+               RGBQUAD color;
+               if(FreeImage_GetPixelColor(fib, i, j, &color) == FALSE) {
+                  ALLEGRO_ERROR("Unable to get pixel data at %d,%d\n", i , j);
```
SiegeLord: Could we initialize `color` to something reasonable here so we're not writing garbage?

addons/image/freeimage.c
```diff
@@ -0,0 +1,168 @@
...
+   freeimage_initialized = false;
+}
+
+static ALLEGRO_BITMAP *_al_fi_to_al_bitmap(FIBITMAP *fib) {
+   int width = 0, height = 0;
+   ALLEGRO_BITMAP *bitmap = NULL;
+   width = FreeImage_GetWidth(fib);
+   height = FreeImage_GetHeight(fib);
+   bitmap = al_create_bitmap(width, height);
+   if(bitmap) {
```
SiegeLord: In Allegro we add a space after the `if`, `for` etc. Would you mind doing a search/replace for all such cases in these files?

addons/image/freeimage.c
```diff
@@ -0,0 +1,168 @@
...
+   FREE_IMAGE_FORMAT fif = FIF_UNKNOWN;
+
+   ASSERT(filename);
+   ASSERT(freeimage_initialized == true);
+
+   fif = FreeImage_GetFIFFromFilename(filename);
+   if(fif == FIF_UNKNOWN)
+      fif = FreeImage_GetFileType(filename, 0);
+   if(fif == FIF_UNKNOWN)
+      return bitmap;
```
SiegeLord: `return NULL` would be clearer. Also, could you add a `ALLEGRO_WARN("Could not determine the file type for '%s'", filename)`

addons/image/freeimage.c
```diff
@@ -0,0 +1,168 @@
...
+
+   {
+     FIBITMAP *fibRaw = FreeImage_Load(fif, filename, flags);
+     if(!fibRaw)
+       return bitmap;
```
SiegeLord: Same as above, and could you add a ALLEGRO_WARN("Could not load image '%s'", filename). Does FreeImage provide any sort of error messages? If so, could they be logged here?

addons/image/freeimage.c
```diff
@@ -0,0 +1,168 @@
...
+     fib = FreeImage_ConvertTo32Bits(fibRaw);
+     FreeImage_Unload(fibRaw);
+     if(!fib)
+       return bitmap;
```
SiegeLord: Same as above.

addons/image/freeimage.c
```diff
@@ -0,0 +1,168 @@
...
+   ASSERT(freeimage_initialized == true);
+
+   fio.read_proc  = _fiio_al_read;
+   fio.write_proc = _fiio_al_write;
+   fio.seek_proc  = _fiio_al_fseek;
+   fio.tell_proc  = _fiio_al_ftell;
+
+   fif = FreeImage_GetFileTypeFromHandle(&fio, (fi_handle)f, 0);
+   if(fif == FIF_UNKNOWN)
+      return bitmap;
```
SiegeLord: Same as above: return NULL + warning.

addons/image/freeimage.c
```diff
@@ -0,0 +1,168 @@
...
+
+   {
+     FIBITMAP *fibRaw = FreeImage_LoadFromHandle(fif, &fio, (fi_handle)f, 0);
+     if(!fibRaw)
+       return bitmap;
+     fib = FreeImage_ConvertTo32Bits(fibRaw);
+     FreeImage_Unload(fibRaw);
+     if(!fib)
+       return bitmap;
```
SiegeLord: Same as above.

addons/image/iio.c
```diff
@@ -52,6 +52,25 @@ bool al_init_image_addon(void)
...
+         ".psd", ".cut", ".xpm", ".hdr", ".fax", ".sgi", ".exr", ".raw",
+         ".j2k", ".jp2", ".jpf", ".jpm", ".jpx", ".mj2", ".pfm", ".pict",
+         ".jxr", ".jbg", NULL};
+      int i;
+
+      if (_al_init_fi()) {
+         for (i = 0; extensions[i]; i++) {
+            success |= al_register_bitmap_loader(extensions[i], _al_load_fi_bitmap);
+            success |= al_register_bitmap_loader_f(extensions[i], _al_load_fi_bitmap_f);
+            success |= al_register_bitmap_identifier(extensions[i], _al_load_fi_bitmap_f);
```
SiegeLord: Is this supposed to be `_al_identify_fi`?

cmake/FindFreeImage.cmake
```diff
@@ -0,0 +1,55 @@
...
+# Try to find the FreeImage library and include path.
+# Once done this will define
+#
+# FREEIMAGE_FOUND
+# FREEIMAGE_INCLUDE_PATH
+# FREEIMAGE_LIBRARY
+#
+
+IF (WIN32)
+	FIND_PATH( FREEIMAGE_INCLUDE_PATH FreeImage.h
```
SiegeLord: Could you replace the tabs in this file with 4 spaces?

addons/image/freeimage.c
guilt: Done!

addons/image/iio.c
guilt: Whoa :o I'm surprised ex_bitmap_file worked without it. Fixed, Thank you!

cmake/FindFreeImage.cmake
guilt: Done

`expand -t 4 FindFreeImage.cmake > FindFreeImage.cmake-tmp && mv FindFreeImage.cmake-tmp FindFreeImage.cmake`

addons/image/freeimage.c
guilt: Done!

addons/image/freeimage.c
guilt: Fixed.

addons/image/freeimage.c
guilt: Fixed.

addons/image/freeimage.c
guilt: Fixed. 

For ALLEGRO_FILE, we don't have a filename though.

addons/image/freeimage.c
guilt: We make use of the FreeImage error handler now. This should handle it.

addons/image/freeimage.c
guilt: Fixed.

If there's a FreeImage error, FreeImage should emit a warning from the error handler installed.

SiegeLord:
Nice, this looks great. Let me just try running it locally as a sanity check, but I think it's basically all set.

SiegeLord:
Merged as https://github.com/liballeg/allegro5/commit/0638d7c3bc40da54a3aff251432d72f6415ea8fa, thanks! Just needed to fix one little thing in the `CMakeLists.txt`.

PR #1043, opened 2019-05-01
Fix Unicode/Non-Unicode path appearing on Output Console.

guilt:
When debugging with Windows today, I noticed that sometimes the Output messages are unreadable, at least for me :(

Old

I didn't understand why Windows messes this up, but long story short, I decided to make it output through the Non-UNICODE version of a function.

New-Fixed

### DIFF
--- a/src/debug.c
+++ b/src/debug.c
@@ -21,6 +21,10 @@
 #include "allegro5/internal/aintern_thread.h"
 #include "allegro5/internal/aintern_vector.h"
 
+#ifdef ALLEGRO_WINDOWS
+#include "allegro5/internal/aintern_wunicode.h"
+#endif
+
 #ifdef ALLEGRO_ANDROID
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


SiegeLord:
Err, isn't the correct thing is to do what https://github.com/liballeg/allegro5/pull/1017 did? I guess you want to use `_twin_utf8_to_tchar`?

dos1:
Yeah, you should convert the string to UTF-16 Windows expects, instead of switching to codepage-dependent ANSI mode. Passing UTF-8 won't work.

guilt:
> 
> 
> Yeah, you should convert the string to UTF-16 Windows expects, instead of switching to codepage-dependent ANSI mode. Passing UTF-8 won't work.

Well, getting UTF-8 to print to a log with %s isn't going to fly :) But okay, I've fixx0red it

SiegeLord:
Thanks!

PR #1045, opened 2019-05-01
Fix (complete) al_inhibit_screensaver function

rmbeer:
For the moment, this is the most optimal solution. X11 has a strange bug that seems to ignore the internal clock of the screensaver. The code must be maintained in this way until X11 is corrected. (Maybe I have to go fix it)


examples/ex_screensaver.c
```diff
@@ -0,0 +1,132 @@
+/* This example displays a picture on the screen, with support for
```
allefant: I think we don't need two examples - ex_disable_screensaver should be good enough.

.gitignore
```diff
@@ -8,6 +8,9 @@ build_android_mips64/
 build_android_mips/
 build_android_x86/
 build_android_x86_64/
+Build/
```
SiegeLord: Let's remove these, doesn't seem like there's a need to keep adding to this file for every variation of junk people add in their directories.

examples/ex_screensaver.c
SiegeLord: Yep, please remove this. If ex_disable_screensaver doesn't support something, add that to that example.

src/x/xevents.c
```diff
@@ -220,13 +220,15 @@ void _al_xwin_background_thread(_AL_THREAD *self, void *arg)
        * screensaver facility.  Probably it won't do anything for other
        * systems, though.
        */
+      #ifndef ALLEGRO_XWINDOWS_WITH_XSCREENSAVER
```
SiegeLord: We don't usually indent preprocessor directives.

src/x/xsystem.c
```diff
@@ -354,6 +360,31 @@ static bool xglx_get_cursor_position(int *ret_x, int *ret_y)
...
+   // static int temp_timeout = 0;
+
+   #ifdef ALLEGRO_XWINDOWS_WITH_XSCREENSAVER
+   if (!XScreenSaverQueryExtension(system->x11display, &temp, &temp) ||
+      !XScreenSaverQueryVersion(system->x11display, &temp_version_max, &temp_version_min) ||
+      temp_version_max < 1 || (temp_version_max == 1 && temp_version_min < 1)) {
+      return false;
+   }
+
+   /*
```
SiegeLord: What is the purpose of this commented out code? Why not remove this?

src/x/xsystem.c
```diff
@@ -354,6 +360,31 @@ static bool xglx_get_cursor_position(int *ret_x, int *ret_y)
 static bool xglx_inhibit_screensaver(bool inhibit)
 {
    ALLEGRO_SYSTEM_XGLX *system = (void *)al_get_system_driver();
+   int temp, temp_version_min, temp_version_max;
+   // int temp_interval, temp_prefer_blank, temp_allow_exp;
```
SiegeLord: What is the purpose of this commented out code? Why not remove this?

src/x/xsystem.c
```diff
@@ -354,6 +360,31 @@ static bool xglx_get_cursor_position(int *ret_x, int *ret_y)
...
+   // static int temp_timeout = 0;
+
+   #ifdef ALLEGRO_XWINDOWS_WITH_XSCREENSAVER
```
SiegeLord: No indentation for preprocessor directives.

src/x/xsystem.c
```diff
@@ -354,6 +360,31 @@ static bool xglx_get_cursor_position(int *ret_x, int *ret_y)
 static bool xglx_inhibit_screensaver(bool inhibit)
 {
    ALLEGRO_SYSTEM_XGLX *system = (void *)al_get_system_driver();
+   int temp, temp_version_min, temp_version_max;
```
SiegeLord: It makes no sense to name the last two variables `temp_`. Just call them `version_min`, `version_max`.

SiegeLord:
Subsumed by https://github.com/liballeg/allegro5/pull/1078.

PR #1047, opened 2019-05-03
al_run_detached_thread: fix segfault on detaching when the thread is already gone

dos1:
detached_thread_func_trampoline freed the outer thread at its end. If outer->proc
was really fast to finish, _al_thread_detach could get called with &outer->thread
as its argument after outer was already freed.

Usually it would be fast enough to not ever be overwritten after freeing, but tools
like asan explicitly overwrite freed memory, leading to reproducible crash.


dos1:
PS. I hope every platform lets you destroy locked mutex. If that's not the case, `_al_mutex_unlock` calls need to be added before `_al_mutex_destroy` :D

dos1:
I'm getting some segfaults with this patch now, let's hold with it.

PR #1049, opened 2019-05-04
fix android built to find opengl es

elias-pschernig:


elias-pschernig:
Actually, there is a better way - the NDK ships with a cmake toolchain which we should be able to use. I'll try and do that instead.

elias-pschernig:
This allows us to completely remove our custom Android toolchain (yay, less cmake code!) and also makes the Android build much easier in general because no more standalone toolchain has to be created.

I have not updated the section about building dependencies yet - I don't know how many of those would build with the Android-provided cmake like Allegro now does. But I intend to find out soon when I next update my Android binaries. This should not really matter for this PR.

SiegeLord:
Looks good to me, merged as 5ef154b806d8c7f5245372d9f69422098f63f039. One caveat is that the Android CMake files require a relatively high cmake version... I had to install one from Kitware's PPA.

PR #1050, opened 2019-05-04
android: use FLAG_KEEP_SCREEN_ON instead of a wakelock for screensaver inhibition

dos1:
Wakelocks are deprecated for this purpose and don't work without aquiring proper permissions, while window flag works without them and integrates properly with app switching.


SiegeLord:
Thanks!

PR #1052, opened 2019-05-12
android: use DisplayListener to listen to orientation changes

### LINKED ISSUE #1048: android: 180° orientation change not registered by Allegro
On Android, when the display orientation is changed by 180° at once, there's no actual display resize involved and `al_get_display_orientation` ends up still returning the old value.

dos1:
Fixes #1048


SiegeLord:
Thanks!

PR #1054, opened 2019-05-12
Fix implicit fallthrough compiler warnings

dos1:
Compilers scan for comment markers that let you explicitly state that you really want a fall through.


dos1:
Wait, is that pixel format thing actually correct?

dos1:
Let's make it a separate PR.

fatcerberus:
I didn’t know there were compilers that actually detected the fallthrough comments; I usually put them in as an aid to people reading the code anyway, but the compiler being able to parse them is even better.

dos1:
Yeah, it was a news for me as well :D 
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough

SiegeLord:
That regex is pretty nuts. Thanks!

PR #1055, opened 2019-05-12
Fix reported bitshifts for ARGB_4444 and RGBA_4444 pixel formats

dos1:
Something went really wrong there (and the compiler was warning about fallthrough, but nobody cared :P )


dos1:
I wonder whether the conversion is correctly done as well.

dos1:
Endianness shouldn't affect it at all, until you need to actually change the CPU representation to an explicitly defined one.

`256 >> 2` is `64` regardless of endianness.

SiegeLord:
Nice, thanks! The conversions are correct, as they are generated using a different mechanism (literally by parsing the `RGBA_4444` string). Maybe these line should have been generated by that script too...

PR #1058, opened 2019-05-16
cmake: make condition that excludes test_big_endian test more generic

dos1:
When cross-compiling to some platforms, the test doesn't work. Checking for variables set earlier instead of hardcoded set of platforms allows one to skip the test by appending proper variable to cmake command line or platform toolchain file.


SiegeLord:
Thanks!

PR #1059, opened 2019-05-16
cmake: don't link with dl on SDL platform

dos1:
It's unnecessary (Allegro doesn't use it directly) and breaks the build on platforms with no dynamic linking.


SiegeLord:
Thanks!

PR #1062, opened 2019-05-30
Make sure a display doesn't keep a reference to a destroyed bitmap

### LINKED ISSUE #1061: Intermittent crash in ex_mouse_cursor
On MacOS I am seeing intermittent crashes with ex_mouse_cursor. When I recompile with the Address Sanitiser it tells me there is a use-after-free.
Briefly:
```
==5065==ERROR: AddressSanitizer: heap-use-after-free on address 0x61400004e5d0 at pc 0x00010129ce31 bp 0x70000e5bc650 sp 0x70000e5bc648
READ of size 8 at 0x61400004e5d0 thread T3
2019-05-27 21:14:13.583029+0100 atos[5068:195577] examining /Users/USER/*/ex_mouse_cursor [5065]
    #0 0x10129ce30 in _al_ogl_finalize_fbo ogl_fbo.c:510
```
and 
```
0x61400004e5d0 is located 400 bytes inside of 424-byte region [0x61400004e440,0x61400004e5e8)
freed by thread T3 here:
    #0 0x1001312fd in wrap_free (libclang_rt.asan_osx_dynamic.dylib:x86_64+0x592fd)
    #1 0x10120c9dd in al_free_with_context memory.c:57
    #2 0x10111582b in al_destroy_bitmap bitmap.c:248
```

pedro-w:
Proposed change to fix #1061 - review needed!

### DIFF
--- a/src/opengl/ogl_bitmap.c
+++ b/src/opengl/ogl_bitmap.c
@@ -567,21 +567,21 @@ static void ogl_destroy_bitmap(ALLEGRO_BITMAP *bitmap)
 {
    ALLEGRO_BITMAP_EXTRA_OPENGL *ogl_bitmap = bitmap->extra;
    ALLEGRO_DISPLAY *disp;
-   ALLEGRO_DISPLAY *thisdisp;
+   ALLEGRO_DISPLAY *bmp_disp;
    ALLEGRO_DISPLAY *old_disp = NULL;
 
    ASSERT(!al_is_sub_bitmap(bitmap));
 
-   thisdisp = _al_get_bitmap_display(bitmap);
+   bmp_disp = _al_get_bitmap_display(bitmap);
    disp = al_get_current_display();
-   if (thisdisp->ogl_extras->is_shared == false &&
-       thisdisp != disp) {
+   if (bmp_disp->ogl_extras->is_shared == false &&
+       bmp_disp != disp) {
       old_disp = disp;
-      _al_set_current_display_only(thisdisp);
+      _al_set_current_display_only(bmp_disp);
    }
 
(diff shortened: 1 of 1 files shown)


src/opengl/ogl_bitmap.c
```diff
@@ -567,15 +567,21 @@ static void ogl_destroy_bitmap(ALLEGRO_BITMAP *bitmap)
 {
    ALLEGRO_BITMAP_EXTRA_OPENGL *ogl_bitmap = bitmap->extra;
    ALLEGRO_DISPLAY *disp;
+   ALLEGRO_DISPLAY *thisdisp;
```
SiegeLord: Would you mind naming this 'bmp_disp' to clarify what it is?

SiegeLord:
This change looks great, but seems too simple given the discussion in the bug, is this really all that's needed? One thing I'd like to verify for myself is that the bitmap reference is only stored in the display that owns it (in which case this change is sufficient).

SiegeLord:
I've spent more time looking at the code, and I'm more convinced that this change is probably correct, the `opengl_target` can only be set by the display that owns the bitmap.

src/opengl/ogl_bitmap.c
pedro-w: will do - I remember thinking at the time that `thisdisp` wasn't a good name for it...

pedro-w:
> This change looks great, but seems too simple given the discussion in the bug, is this really all that's needed? 
It does seem simple - I tried to justify it by comparing the analogous code in the directx driver but they don't really line up enough to make that approach useful.

SiegeLord:
Thanks a lot!

PR #1063, opened 2019-06-01
MacOS: make sure all relevant code runs on the main thread.

### LINKED ISSUE #876: ALLEGRO_FULLSCREEN_WINDOW has UI calls on a background thread
See https://www.allegro.cc/forums/thread/617208/1034566#target

https://github.com/liballeg/allegro5/blob/da4b7162941b8f364a02bba85b5c2a1b168f5478/src/macosx/osxgl.m#L1578

pedro-w:
Recent versions of MacOS are stricter about always running certain functions on the main thread (particularly, functions to create Cocoa objects.) Xcode includes a method to detect calls made on the 'wrong' thread. This patch fixes all reported instances.
Note that it uses `dispatch_sync`, as recommended by Apple, instead of `performSelectorOnMainThread:withObject:waitUntilDone:` on a helper object. This simplified the code. I kept uses of the latter when it was just calling a single method on a Cocoa object.
There is also a small modification to `ex_native_dialog` to demonstrate the Save dialog as well as the Open dialog.
Closes #876

### DIFF
--- a/addons/native_dialog/osx_dialog.m
+++ b/addons/native_dialog/osx_dialog.m
@@ -17,27 +17,21 @@ void _al_shutdown_native_dialog_addon(void)
 }
 #pragma mark File Dialog
 
-/* We need to run the dialog box on the main thread because AppKit is not
- * re-entrant and running it from another thread can cause unpredictable
- * crashes.
- * We use a dedicated class for this and simply forward the call.
- */
-@interface ALLEGFileDialog : NSObject
-+(void) displayFileDialog : (NSValue *)param;
-@end
-@implementation ALLEGFileDialog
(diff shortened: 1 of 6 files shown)


pedro-w:
As an aside, `ex_fs_resize` doesn't seem to work for me.

SiegeLord:
Amazing work as always, thanks a ton! I love the simplified code. Merged as https://github.com/liballeg/allegro5/commit/40d04ef9f8e01fd5f07ac827802d73b28e9bc23b.
> As an aside, ex_fs_resize doesn't seem to work for me.

Yeah, real fullscreen is hopelessly broken unfortunately.

PR #1064, opened 2019-06-27
utf8: fix al_utf8_width argument type mismatch on platforms where sizeof(int) != 4

dos1:
The function prototype already specified int32_t, it just doesn't produce any warnings, cause usually int == int32_t.


SiegeLord:
Thanks!

Issue #1041, opened 2019-04-30
Add streaming support to the minimp3 backend.


SiegeLord:

guilt:
Sidenote: We should consider looking at ffmpeg / avcodec for streaming as well.

pmprog:
Hi @SiegeLord 

I've started some work on this, but I've got a bit stuck.
I load and decode the whole file, and track the position using sample number from the decoded track. It seems to play only about half the file, and then when mp3_stream_update returns a value != buf_size, it crashes the kcm_stream when checking mutex at
https://github.com/pmprog/allegro5/blob/master/addons/audio/kcm_stream.c#L697

My source code is here:
https://github.com/pmprog/allegro5

I haven't tried the actual seeking code, because I was waiting to get the stream to load and play in full, but I can't figure out what's going wrong

SiegeLord:
Hmm, well decoding the whole file into memory seems to defeat the point of doing the streaming, so that seems like a non-starter. I looked at what minimp3 actually supports, and from what it looks like it doesn't actually support streaming... the relevant methods (`mp3dec_ex_read` and `mp3dec_ex_seek`) are stubbed out. The reading one seems not too bad to implement (and contribute back to their library), but seeking might be more tough... maybe I'd maintain a list of frame<->byte offsets.

SiegeLord:
I started implementing the streaming now, just doing a basic version which loads the whole file (without decoding) into memory, and then does basic decoding.

pmprog:
Yes, I noticed that they were stubbed - however, I did come up with an idea, but like most things, I've not really had the time to look into it any further. 

There's this function: `mp3dec_iterate_buf` which goes through checking all the frames. So what I was thinking is that on opening the MP3, make a copy of this function that scans through all the frames in the MP3, and make a note of the file position of each frame, and use that the bases of seeking.

fatcerberus:
@SiegeLord Does 5.2.6 include MP3 support?  It seems like it should, based on the commit history, but it's not mentioned in the changelog so I wanted to make sure.

SiegeLord:
Yep, 5.2.6 has it. Here it is in the changelog: https://github.com/liballeg/allegro5/blob/master/docs/src/changes-5.2.txt#L89

Note that it's unlikely that non-Windows binaries will have it, since MiniMP3 isn't actually packaged in Debian/Ubuntu/OSX. We'd need to figure something extra-ordinary for that to happen.

[...]

PR #1065, opened 2019-06-29
Add simple MP3 streaming.

SiegeLord:
This partially implements #1041.

This works by completely reading the encoded file, and then decoding it
on demand. Some things don't work:

- Can't get the length
- Can't set loop points
- Can't seek

All 3 can be solved by creating a lookup table at load time. At some
future point we shouldn't load the whole file in either, but it was
simpler to do that to start.

### DIFF
--- a/addons/acodec/mp3.c
+++ b/addons/acodec/mp3.c
@@ -28,10 +28,18 @@ typedef struct MP3FILE
    uint8_t* file_buffer;      /* encoded MP3 file */
    int64_t file_size;         /* in bytes */
    int64_t next_frame_offset; /* next frame offset, in bytes */
+
    int file_pos;              /* position in samples */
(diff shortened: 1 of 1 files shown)


SiegeLord:
Ended up adding streaming, it wasn't that hard once I RTFM'd.

fatcerberus:
@SiegeLord Any idea when this will make it into a release?  Really looking forward to mp3 support! 😃 

Also to be clear: The OP says this doesn't support length, loop points, or seeking, but the actual diff suggests these things *are* supported so... I'm confused.

SiegeLord:
I think the PR description was just the first commits message, which indeed was accurate, but when I implemented those bullet points I forgot to update it. I'll try making a release sometime this year.

PR #1070, opened 2019-08-04
Fix alpha test with the GLSL version.

allefant:
I added a commandline argument to ex_depth_mask to use the programmable pipeline. In the version using the GLSL shader it is broken. This patch fixes it so ex_depth_mask will work the same with and without the new argument:
./ex_depth_mask
./ex_depth_mask shader

The HLSL version is likely broken as well (but alpha testing may happen
outside of the shader there). I have no way to verify it.


SiegeLord:
Nice!!! I tested HLSL, and it works without this change, so I guess there's a different mechanism at play there. I think we'll want to add some 'else's in the shader, maybe to make the shader compiler's easier. Also, we need to document this.

Want me to do that, or do you want to handle it?

allefant:
Yeah, DirectX has a different approach to mixing "fixed pipeline" and shaders. Some Google results suggest alpha testing behaves differently between DirectX 9 and 10 as well - so if we ever rewrite for a newer DX the issue could return. Anyway, it's a useful feature for me so I'm glad Allegro has it :)

I can take a pass at adding those else statements, not sure they would really add any extra info to the compiler in addition to the return though. I.e.:

if (A) return;
if (B) return;
if (C) return;

vs

if (A) return;
else if (B) return;
else if (C) return;

In my mind can be optimized in the same way already.

Originally I wrote it with a switch() which should be the best hint to the optimizer - but I got a shader compiler error about there being no such statement in GLSL. Maybe I should also split the comparisons off into a separate function for readability.

And I may be able to use some recursive macro magic to get the proper enum values into the shader string - but it's a while I've played with the C pre-processor so not 100% sure I will succeed.

allefant:
Added the else statements in a separate function. I saw no way to use macro magic for the enums but I think it's more readable now.

elias-pschernig:
I guess we could always simulate a switch with an array of function pointers and have one function per comparison mode. Maybe I'll attempt that... but benchmarking glsl is hard.

allefant:
Hm, nope, no function pointers in GLSL. I found some references to implementing this with switch() - but it's not available in all GLSL versions. The real solution would be to re-create the shader on each al_set_render_state() call. It would be slow if someone calls it 1000ds of times per frame of course - but assuming the alpha function is only changed once or a handful of times it clearly would be best.

It means we'd need a new subsystem in Allegro to dynamically create shaders and probably also some sort of shader cache so we don't recompile from the string each time a specific version is needed.

So, for now I think the patch is fine. Hardware is good with branching if the condition doesn't change so I'm quite sure the extra if for the default case won't matter (and there already is another branch like that in the shader for texturing on/off which has the same problem).

SiegeLord:
Yeah, for sure. I am certain the chain of if/else's will get compiled to efficient code. In modern games back in the day (maybe still today?) they'd make mega shaders with all sorts of control flow like this, so I'm sure the driver writers optimized for this kind of code.

SiegeLord:
Thanks Elias!

PR #1071, opened 2019-08-06
implement ALLEGRO_FRAMELESS on Android

allefant:
This should be used for #811 - same idea but this version uses the ALLEGRO_FRAMELESS flag and also will work dynamically with al_set_display_flag.


SiegeLord:
Nice, this works great! Thanks!

PR #1074, opened 2019-08-22
Include limits.h for PTHREAD_STACK_MIN

gusnan:
We get a build error on hurd in Debian, see 
https://buildd.debian.org/status/fetch.php?pkg=allegro5&arch=hurd-i386&ver=2%3A5.2.5.0-1&stamp=1566474705&raw=0

for full build log.


SiegeLord:
Thanks!

PR #1075, opened 2019-08-26
No PTHREAD_STACK_MIN on Hurd

gusnan:
This patch also removes the #include <limits.h> which didn't
work - The proper fix is to simply not check for PTHREAD_STACK_MIN
on Hurd, since it's not defined, and the real minimum on Hurd
is much smaller than on Linux.

See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=823049
for the same fix that was applied on xscreensaver.

### DIFF
--- a/include/allegro5/platform/aintuthr.h
+++ b/include/allegro5/platform/aintuthr.h
@@ -5,7 +5,6 @@
 #define __al_included_allegro5_aintuthr_h
 
 #include <pthread.h>
-#include <limits.h>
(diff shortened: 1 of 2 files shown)


gusnan:
And here's the buildlog with the patch applied:
https://buildd.debian.org/status/fetch.php?pkg=allegro5&arch=hurd-i386&ver=2%3A5.2.5.0-2&stamp=1566908133&raw=0

PR #1081, opened 2019-09-07
Fix building for GNU/Linux without GNU extensions enabled

dos1:
GCC doesn't define alloca, and timerclear isn't defined when compiling in standards mode that disables GNU extensions.


SiegeLord:
Thanks!

PR #1082, opened 2019-09-07
the alpha test commit broke GLES, this fixes it

allefant:
There is a bit of a chicken and egg problem - we need to call al_set_target_bitmap first because setting the shader requires the display. But we need to do the setup after setting the shader.

### DIFF
--- a/include/allegro5/internal/aintern_opengl.h
+++ b/include/allegro5/internal/aintern_opengl.h
@@ -119,6 +119,7 @@ typedef struct ALLEGRO_OGL_VARLOCS
    GLint tex_loc;
    GLint use_tex_matrix_loc;
    GLint tex_matrix_loc;
+   GLint alpha_test_loc;
    GLint alpha_func_loc;
    GLint alpha_test_val_loc;
    GLint user_attr_loc[_ALLEGRO_PRIM_MAX_USER_ATTR];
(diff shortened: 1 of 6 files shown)


allefant:
Maybe a better solution would be to simply initialize the uniform to have alpha testing disabled - however uniforms can only be initialized in GLES >= 1.2. Not that it should be a problem.

I basically decided to skip adding a separate uniform for turning alpha testing on/off since setting the alpha test function to ALWAYS has the same effect as turning it off - but since the value 0 means NEVER by default this new shader will not render anything at all. So with normal OpenGL, the problem only appears if PROGRAMMABLE_PIPELINE is used and no calls to al_set_render_state are made. The default is not to use the programmable pipeline so it managed to escape my (much too limited) testing :/

allefant:
I fixed it in a different way now, adding another uniform. This more directly mirrors the alpha test API and is probably how I should have done it to begin with. This way the uniform is initialized to false and the shader will work as expected.

The original fix had the problems that any call to al_use_shader() would reset the behavior to not drawing anything.

allefant:
And one last commit to make sure cases like these work:

    al_set_render_state(ALLEGRO_ALPHA_TEST, true);
    al_use_shader(NULL);

Passing NULL to al_use_shader will load the default shader when using programmable pipeline, and so the alpha testing uniforms have to be re-synchronized afterwards.

There is an existing TODO in _al_ogl_update_render_state about only making all the OpenGL calls when needed - I may work on that soon. As part of that will also only update the uniforms if the state is different from the default so then no extra OpenGL calls would be made.

SiegeLord:
Looks good to me, thanks Elias!

PR #1083, opened 2019-09-07
iio: refine the order of image loaders

dos1:
We don't want to use freeimage or android loaders in cases where we
already have some specialized ones available.

### DIFF
--- a/addons/image/iio.c
+++ b/addons/image/iio.c
@@ -52,25 +52,6 @@ bool al_init_image_addon(void)
    should be used. i.e., ALLEGRO_CFG_IIO_HAVE_GDIPLUS and
    ALLEGRO_CFG_IIO_HAVE_PNG will never both be set. */
 
-#ifdef ALLEGRO_CFG_IIO_HAVE_FREEIMAGE
-   {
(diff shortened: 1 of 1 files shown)


allefant:
Maybe a better solution would be to simply initialize the uniform to have alpha testing disabled - however uniforms can only be initialized in GLES >= 1.2. Not that it should be a problem.

I basically decided to skip adding a separate uniform for turning alpha testing on/off since setting the alpha test function to ALWAYS has the same effect as turning it off - but since the value 0 means NEVER by default this new shader will not render anything at all. So with normal OpenGL, the problem only appears if PROGRAMMABLE_PIPELINE is used *and* no calls to al_set_render_state are made. The default is not to use the programmable pipeline so it managed to escape my (much too limited) testing :/

SiegeLord:
Ok, let's merge this and see if anybody complains in which case we'll add a configuration option to control this order. Thanks!

dos1:
Just for the reference: freeimage taking over the loading of webp assets was breaking my game, as it wasn't properly supporting the kind of lossless files I've been using :P

PR #1084, opened 2019-09-10
Add instructions to install Allegro using vcpkg

vicroms:
Vcpkg is a C++ library manager that is able to build and install Allegro and its dependencies using a single line command:

`vcpkg install allegro5`

This PR adds instructions on how to acquire Allegro using vcpkg with the purpose of making it easier for users to install and use Allegro.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -72,6 +72,16 @@ The addons, too, may require additional libraries.  Since the addons are
 strictly optional, they are not required to build Allegro, but a lot of
 functionality may be disabled if they are not present.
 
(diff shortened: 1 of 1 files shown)


SiegeLord:
Thanks! I don't know if the readme is the right location for this; could I convince you to add this to the wiki instead? https://github.com/liballeg/allegro_wiki/wiki/Quickstart

guilt:
I think it's handy for a new dev to see that file there, as opposed to look for it in other places. The README also points people to precompiled binaries and what not :)

PR #1092, opened 2019-10-20
Fix regex on line 65 of generate_python_ctypes.py

zachs18:
The previous regex on line 65 r"\bstruct|union\b" matches "struct" at the beginning of a word or "union" at the end of a word (e.g. it matches twice on "structured reunion" and re.sub will give "ured re").
The proposed regex r"\b(struct|union)\b" matches "struct" or "union" as a whole word only.

### DIFF
--- a/python/generate_python_ctypes.py
+++ b/python/generate_python_ctypes.py
@@ -62,7 +62,7 @@ class Allegro:
             "postprocess_callback_t": c_void_p,
             }
 
-        ptype = re.sub(r"\bstruct|union\b", "", ptype)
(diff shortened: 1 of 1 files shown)


elias-pschernig:
Good catch! I tried with and without this applied and the generated allegro.py is identical, so I assume this never actually encountered one of the cases where it would fail?

zachs18:
Yeah, I think the original regex would only fail if it encountered a struct or union with a name starting with "struct" and/or ending with "union". There aren't any currently (and it's unlikely there will be), so the output is the same.

SiegeLord:
Thanks!

PR #1093, opened 2019-10-20
fix mp3 files with multiple channels

elias-pschernig:
I tested adding some sounds from this pack to my game, and since we now support mp3 just used the mp3 files as is: https://opengameart.org/content/monster-and-creatures-sound-effects-pack-001

The files play fine, however once Allegro is done playing any of them it crashes. Turns out there is some kind of miscalculation for the sample length. This PR fixes it for me but someone who knows MiniMp3 should double check.


SiegeLord:
That looks right, judging by the minimp3 source: https://github.com/lieff/minimp3/blob/master/minimp3_ex.h#L115,L120

Thanks!

Issue #1091, opened 2019-10-19
Mac OS Catalina al_create_display() crashes 


mbkorecki:
Hi,

I am experiencing the following problem: 
I am able to compile my allegro application with a make file without errors, however when I run the ./file I get a "zsh: illegal hardware instruction  ./file" message. 
I commented out some parts of the code and it seems that the issue is cause (or perhaps this is only one of the casuses) by the al_create_display() function. I replicated the problem with a very simple allegro program (basically just opening a display). 

My allegro is installed with homebrew and I compile the program with g++-9. 
```
Some information from the crash report:
Crashed Thread:        3

Exception Type:        EXC_BAD_INSTRUCTION (SIGILL)
Exception Codes:       0x0000000000000001, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Illegal instruction: 4
Termination Reason:    Namespace SIGNAL, Code 0x4
Terminating Process:   exc handler [1579]

Application Specific Information:
-[NSOpenGLContext update] must be called from the main thread if the context has a view.

And the thread that crashed:
Thread 3 Crashed:
0   com.apple.AppKit              	0x00007fff3a4878a1 -[NSOpenGLContext update] + 520
1   liballegro.5.2.dylib          	0x0000000102876943 create_display_win + 1079
2   liballegro.5.2.dylib          	0x000000010283b3d4 al_create_display + 40
3   file                          	0x0000000102724e72 Display::Display() + 130
4   file                          	0x0000000102724f22 Display::instance() + 62
5   file                          	0x00000001026e2f4f _al_mangled_main + 508
6   liballegro.5.2.dylib          	0x0000000102878d47 call_user_main + 23
7   liballegro.5.2.dylib          	0x0000000102878d30 +[AllegroAppDelegate app_main:] + 9
8   com.apple.Foundation          	0x00007fff3f58cef2 __NSThread__start__ + 1064
9   libsystem_pthread.dylib       	0x00007fff74405d76 _pthread_start + 125
10  libsystem_pthread.dylib       	0x00007fff744025d7 thread_start + 15
```

beoran:
From what I hear Catalina breaks a lot of software, unfortunately Allegro seems to be one of them. Could you please post the program that produces the crash above?

wgnrd:
Hi,
I'm experiencing a similar problem with Catalina on a Macbook Pro 2018 (no touchbar).

This is a simple program to reproduce the error with.
``` c
[...]

PR #1094, opened 2019-10-24
fix osx catalina

allefant:


allefant:
Unfortunately could not find any documentation on what the update() call is supposed to do exactly, but the skater demo works again with this on OSX Catalina, including window resizing.

elias-pschernig:
Oh, and I forgot to mention, this also places the activateIgnoringOtherApps call on the main thread - which turned out to be just a harmless warning, but is still the right thing to do.

Anyway, issue #1091 seems to be fixed with this, unless we discover the original purpose of the update() call.

elias-pschernig:
Well, after figuring out the difference in objective C between @selector(update) and @selector(update:) it now does call update again, from the main thread. Should be 100% identical to before now, just works in Catalina.

djn658:
Thank you very much! It works now.

mbkorecki:
Hi, thanks a lot for the fix. I wanted to check it out but I am having issues building allegro. I tried building with gcc/g++-9 and with clang as well as an xcode project. In each case I get errors (different depending on compilers). For clang I get an error stating that the problem is: "use of undeclared identifier 'errno'". It seems weird as the file that throws the error does include the errno.h. May I ask how you build allegro on your system? Perhaps someone experienced this error as well?

mbkorecki:
Thanks a lot. It works for me as well.

SiegeLord:
Thanks for all the testing everyone, and Elias for writing this!

PR #1095, opened 2019-10-25
Xcode in osx

elias-pschernig:


mbkorecki:
It works for me as described.

SiegeLord:
Lol, that's how I always built it... didn't know other ways worked. Thanks Elias.

PR #1102, opened 2019-12-11
Fix for window resizing on Windows when a menubar is visible

kwellwood:
This fix enables AdjustWindowRectEx to correctly calculate the window's client area. I have tested this fix since 5.2.2 without any issues.


SiegeLord:
Looks good to me, thanks a lot!

PR #1106, opened 2020-01-02
Change allegro graphic to full screen in execution time

saisilcastro:
I am updating my engine while developing my game menu. I use two libraries to manipulate the engine. One of them is allegro 5.2.5. I am trying to change the graphic mode in execution time but until now I could not. I've tried to use al_toggle_display_flag, al_set_display_flag and other functions, but none of them is able to change from window screen to fullscreen and vice-versa.
It could be really good to have a function like al_set_window_fullscreen which could be able to change the window mode in execution time.
Thanks for the attention.

Lísias de Castro Martins

### DIFF
--- a/docs/src/changes-5.2.txt
+++ b/docs/src/changes-5.2.txt
@@ -2,6 +2,91 @@
 
 These lists serve as summaries; the full histories are in the git repository.
 
+Changes from 5.2.3 to 5.2.4 (February 2018)
+==========================================
+
+The main developers this time were: Sebastian Krzyszkowiak, Elias Pschernig,
+SiegeLord
+
+Core:
+
+  - Fix errors when reading/writing 0 byte buffers (Bruce Pascoe).
+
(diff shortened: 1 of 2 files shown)


elias-pschernig:
You can already use this:

al_set_display_flag(display, ALLEGRO_FULLSCREEN_WINDOW, true)

However it will not change the monitor resolution, just set your window to be fullscreen. I agree that a way to modify the monitor resolution without recreating the window would also be useful (even if internally it will have to re-create the window anyway).

saisilcastro:
I have tried this one. But the screen does not keep centered compared to the "real" fullscreen mode neither stretch the screen in order to show the whole image "closer" than the higher resolution. It makes the application appearance weird. Is there some way in order to make this with this function?

saisilcastro:
Ah, about the version, that was my mistake. I am using the 5.2.5 version, not the 5.2.4.

PR #1107, opened 2020-01-03
Adds floating point coordinate mouse events.

### LINKED ISSUE #1105: How to get float values from mouse functions
hello everyone, sorry for my bad english

I'm doing the air hockey video game. to move the player's handle I use the mouse.
the problem is that the mouse, mouse.dx (int) and mouse.dy (int) functions return integer values. for greater accuracy of the player vector I need to have float values.
the windows operating system returns float values for mouse functions.
how can i get float values with Allegro5 functions?
thank you

SiegeLord:
Now you can catch _FLOAT variants of mouse events which have floating
point coordinates. The regular mouse events are still emitted alongside
the new events.

I haven't added the corresponding floating point API for mouse state and
mouse warping, as it doesn't seem as urgent. We can add it later if
requested.

Fixes #1105

### DIFF
--- a/src/linux/lmseev.c
+++ b/src/linux/lmseev.c
@@ -54,7 +54,7 @@ typedef struct AL_MOUSE_EVDEV
 {
    ALLEGRO_MOUSE parent;
    int fd;
-   ALLEGRO_MOUSE_STATE state;
+   ALLEGRO_MOUSE_FLOAT_STATE state;
 } AL_MOUSE_EVDEV;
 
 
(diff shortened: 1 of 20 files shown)


src/linux/lmseev.c
```diff
@@ -179,10 +179,10 @@ static int in_to_screen(const AXIS *axis, int v)
  *  returns the new screen position, given the input relative one.
  *  The tool mode is always relative
  */
-static int rel_event(AXIS *axis, int v)
+static float rel_event(AXIS *axis, int float)
```
allefant: Is this really a variable named "float"? :)

SiegeLord:
Thanks Elias. I'm going to abandon this change, as it's unimplementable as is (namely, we can't reasonably produce floating point absolute coordinates). Instead, we'll add a raw-input mouse events which only send the deltas, like SDL and GLFW do.

SergioMayo:
sorry for my bad english
I didn't understand what was done:
will the mouse continue to return only integers (int)?

SiegeLord:
There's no code for this at the moment, but what we'll do is add a new event, called `ALLEGRO_EVENT_RAW_MOUSE_AXES`, which will still have int `dx` and `dy` like the regular `ALLEGRO_EVENT_MOUSE_AXES`, but they'll be higher resolution, so you can detect smaller mouse movements. You'll still do `float_dx = scaling_factor * event.raw_mouse.dx`, but it'll be more precise.

PR #1110, opened 2020-01-19
Setup Windows Github CI.

SiegeLord:


SiegeLord:
The issue in this PR is that we need a DX9 SDK installed. Honestly, I'd rather make Allegro not need DX9 SDK rather than figure out how to get DX9 SDK to install on the CI servers.

PR #1114, opened 2020-01-22
Fix conversion of ALLEGRO_USTR to wchar.

SiegeLord:
This fixes file patterns in the native dialog addon, which had embedded
NULs.


SiegeLord:
Good catch on the zero length strings, I fixed the code. Also added the fix for the clipboard.

PR #1115, opened 2020-02-02
Move another method to the main thread

pedro-w:
`setApplicationIconImage:` needs to be run from the main thread (seen when running ex_icon under the debugger)
Also fixed calling `activateIgnoringOtherApps:` - if a method takes an argument that is not an `NSObject` it needs to be wrapped in an `NSObject` before using the `performSelectorOnMainThread:` mechanism.

(MacOS Mojave 10.14.6)

### DIFF
--- a/src/macosx/osxgl.m
+++ b/src/macosx/osxgl.m
@@ -2319,8 +2319,8 @@ static void set_icons(ALLEGRO_DISPLAY *display, int num_icons, ALLEGRO_BITMAP* b
    NSImage *image = NSImageFromAllegroBitmap(bitmap);
    (void)display;
    [NSApp performSelectorOnMainThread: @selector(setApplicationIconImage:)
-                           withObject: image
-                        waitUntilDone: YES];
+      withObject: image
+      waitUntilDone: YES];
    [image release];
    [pool drain];
 }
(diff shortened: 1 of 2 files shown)


src/macosx/osxgl.m
```diff
@@ -2318,7 +2318,9 @@ static void set_icons(ALLEGRO_DISPLAY *display, int num_icons, ALLEGRO_BITMAP* b
    NSAutoreleasePool *pool = [[NSAutoreleasePool alloc] init];
    NSImage *image = NSImageFromAllegroBitmap(bitmap);
    (void)display;
-   [NSApp setApplicationIconImage: image];
+   [NSApp performSelectorOnMainThread: @selector(setApplicationIconImage:)
+                           withObject: image
```
SiegeLord: Do you mind fixing the indentation here and in the other file?

pedro-w: Not sure if this was what you wanted but I have indented 3 spaces instead of what was there before (Xcode likes to auto-align the colons, I left it because system.m was done that way)
https://github.com/liballeg/allegro5/blob/7e3ba4b5856bf8ac1008bed5447bb26477a8a5af/src/macosx/system.m#L66-L72

SiegeLord:
Thanks!

Issue #1116, opened 2020-02-16
Allegro 4.4.3: DirectInput mouse & keyboard are not working under Wine


ghost:
Forgive me for bringing the old 4.4 up again, it's just that our project still uses it (although we are in process of upgrading to modern backend).

We recently realized that after upgrade from Allegro 4.4.2 to 4.4.3 mouse and keyboard stopped working under Wine. After short investigation, it was found that initalization fails here:
https://github.com/liballeg/allegro5/blob/4.4/src/win/wmouse.c#L618
HRESULT is 0x80040154, which means "Class not registered".

I am not a Wine expert, but if I do a quick search in their mirror repo (https://github.com/wine-mirror/wine) it seems that mouse & keyboard Em2 classes are only declared but not implemented.

Regarding the change in Allegro, there's a distinct difference between 4.4.2 and 4.4.3 there:
https://github.com/liballeg/allegro5/commit/234ac0ab5b5790ff6d63e0433c89007f491cefde#diff-33325ce8d09b066271fbb6eae5780288R688

This change was a part of commit: 234ac0ab5b5790ff6d63e0433c89007f491cefde
It struck me that this particular change does not seem to relate to the commit's description and purpose.
But then I found this forum thread from 2015:
https://www.allegro.cc/forums/thread/615012

with the following exchange:

> I understand most of the patch, but one bit confuses me... what's up with the GUID_SysMouseEm2 and GUID_SysKeyboardEm2?

> That was my feeble attempt to use the emulation layer so Visual Studio would still be usable on break points.
> **However, although this did not fix anything, it doesn't seem to make a difference whether to use GUID_SysMouse, GUID_SysMouseEm or GUID_SysMouseEm2**. I simply didn't bother to change it back. (Lazy me, yes...)

It is confirmed by our user that changing GUID back fixes problem on Wine.

Would it be possible to revert this, or make this GUID configurable?

SiegeLord:
I don't mind reverting this, feel free to send a pull request.

PR #1117, opened 2020-02-17
4.4: fixed DirectInput not working under Wine

ivan-mogilko:
Problem details are here: #1116

This reverts couple of changes from 234ac0ab5b5790ff6d63e0433c89007f491cefde which were not relevant to original commit, but broke DirectInput driver on Wine, because Em2 keyboard and mouse COM classes are not implemented.


SiegeLord:
Thanks!

PR #1143, opened 2020-06-10
audio: Fix edge-case looping in Ogg Vorbis stream

codylico:
When playing an Ogg Vorbis stream, the loop case is handled partially
incorrectly. The number of channels in the stream was not taken into
account when advancing the next read length to a multiple of the frame
size. This problem manifests with carefully crafted (or accidentally
input) loop points.

Enhances commit 20ed5af76b2a1cd00fda3c3ceaa86e76366fb828.


beoran:
I didn't get the chance to test this but the change is small and looks plausible.

codylico:
Take as much time as you need to test it. I tested it a little myself, but there may be more cases I don't know about that the code needs to handle.

SiegeLord:
This is great, the old code was horribly broken (the way you said it, and the rounding up to what it thought the frame size was wrong too), and it was easy to reproduce the clicking (which your change fixed). Thanks for fixing this!

PR #1147, opened 2020-06-20
fix wrong LC_ALL: x/xkeyboard.c Changes setlocale to  LC_CTYPE 

ericoporto:
This fixes allegro changing the locale for LC_NUMERIC and more when installing keyboard in the x platform. This is for allegro 4.

Mention in the forums: https://www.allegro.cc/forums/thread/609919
Fixed in newer allegro: https://github.com/liballeg/allegro5/blob/ce68911b41684d6353b23ac06ec6cd6e913d87d1/src/x/xkeyboard.c#L743


SiegeLord:
Seems reasonable, thanks!

PR #1148, opened 2020-06-21
allegro 4.4: fixes wrong assert in mouse.c in set_mouse_range

ericoporto:
minor fix for a wrong assert in the mouse.c code when setting the mouse range in `set_mouse_range`.

It used to do this:
```C
ASSERT(y2 >= y2);
```

which doesn't make sense. To make sure the mouse boundaries are right the assert should be:

```C
ASSERT(y2 >= y1);
```

for a rectangle from (x1,y1) to (x2,y2).


SiegeLord:
Nice, thanks!

PR #1155, opened 2020-08-11
Migrate to GTK3

gusnan:
Debian is working to reduce the dependencies on GTK2, and has posted a bug on the Allegro5-package:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=967247

I have done some work on this migrating to GTK3, and got some ways through - see the diffs here in this pull request.

The problems I have ran in to: 

First - and the biggest problem: I am not sure at all about how the function _really_make_transient_ should work. What I have in this pull request works for me, but chances are great that there's some other simpler method using internal Allegro calls. A my diff is now, works on my system, but it surely can be nicer.

Number two is regarding icons in GTK menus. GTK_STOCK_* icons has been deprecated, so in my work I have simply removed the icons here. 

Number three: Fonts in GTK uses GTK_STYLEs and css styling. In my case it affects the log window, which doesn't get a monospace font. Really minor problem to me.

Other than this, on my system this builds, and works alright in the testing I have made.

Please review and test my changes and help out with the _really_make_transient_ function.

### DIFF
--- a/addons/native_dialog/gtk_menu.c
+++ b/addons/native_dialog/gtk_menu.c
@@ -85,14 +85,6 @@ static void checkbox_on_toggle(ALLEGRO_MENU_ITEM *item)
 }
 
 /* [gtk thread] */
-static void destroy_pixbuf(guchar *pixels, gpointer data)
-{
-   (void) data;
-   
-   al_free(pixels);
-}
-
-/* [gtk thread] */
 static GtkWidget *build_menu_item(ALLEGRO_MENU_ITEM *aitem)
 {
    GtkWidget *gitem;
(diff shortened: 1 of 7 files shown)


addons/native_dialog/gtk_menu.c
```diff
@@ -116,39 +108,8 @@ static GtkWidget *build_menu_item(ALLEGRO_MENU_ITEM *aitem)
...
-               aitem->extra2 = gtk_image_new_from_pixbuf(pixbuf);
-               
-               gtk_image_menu_item_set_image(GTK_IMAGE_MENU_ITEM(gitem), aitem->extra2);
-               
-               /* Subtract the main reference. the image still holds a reference, so the
-                * pixbuf won't be destroyed until the image itself is. */
-               g_object_unref(pixbuf);
-            }
-         }
+         gitem = gtk_menu_item_new_with_label(al_cstr(caption));
```
SiegeLord, line 111: This should be `gtk_menu_item_new_with_mnemonic`

gusnan: Oops - new commit pushed to fix this. Thanks!

gusnan:
I am looking into the monospace bit - It might not be as hard as I thought.

gusnan:
I have added a fix so that it (as the commitmessage) don't change font size, only set monospace font if monospace is requested.

SiegeLord:
Thanks!

PR #1158, opened 2020-08-16
Fix segfault in al_draw_ribbon() when num_segments>128

### LINKED ISSUE #1157: segfault when trying to draw ribbons with more than 128 points
`ALLEGRO_VERTEX_CACHE_SIZE` is fixed, and small.

It would be nice to be able to draw arbitrarily long ribbons, but I could split them on my side.
Anyway, it should not segfault on Allegro side.

The simplest solution is to include a simple check at the beginning of `al_draw_ribbon()`:

    if (num_segments*2 > ALLEGRO_VERTEX_CACHE_SIZE) {
        fprintf(stderr,
            "Allegro is refusing to draw a ribbon with more than %d points.\n",
            ALLEGRO_VERTEX_CACHE_SIZE/2);
        return;
    }

bart9h:
Fixes #1157

### DIFF
--- a/addons/primitives/high_primitives.c
+++ b/addons/primitives/high_primitives.c
@@ -1248,6 +1248,11 @@ void al_draw_ribbon(const float *points, int points_stride, ALLEGRO_COLOR color,
    LOCAL_VERTEX_CACHE;
    int ii;
 
+   if (num_segments*2 > ALLEGRO_VERTEX_CACHE_SIZE) {
+      fprintf(stderr, "Ribbon has too many segments.\n");
+      return;
+   }
+
    al_calculate_ribbon(&(vertex_cache[0].x), sizeof(ALLEGRO_VERTEX), points, points_stride, thickness, num_segments);
    
    if (thickness > 0) {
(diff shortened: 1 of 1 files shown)


pedro-w:
I don't think this is right because, if thickness is zero, you can draw up
to ALLEGRO_VERTEX_CACHE_SIZE segments (agreed it's only
ALLEGRO_VERTEX_CACHE_SIZE/2 if thickness > 0)

addons/primitives/high_primitives.c
```diff
@@ -1248,6 +1248,11 @@ void al_draw_ribbon(const float *points, int points_stride, ALLEGRO_COLOR color,
    LOCAL_VERTEX_CACHE;
    int ii;
 
+   if (num_segments*2 > ALLEGRO_VERTEX_CACHE_SIZE) {
+      fprintf(stderr, "Ribbon has too many segments.\n");
```
pedro-w, line 1252: Probably should use `ALLEGRO_ERROR` here?

bart9h:
Now I don't know if I should have left that \n or not..

SiegeLord:
Thanks!

PR #1161, opened 2020-08-23
Added missing include that was preventing build from completing on MacOS.

NewCreature:
Commit 0b925657edc1c7b08c9cce4d3062dee74ea2f232 added an `_al_add_exit_func()` call, but did not `#include` the file with the function declaration. This was causing the build to fail for me under MacOS with an implicit declaration error.


SiegeLord:
Merged via https://github.com/liballeg/allegro5/commit/f1eacca71b4ba09980397220c20a13594ce4c203.

Thanks!

Issue #1151, opened 2020-07-18
Supply modifiers with key events - Issue for pre-AT keyboards


SteveFosdick:
A pre-AT keyboard (for example IBM type F with original PC or XT) does not have the separate cursor and Home, End, Insert, Delete, PgUp and PgDown keys.  To support such a keyboard an application needs to look at the state of NumLock and thus treat the numeric keypad keys as numbers (if numlock is engaged) or as these other functions if it is not.

At the moment Allegro does not supply the key modifiers (shift state) with ALLEGRO_EVENT_KEY_DOWN and ALLEGRO_EVENT_KEY_UP events, only with ALLEGRO_EVENT_KEY_CHAR events.

Whilst one might expect an application to be only interested in the state of NumLock when processing ALLEGRO_EVENT_KEY_CHAR events this presents an issue when the application using Allegro is emulating another machine and needs the up/down events to emulate the keyboard on that machine.

Please populate the modifiers field of the key up/down events to resolve this.

SiegeLord:
We now populate the modifiers, please verify that this actually works for this use case.

SteveFosdick:
Thanks very much for doing this.  I have tried testing with the latest git master but this crashes with what I suspect is an unrelated error:
```
(b-em:18168): Gtk-WARNING **: 19:13:59.698: GtkSocket: only works under X11
(b-em:18168): Gdk-WARNING **: 19:13:59.700: ../gtk/gdk/x11/gdkwindow-x11.c:5633 drawable is not a native X11 window
```
then:
```
Thread 11 "gtk thread" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffd6ffd640 (LWP 18308)]
0x00007ffff68dca39 in ?? () from /usr/lib/libX11.so.6
(gdb) bt
#0  0x00007ffff68dca39 in  () at /usr/lib/libX11.so.6
#1  0x00007ffff68dd2a5 in _XFlush () at /usr/lib/libX11.so.6
#2  0x00007ffff68dff46 in _XGetRequest () at /usr/lib/libX11.so.6
#3  0x00007ffff68c38ab in _XGetWindowAttributes () at /usr/lib/libX11.so.6
#4  0x00007ffff68c3a9b in XGetWindowAttributes () at /usr/lib/libX11.so.6
#5  0x00007ffff718c367 in  () at /usr/lib/libgtk-3.so.0
[...]

PR #1162, opened 2020-08-24
Modifier keys for ALLEGRO_EVENT_KEY_DOWN and ALLEGRO_EVENT_KEY_UP

aldrikboy:
Fix for issue #1151


SiegeLord:
Looks good, I only tried in on Linux, but it looks right everywhere else so seems okay to merge this.

Issue #1100, opened 2019-12-04
al_get_display_refresh_rate() on Windows (tested on Windows 10, two different PCs)


mikiZX:
Issue: the function returns 0 (zero) when using windowed mode though calling Windows.h function EnumDisplaySettings that provides the refresh rate of the monitor returns correctly the actual monitor refresh rate even from the windowed thread.

output of ex_monitorinfo on PC where this was observed:

1 adapters found...
Adapter 0: (0, 0) - (1280, 1024)
Available fullscreen display modes:
Mode 0: 640 x 480, 60 Hz
Mode 1: 640 x 480, 72 Hz
Mode 2: 640 x 480, 75 Hz
Mode 3: 640 x 480, 59 Hz
Mode 4: 720 x 480, 56 Hz
Mode 5: 720 x 480, 60 Hz
Mode 6: 720 x 480, 72 Hz
Mode 7: 720 x 480, 75 Hz
Mode 8: 720 x 576, 56 Hz
Mode 9: 720 x 576, 60 Hz
Mode 10: 720 x 576, 72 Hz
Mode 11: 720 x 576, 75 Hz
Mode 12: 800 x 600, 56 Hz
Mode 13: 800 x 600, 60 Hz
Mode 14: 800 x 600, 72 Hz
Mode 15: 800 x 600, 75 Hz
Mode 16: 1024 x 768, 60 Hz
Mode 17: 1024 x 768, 70 Hz
Mode 18: 1024 x 768, 75 Hz
Mode 19: 1152 x 864, 75 Hz
Mode 20: 1280 x 720, 60 Hz
Mode 21: 1280 x 720, 75 Hz
Mode 22: 1280 x 720, 59 Hz
Mode 23: 1280 x 768, 60 Hz
Mode 24: 1280 x 768, 75 Hz
Mode 25: 1280 x 800, 60 Hz
Mode 26: 1280 x 800, 75 Hz
Mode 27: 1280 x 960, 60 Hz
Mode 28: 1280 x 960, 75 Hz
Mode 29: 1280 x 1024, 60 Hz
Mode 30: 1280 x 1024, 75 Hz

And ex_displayoptions output (png file):
https://imgur.com/a/pluFnRe 

Running the following code gives back output of 'refresh rate:0' (please disregard the FPS and  timer setup part):
//-----------------------------------------
const float FPS = 60;

int main(int argc, char* argv[])
{
ALLEGRO_DISPLAY* display = NULL;
[...]

PR #1163, opened 2020-08-24
Set display refresh_rate in d3d

aldrikboy:
Fix issue #1100


SiegeLord:
This seems to work, but the original bug was incorrect: OpenGL doesn't fill out this field on Windows either so. I'll merge this and open another but about OpenGL.

Thanks!

PR #1164, opened 2020-08-29
Reverse audio playmode

aldrikboy:
Implementation of ALLEGRO_PLAYMODE_REVERSED to play samples backwards,
this should work on all audio drivers that support ALLEGRO_PLAYMODE_BIDIR.

This also includes an implementation for al_get_audio_driver_specifier(void) to get the name of the active audio driver.

I have only tested this on DirectSound.


SiegeLord:
This is an interesting idea, but I don't think a new playmode is necessary. Conceptually, backwards playing is just playing with `speed = -1` (which is poorly supported at the moment, but we could fix it). We could also make the `al_play_sample` start at the end of the sample if the speed is negative. Something like this:

```diff
diff --git a/addons/audio/kcm_mixer.c b/addons/audio/kcm_mixer.c
index d042fdf95..6fcf74f02 100644
--- a/addons/audio/kcm_mixer.c
+++ b/addons/audio/kcm_mixer.c
@@ -215,7 +215,7 @@ static bool fix_looped_position(ALLEGRO_SAMPLE_INSTANCE *spl)
          return true;
 
       case ALLEGRO_PLAYMODE_ONCE:
-         if (spl->pos < spl->spl_data.len) {
+         if (spl->pos < spl->spl_data.len && spl->pos > 0) {
             return true;
          }
          spl->pos = 0;
diff --git a/addons/audio/kcm_sample.c b/addons/audio/kcm_sample.c
index 96576fd4b..bb73fe5ea 100644
--- a/addons/audio/kcm_sample.c
+++ b/addons/audio/kcm_sample.c
@@ -375,14 +375,20 @@ bool al_play_sample(ALLEGRO_SAMPLE *spl, float gain, float pan, float speed,
 static bool do_play_sample(ALLEGRO_SAMPLE_INSTANCE *splinst,
    ALLEGRO_SAMPLE *spl, float gain, float pan, float speed, ALLEGRO_PLAYMODE loop)
 {
+   unsigned int start_pos = 0;
    if (!al_set_sample(splinst, spl)) {
       ALLEGRO_ERROR("al_set_sample failed\n");
       return false;
    }
 
+   if (speed < 0.) {
+      if (al_get_sample_instance_length(splinst) > 1)
+         start_pos = al_get_sample_instance_length(splinst) - 1;
+   }
    if (!al_set_sample_instance_gain(splinst, gain) ||
          !al_set_sample_instance_pan(splinst, pan) ||
          !al_set_sample_instance_speed(splinst, speed) ||
+         !al_set_sample_instance_position(splinst, start_pos) ||
          !al_set_sample_instance_playmode(splinst, loop)) {
       return false;
    }
```

Issue #1165, opened 2020-08-30
al_get_display_refresh_rate not working on WGL


SiegeLord:

ReiquelApplegate:
By WGL I suppose you mean https://learn.microsoft.com/en-us/windows/win32/opengl/wgl-functions ?

SiegeLord:
WGL is Windows OpenGL in this case.

PR #1166, opened 2020-08-30
al_get_display_refresh_rate fix under wgl

aldrikboy:
set refresh_rate for ALLEGRO_FULLSCREEN, ALLEGRO_FULLSCREEN_WINDOW and windowed mode under wgl

fix issue #1165

### DIFF
--- a/include/allegro5/monitor.h
+++ b/include/allegro5/monitor.h
@@ -16,6 +16,7 @@ typedef struct ALLEGRO_MONITOR_INFO
    int y1;
    int x2;
    int y2;
+   int refresh_rate;
 } ALLEGRO_MONITOR_INFO;
 
 enum {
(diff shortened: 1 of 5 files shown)


include/allegro5/monitor.h
```diff
@@ -16,6 +16,7 @@ typedef struct ALLEGRO_MONITOR_INFO
    int y1;
    int x2;
    int y2;
+   int refresh_rate;
```
SiegeLord, line 19: Sadly we can't do this as this breaks backwards compatibility. It'd need to be a new function `al_get_monitor_refresh_rate`.

Also, if we're using it just in the windows backend (and I see you've only implemented it on Windows), it should be a private function for now.

aldrikboy: monitor refresh rate can now be retrieved with al_get_monitor_refresh_rate

SiegeLord:
(just FYI, I'm just waiting on getting the time to figure out how well this can be implemented on Linux/OSX etc)

SiegeLord:
Whoops, I got confused, I didn't mean to merge this quite yet.

PR #1170, opened 2020-09-16
Move more Cocoa operations onto main thread.

### LINKED ISSUE #1168: al_acknowledge_resize() throwing NSInternalInconsistencyException on OSX
I'm using allegro 5.2.6.0 from homebrew, on osx 10.14.6 (mojave).

I'd like to use a resizable window with constraints for a minimum size. I've found that in some situations calling `al_acknowledge_resize()` will cause a crash something like the following:
```
[34145:94913800] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'NSWindow drag regions should only be invalidated on the Main Thread!'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff349f42fd __exceptionPreprocess + 256
	1   libobjc.A.dylib                     0x00007fff5f0c1a17 objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff34a0de59 -[NSException raise] + 9
	3   AppKit                              0x00007fff31fb45ca -[NSWindow(NSWindow_Theme) _postWindowNeedsToResetDragMarginsUnlessPostingDisabled] + 317
	4   AppKit                              0x00007fff31fc5a19 -[NSWindow _commonMinMaxSizeChanged] + 23
	5   liballegro.5.2.dylib                0x0000000100e85f8b -[ALOpenGLView windowWillUseStandardFrame:defaultFrame:] + 223
	6   AppKit                              0x00007fff32297fa9 -[NSWindow _standardFrame] + 482
	7   AppKit                              0x00007fff3236eb3f -[NSWindow isZoomed] + 41
	8   liballegro.5.2.dylib                0x0000000100e86dcc acknowledge_resize_display_win + 259
	9   a.out                               0x0000000100e22ea3 _al_mangled_main + 211
	10  liballegro.5.2.dylib                0x0000000100e89a26 call_user_main + 23
	11  liballegro.5.2.dylib                0x0000000100e89a0f call_user_main + 0
	12  Foundation                          0x00007fff36b975c2 __NSThread__start__ + 1194
	13  libsystem_pthread.dylib             0x00007fff60a832eb _pthread_body + 126
	14  libsystem_pthread.dylib             0x00007fff60a86249 _pthread_start + 66
	15  libsystem_pthread.dylib             0x00007fff60a8240d thread_start + 13
)
libc++abi.dylib: terminating with uncaught exception of type NSException
Abort trap: 6
```

The following is a small reproduction case:

```
#include <allegro5/allegro5.h>
#define INIT_WIDTH  300
#define INIT_HEIGHT 200

int main() {
  al_init();
  al_set_new_display_flags(ALLEGRO_WINDOWED | ALLEGRO_RESIZABLE);

  ALLEGRO_EVENT_QUEUE* queue=al_create_event_queue();
  ALLEGRO_DISPLAY*     disp =al_create_display(INIT_WIDTH, INIT_HEIGHT);
  al_set_window_constraints(disp, INIT_WIDTH,INIT_HEIGHT,0,0);
  al_apply_window_constraints(disp, true);

  al_register_event_source(queue, al_get_display_event_source(disp));
  bool running=true;
  while(running) {
    ALLEGRO_EVENT event;
    al_wait_for_event(queue, &event);
    if (event.type == ALLEGRO_EVENT_DISPLAY_CLOSE) running=false;
    if (event.type == ALLEGRO_EVENT_DISPLAY_RESIZE) al_acknowledge_resize(disp);
    if (al_is_event_queue_empty(queue)) al_flip_display();
  }
  return 0;
}
```
compiled like:
```
gcc resizable.c $(pkg-config allegro-5 allegro_main-5 --libs --cflags)
```

Since the problem happens somewhat inconsistently, I'm not certain that this is minimal. To produce the behavior, I run the program, grab the bottom right corner of the display and wiggle it around for a few seconds. Sometimes it takes a few tries to produce the crash.

Let me know if there's any other information I can provide.

pedro-w:
Also replaced some performSelectorOnMainThread calls with
dispatch_sync calls. This simplifies the implementation.

Fixes #1168

### DIFF
--- a/src/macosx/osxgl.m
+++ b/src/macosx/osxgl.m
@@ -247,32 +247,7 @@ void _al_osx_keyboard_was_installed(BOOL install) {
    [super zoom:sender];
 }
 
-- (void)setStyleMaskSelector:(NSNumber *)mask
-{
-   [self setStyleMask: [mask intValue]];
-}
 @end
-
-/* ALSetWindowFrame:
- * Because we create the window frame on the main thread, we should change
- * it on the main thread as well, otherwise we will get a warning similar
- * to "-[NSLock unlock]: lock  unlocked from thread which did not lock it"
- * everytime we change the window frame, which happens when we resize or
- * move the window.
- */
-//@interface ALSetWindowFrame : NSObject
-//+(void) set_frame : (NSValue *)param;
-//
-//@end
-//@implementation ALSetWindowFrame
-//+(void) set_frame : (NSValue *) param {
-//   NSArray *array = [param pointerValue];
-//   NSRect *rc = [[array objectAtIndex:0] pointerValue];
-//   NSWindow *win = [[array objectAtIndex:1] pointerValue];
-//
-//   [win setFrame:*rc display:YES animate:NO];
-//}
-//@end
 
 /* _al_osx_mouse_was_installed:
  * Called by the mouse driver when the driver is installed or uninstalled.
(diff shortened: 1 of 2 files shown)


pedro-w:
@SiegeLord please can you have a look at this when you've got time? Thanks!

src/macosx/osxgl.m
```diff
@@ -1644,15 +1631,14 @@ static void init_new_vsync(ALLEGRO_DISPLAY_OSX_WIN *dpy)
       }
       [fmt release];
       [view release];
-
+      if (display->flags & ALLEGRO_FULLSCREEN_WINDOW) {
+           // TODO fix main thread
```
SiegeLord, line 1635: Could you clarify what needs to be done here? is this not running on the right thread?

SiegeLord (changes requested):
Thanks for your amazing work, as always. Just one little clarification, if you don't mind.

src/macosx/osxgl.m
pedro-w: It was a note to myself. I fixed it (by moving that whole block up inside the `dispatch_sync`) but didn't delete the note. Sorry about that, I've pushed a new commit to get rid of it.

SiegeLord:
Thanks!

SiegeLord (approved):

Issue #1026, opened 2019-02-24
Joystick lag


arganoid:
The following problems are exhibited in both the example joystick programs, as well as my own code.

I have three controllers. One exhibits no problems. With two of them, there are frequent random delays in input changes (axes and button presses) being recognised – delays of about half a second. I will write more about the fourth controller later. I have tested all the controllers with Windows 10's "Set up USB game controllers", and they all respond immediately with no lag. They all also work fine in Allegro 4. I have not yet found another game which exhibits the delay.

The only controller which operates with no lag 100% of the time is this:
https://www.amazon.co.uk/gp/product/B07BJ273SC/
It's a clone of the Xbox 360 controller and uses the Xbox 360 driver. In the Game Controllers control panel, it is listed as "Controller (Inno GamePad..)". In my test game, the name is reported as "XInput Joystick".

The second controller, which does have the random lag, is this:
https://www.amazon.co.uk/gp/product/B01HTBX9AK/all
The brand name shown on Amazon is CSL. It is listed in Device Manager as "USB Input Device". In the game controllers control panel, and in my test game, it is listed as "USB Joystick".

The third controller, which also has the lag, is:
https://www.amazon.co.uk/gp/product/B00VPMT68E/
The brand name is Kabalo. As with the device above, the names in Device Manager and Game Controllers/test game are "USB Input Device" and "USB Joystick".

I did some investigation and found a connection with joydx_scan. Changing the frequency by which joysticks are re-scanned changes the interval between each incident of lag. I did some profiling and found that IDirectInput8_EnumDevices always takes about 0.4 seconds (regardless of which controller is attached). If I unplug my keyboard, mouse and graphics tablet, that number drops to about 0.2s, and the lag seems noticeably better. But this doesn't explain why the XInput controller shows no lag at all. On a different PC, which doesn't show the lag issue, that number is around 0.03s.

I have confirmed that the time taken by IDirectInput8_EnumDevices is not due to any code in joystick_enum_callback. Each controller returns from that function near the top (either in the 'Filtered out XInput device' block or the 'Device still exists' block.

I have three questions:

1. Does anyone know why this happens and it could be fixed?

2. If not, could there be an option to rescan controllers on demand, rather than automatically?

[...]

PR #1173, opened 2020-09-20
Fixes DirectInput Joystick Lag Issue

NewCreature:
As discussed in issue #1026, this fixes the joystick lag issue caused by frequently calling `IDirectInput8_EnumDevices()`. Instead, we register to receive notifications when devices are connected/disconnected. If we are notified of a device change, we notify the joystick thread to re-enumerate the devices.

### DIFF
--- a/src/win/wjoydxnu.cpp
+++ b/src/win/wjoydxnu.cpp
@@ -251,6 +251,9 @@ static ALLEGRO_JOYSTICK_DIRECTX joydx_joystick[MAX_JOYSTICKS];
 static HANDLE joydx_thread = NULL;
 static CRITICAL_SECTION joydx_thread_cs;
 
+/* whether the DirectInput devices need to be enumerated again */
+static bool need_device_enumeration = false;
+
 /* whether the user should call al_reconfigure_joysticks */
 static bool config_needs_merging = false;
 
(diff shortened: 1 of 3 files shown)


beoran (approved):

beoran:
LGTM. I assume you tested this well with several different joysticks. 

One extra thing we could do is see if the connected device is an XInput one, in which case no rescan is needed.

src/win/wjoydxnu.cpp
```diff
@@ -379,6 +382,14 @@ static void joystick_dinput_acquire(void)
 }
 
 
+/* _al_win_joystick_dinput_trigger_enumeration: [window thread]
+ *  Let joydx_thread_proc() know to reenumerate joysticks.
+ */
+void _al_win_joystick_dinput_trigger_enumeration(void)
+{
+  need_device_enumeration = true;
```
SiegeLord, line 390: Shouldn't this be guarded by joydx_thread_cs?

SiegeLord (changes requested):
Looks great, and appears to work for me. Just one potential race condition I think might be a problem.

SiegeLord:
Merged as https://github.com/liballeg/allegro5/commit/07a7ce7a70b322657259447d6ca0d11205a0efed. Thanks!

PR #1174, opened 2020-09-21
Add support to generate projects with suffix

lorry-lee:
So to create visual studio solutions like: ALLEGRO_PCCLIENT.sln, ALLEGRO_PCSERVER.sln.

Example:
cmake -B /build/pcserver -G "Visual Studio 15 2017" -DIS_SERVER=True -DALLEGRO_PROJECT_SUFFIX:STRING="_PCSERVER"

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,10 @@
 
 # Set the project name.
 # We use C++ in a few cases.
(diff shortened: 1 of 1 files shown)


SiegeLord:
Thanks!

PR #1176, opened 2020-09-21
Windows Joystick - use the correct ANSI/Unicode interface for Direct Input

### LINKED ISSUE #1156: 3rd party controller: crash while enumerating buttons, and corrupt instance name
This is with the latest version from Nuget- 5.2.6.

Connecting this controller https://www.amazon.co.uk/gp/product/B01HTBX9AK/all leads to the following problems:

1. Intermittent assert/crash while getting button names, during "info->button[i].name = ADD_STRING(can->name_button[i], default_name_button[i]);". The assert is ALLEGRO_ASSERT(dest != 0); in add_string. This was happening every time but now is not happening even though I have not recompiled.

2. The instance name returned by this controller is corrupt. In joystick_enum_callback, lpddi->tszInstanceName is "单⁂潊獹楴正넀肊￾￿Þ䵈瘴4쀀". This is not the case in the equivalent Allegro 4 code, where the name is "USB Joystick". Not sure that it's valid Chinese given some of the characters, including what looks like a Korean character at the end.

Attached: screenshots showing contents of lpddi in both Allegro 4 and 5.

Allegro 4
Controller Allegro4

Allegro 5
Controller Allegro5

pedro-w:
This will use ANSI or Unicode, depending on how Allegro itself is configured. (Unicode by default these days)
#Fixes #1156


beoran (approved):
LGTM

SiegeLord:
Looks great, thanks (and thanks Beoran for another pair of eyes on this).

PR #1183, opened 2020-10-21
Update al_draw_multiline_text's `See also` section

abdnh:
Just a trivial change:
The “See also“ section of the `al_draw_multiline_text` function was referencing the same function.


SiegeLord:
Thanks!

PR #1184, opened 2020-10-23
Restore Windows XP compatibility

tehsausage:
Replaces a single function call to a C11 function (`_tcsncpy_s`, which is an alias for `wcsncpy_s`) in wsystem.c with its C99 equivalent, and resolves the following error on Windows XP:

> The procedure entry point wcsncpy_s could not be located in the dynamic link library msvcrt.dll.


beoran:
Will this not cause any regressions on later Windows systems? Have you tested it both on Windows XP and later versions?

tehsausage:
I tested to make sure that it didn't break Allegro on Windows 10. `_tscncpy` is equivalent in behavior to `_tscncpy_s` except that it fills the destination with zero bytes to the end of the buffer.

SiegeLord:
Thanks!

PR #1185, opened 2020-10-23
Fix building with HLSL disabled

tehsausage:
Fixes d3dx9-related build failure when building with `WANT_SHADERS_HLSL` de-selected.

### DIFF
--- a/include/allegro5/allegro_direct3d.h
+++ b/include/allegro5/allegro_direct3d.h
@@ -18,7 +18,7 @@
 #define __al_included_allegro5_allegro_direct3d_h
 
 #include <d3d9.h>
-#if defined ALLEGRO_CFG_SHADER_HLSL && defined __cplusplus
+#if defined ALLEGRO_CFG_D3DX9 && defined __cplusplus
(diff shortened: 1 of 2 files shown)


beoran:
Is the combination of using d3dx9 without HLSL even useful? The dynamic pipeline is likely not to work in this case.

tehsausage:
`ALLEGRO_CFG_D3DX9` is enabled automatically as long as d3dx9 is present on the system. It is required, independent of HLSL support, for DDS textures / DXT compression, I believe.

Having `WANT_DDS` or `WANT_D3DX9` to disable it entirely would be good too, but this is just the minimum to make a `WANT_SHADERS_HLSL=OFF` configuration build consistently.

beoran:
I like your last suggestion, if you have time to work on it, then that would be preferrable. Otherwise, if you only can contribute a minimal patch, then I have no further objections. @SiegeLord will likely review this later as well.

SiegeLord:
Looks good to me, it matches the intent of the macro with the code it actually guards.

Thanks!

PR #1188, opened 2020-10-30
allegro 4.2: Several fixes to compilation using newer DJGPP releases

greg-kennedy:
Allegro 4.2 series has several issues that prevent it from building with DJGPP (MS-DOS) using GCC versions 5.0.0+.  This PR has some commits to resolve them.  More details available in the message for each commit.

Build tested using DJGPP w/ GCC 9.2, latest binutils, Make 3.79.1 in a Windows XP VM.

* Re-add missing obj/djgpp and docs/build folders
* Revert fix.bat to MS-DOS (CRLF) format
* Explicitly set -std=gnu89 and -std=gnu++98 for DJGPP compilation
* Remove `static` qualifier from saved_ds


SiegeLord:
Thanks!

PR #1193, opened 2020-11-17
Raise DirectInput MAX_JOYSTICKS to 32 and DEVICE_BUFFER_SIZE to 128

arganoid:
As discussed at https://www.allegro.cc/forums/thread/618263/1047660


SiegeLord:
Thanks!

PR #1201, opened 2020-12-07
add a new display flag ALLEGRO_OPENGL_CORE_PROFILE

allefant:
This adds a new display flag, just like the existing ALLEGRO_OPENGL_ES_PROFILE but instead of "ES" requests the "core" profile. I do not have a specific use for this feature at the moment but it's something also supported by SDL and I believe is required on MacOS to use any OpenGL past version 2.x. Currently only implemented for X11.

This PR also fixes Allegro to work with an OpenGL core profile (on all platforms). Most of the work was already done when it was made to work with the ES profile in the past - there just was 4 uses of a "glPushClientAttrib" function which does not exist in the core profile - I replaced its use with an explicit query and restore instead.

With this all examples and tests still work like normal, and all the examples I tried now even work when requesting OpenGL 3.2 with core profile. (To test it's enough to add ALLEGRO_OPENGL | ALLEGRO_OPENGL_3_0 | ALLEGRO_OPENGL_ES_PROFILE to the display flags, that will automatically also set the minimum version to 3.2.)

### DIFF
--- a/docs/src/refman/display.txt
+++ b/docs/src/refman/display.txt
@@ -160,6 +160,13 @@ ALLEGRO_OPENGL_ES_PROFILE
     Note: Currently this is only supported by the X11/GLX driver.
     Since: 5.1.13
 
+ALLEGRO_OPENGL_CORE_PROFILE
+:   Used together with ALLEGRO_OPENGL, requests that the OpenGL context
+    uses the OpenGL Core profile. A specific version can be requested
+    with [al_set_new_display_option].
+    Note: Currently this is only supported by the X11/GLX driver.
+    Since: 5.2.7
+
 ALLEGRO_DIRECT3D
 :   Require the driver to do rendering with Direct3D and
     provide a Direct3D device.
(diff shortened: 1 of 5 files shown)


dos1:
Seems like the primitives addon gets broken with it.

allefant:
True, and in ex_prim in a slightly different way depending on if the --shader flag is used or not. I wonder if like with the locking code there already is a working code path which is only being used for OpenGL ES right now. Maybe I'll be able to look at it next weekend.

Anyway, means ALLEGRO_OPENGL_CORE_PROFILE just like the existing ALLEGRO_OPENGL_FORWARD_COMPATIBLE will not support the primitives addon (and possibly other things). I'll clarify the documentation about that. I think it's still useful to remove the glPushClientAttrib calls, that seems to at least make basic bitmap loading and drawing work with a non-backward-compatible GL profile.

SiegeLord:
Looks good to me. I'll take a look at the primitives when I can. I hit the same issue on OSX.

PR #1205, opened 2020-12-12
Update documentation of utf8

rmbeer:


rmbeer:
Fallo manejando github.

pedro-w:
You can link a PR to the issue it's going to fix - see [GitHub docs](https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#about-linked-issues-and-pull-requests)

SiegeLord:
That's not quite the issue with the documentation. The issue is that the returned string is also tied to the info struct, not the other way around. The sentence should be "The returned string is valid while the underlying memory buffer and the info struct are valid."

PR #1207, opened 2020-12-14
Fix compile warning in al_findfirst (src/dos/dfile.c)

greg-kennedy:
This function is returning `NULL` when hitting an out-of-memory error, but it is supposed to return `int`.  The correct "error" return code for this function is `-1`.


SiegeLord:
Thanks!

PR #1208, opened 2020-12-15
Correct initialisation for WaitableTimer

### LINKED ISSUE #1202: Possibly incorrect code in Windows (DX) joystick
Looking at this code:
https://github.com/liballeg/allegro5/blob/1ae91bb949950b0dce5634b42075c81f82bae400/src/win/wjoydxnu.cpp#L993-L1000
I believe it doesn't match the docs for [SetWaitableTimer](https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-setwaitabletimer).

```c
BOOL SetWaitableTimer(
  HANDLE              hTimer,  // joy->waker_event (OK)
  const LARGE_INTEGER *lpDueTime, // &due_time (see below)
  LONG                lPeriod, // true (Not OK)
  PTIMERAPCROUTINE    pfnCompletionRoutine, // NULL (OK)
  LPVOID              lpArgToCompletionRoutine, // NULL (OK)
  BOOL                fResume // false (OK)
);
```
The docs for `lpDueTime` say "The time after which the state of the timer is to be set to signaled, in 100 nanosecond intervals... Positive values indicate absolute time... Negative values indicate relative time. " (I summarised a bit)

I think these args would set the timer to fire once at 15000ns past the epoch and then every 1ms thereafter (assuming `(LONG) true` == `1`)

I assume the intention was a repeating timer every 15ms which would be
```c
LARGE_INTEGER due_time;
due_time.QuadPart = -1; /* Start as soon as possible in the future */
SetWaitableTimer(joy->waker_event,
             &due_time, 
             15, /* repeat every 15ms*/
             NULL, NULL, false);
```
Please can someone check this? Also, not sure what effect it would have in practice, it affects poll-only joysticks but not event-driven ones.

pedro-w:
fixes #1202

### DIFF
--- a/src/win/wjoydxnu.cpp
+++ b/src/win/wjoydxnu.cpp
@@ -991,11 +991,10 @@ static BOOL CALLBACK joystick_enum_callback(LPCDIDEVICEINSTANCE lpddi, LPVOID pv
       }
 
       {
-         LARGE_INTEGER due_time;
-         due_time.HighPart = 0;
-         due_time.LowPart = 150; /* 15 ms (arbitrary) */
+         LARGE_INTEGER due_time = { 0 };
+         due_time.QuadPart = -1; /* Start ASAP... */
          SetWaitableTimer(joy->waker_event,
-                          &due_time, true, /* periodic */
+                          &due_time, 15, /* ... then periodic, every 15ms*/
                           NULL, NULL, false);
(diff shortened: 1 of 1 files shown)


SiegeLord:
Thanks!

PR #1214, opened 2021-01-08
allegro4: xwin.c: fix broken xwin_set_window_name

fleger:
Commit f05270883f2b4382b79ea3248d985938cd9934dc that introduced the use of three windows
broke xwin_set_window_name.

Adopt the same approach as commit 2e64350d03ff958960c3ccf4c2cc87437727afea
and fix this by setting the Class Hints for the managed window instead of
the real window.


SiegeLord:
Seems reasonable. Thanks!

fleger:
Thank you!

PR #1218, opened 2021-01-13
Marek Ratajczak's 64-bit patch

EdgarReynaldo:


SiegeLord:
I wouldn't even begin to know how to compile A4 on windows, could someone try this?

src/unicode.c
```diff
@@ -737,7 +737,7 @@ int uoffset(AL_CONST char *s, int index)
       }
    }
 
-   return (long)s - (long)orig;
+   return (long long)s - (long long)orig;
```
allefant, line 740: The cast is pointless to begin with, just change it to: "return s - orig;" instead

src/file.c
```diff
@@ -2065,7 +2065,7 @@ PACKFILE *pack_fopen_chunk(PACKFILE *f, int pack)
 	       return NULL;
 	    }
 	    _al_sane_strncpy(chunk->normal.passdata, f->normal.passdata, strlen(f->normal.passdata)+1);
-	    chunk->normal.passpos = chunk->normal.passdata + (long)f->normal.passpos - (long)f->normal.passdata;
+	    chunk->normal.passpos = chunk->normal.passdata + (ULONG_PTR)f->normal.passpos - (ULONG_PTR)f->normal.passdata;  //64bit   long
```
allefant, line 2068: ULONG_PTR will not exist in most compiles, did you mean intptr_t?

pedro-w:
This may be a bit late but I wish we could use the `stdint.h` support and say `int64_t`, `int32_t` etc instead of relying on `long long` or `long` or whatever.

EdgarReynaldo:
I will try to build in next day or so - @pedro-w why can't we use stdint.h? don't we use it else where?

pedro-w:
@EdgarReynaldo I didn't mean we *can't*, just that we *haven't* in this patch. It is used quite extensively in the rest of the code. I think it would be better but it's not my place to insist on anything and I don't have the time to propose the changes myself.

ericoporto:
@EdgarReynaldo did you ever figure this out? I am trying to figure out why the "normal" packfile lseek returns a gigantic number in 64-bit systems, breaking the normal refill buffer and the normal get.

EdgarReynaldo:
@ericoporto I forgot all about this. I think it would be best to take Pedro's advice and use stdint.h types instead of long modifiers, which vary per platform. If it can be adjusted, I think it would apply cleanly, as it already does. If you do it I will test building it for you, but I don't really have the understanding to modify it with confidence.

Issue #1215, opened 2021-01-09
high level primitives don't tessellate correctly when scale is very small


goliath-tornado:
When the scale values for high level primitives is very small, e.g. 0.001, then the number of triangles used to generate the triangle fans or strips is incorrect.
I believe the problem is that the scale is applied outside of the sqrt.
Consider the filled ellipse function. It first determines the scale of the output, and then multiplies that by the sqrt of the sum of the two radii.

    num_segments = ALLEGRO_PRIM_QUALITY * scale * sqrtf((rx + ry) / 2.0f);

If the rx and ry values are say 100 each, then the sqrtf function will return 10, and then this would be multiplied by the primitive quality and scale. If scale was 1, i.e. we are drawing a single unit per pixel, then num_segments would generate 100 triangles.

However, if the scale value was 0.001, and rx and ry were 100,000 - which would result in the same size circle on the screen, the calculation would become ~0.3. This is probably not enough triangles.
If the code instead used:

    num_segments = ALLEGRO_PRIM_QUALITY * sqrtf(scale * (rx + ry) / 2.0f);

Then the sqrt would be calculated on 0.001 * 100000 = 100. This results in the same 100 triangles. Therefore the number of triangles drawn is dependant on the size on the resultant circle and not on the scale used.

The above assumes that al_scale_transform is being used to scale everything being drawn.

[](url)

goliath-tornado:
I think these are all the places that this needs to be fixed.

```
diff --git a/addons/primitives/high_primitives.c b/addons/primitives/high_primitives.c
index 91742a4b1..bb5630d08 100644
--- a/addons/primitives/high_primitives.c
+++ b/addons/primitives/high_primitives.c
@@ -520,7 +520,7 @@ void al_draw_pieslice(float cx, float cy, float r, float start_theta,
    }
    
    if (thickness <= 0) {
-      num_segments = fabs(delta_theta / (2 * ALLEGRO_PI) * ALLEGRO_PRIM_QUALITY * scale * sqrtf(r));
+      num_segments = fabs(delta_theta / (2 * ALLEGRO_PI) * ALLEGRO_PRIM_QUALITY * sqrtf(scale * r));
 
       if (num_segments < 2)
          num_segments = 2;
@@ -566,7 +566,7 @@ void al_draw_pieslice(float cx, float cy, float r, float start_theta,
          vertex_cache[0].x = cx + (r - thickness / 2) * cosf(central_start_angle);
[...]

PR #1220, opened 2021-01-26
high level primitives don't tessellate correctly when scale is very small 

goliath-tornado:
This should fully resolve issue #1215


SiegeLord:
Thanks!

PR #1221, opened 2021-01-31
Create SECURITY.md

CarlosAndresRamirez:
The Allegro library is being used in academic environments and production environments for games (commercial and free) and other tools. A policy to assure some degree of security and quality must be put in place.

I am a security engineer (and Allegro user for 17+ years). I am willing to contribute in this process.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,17 @@
+# Security Policy
+
+## Report a security issue
+
+The Allegro project team welcomes security reports and is committed to providing prompt attention to security issues.
+Security issues should be reported privately via [email]
+
+| Version | Supported          |
+| ------- | ------------------ |
+| >= 5.0.x| :white_check_mark: |
+| < 5.0   | :x:                |
+
+## Security advisories
+
+Remediation of security vulnerabilities is prioritized by the project team based on the overall impact.
+The project team is committed to transparency in the disclosure process.
+The Allegro team announces security issues via Github Release notes, Allegro Release notes, as well as the Allegro website on a best-effort basis.


CarlosAndresRamirez (commented):
Please, edit email address for contact.

SiegeLord:
This seems to be alright. What email is typically used? A developer's email? Some sort of private mail list?

pedro-w:
Relevant info is here:
https://docs.github.com/en/github/managing-security-vulnerabilities/managing-security-vulnerabilities-in-your-project
I haven't read it fully yet but AFAICS someone with admin access would create a (private) security advisory and then invite the person reporting the issue to collaborate on it. Then when it's fixed the advisory can be made public. So we need some way of the person who has found the vulnerability to make initial contact. This could be someone's email or a special address set up for that purpose. I don't think github helps with this.
@CarlosAndresRamirez  is that the way it works in your experience?

CarlosAndresRamirez:
Yes,  @pedro-w is right. Usually a private mailing list is used (devs only).

The security policy has to be realistic, what I provided is just an initial template - feel free to adjust to what you think is doable. While most of the the projects that integrate Allegro are small projects, from time to time some popular/commercial games (Liquid War and Factorio come to my mind) do use the library. It is important to provide some visibility on what can be expected, or at least provide the assurance that security reports get looked into.

beoran:
Too bad GitHub doesn't support private issues like GitLab does, we could have used those. But yes, a simple security policy would be helpful.

SiegeLord:
Ok, we added something like this in b8076fea91659f6b9550b5fdf156620246e478bd. I ended up creating a private mailing list. I just added myself and Elias to it as members, but feel free to ping me and request membership. I fully expect this to never receive a single non-spam email though, hah!

CarlosAndresRamirez:
Thanks @SiegeLord 
This will give some assurance regarding quality and best practices of the project. Hopefully there will not be many reports in the future.

I'll be happy to support any security reviews/reports/fixes.

pedro-w:
@CarlosAndresRamirez what do you think of https://securitytxt.org/ ?

CarlosAndresRamirez:
@pedro-w 
Yes, the security.txt file from https://securitytxt.org/ is a good approach.

pedro-w:
I suppose the question was really : do you think we should create a `/.well-known/security.txt` on github and/or liballeg.org? It's additional to what's already been done on `SECURITY.md`, so more work to maintain. Is there one approach that's becoming the accepted standard?

CarlosAndresRamirez:
Having the SECURITY.md included in the sources should be enough; it is what most projects do. Adding it to /.well-known/security.txt and following the https://securitytxt.org/ standards would be a plus, but yes, as you said, it would be an additional file to maintain.

PR #1223, opened 2021-02-07
enable logs and fix al_get_next_event for emscripten

elias-pschernig:
For getting logs, with emscripten stdout appears in the logs so I simply made Allegro's logging functions write to stdout instead of a file when compiling with emcc.

With emscripten you normally cannot directly use the main() function once converted to javascript because it never returns, so the website would just hang. Instead you create a function which handles one frame and then call it repeatedly from a javascript timer. That's how we previously got some Allegro programs to work by compiling against the SDL2 port and then modifying code so it would not use main().

However there is an alternative way called ASYNCIFY: https://emscripten.org/docs/porting/asyncify.html
It transforms the C code into javascript code which can pause and resume any function, and specifically does so whenever a call to emscripten_sleep() is made. I added a small hack to al_wait_next_event to instead of waiting forever (which would immediately deadlock all of the Allegro examples) repeatedly call emscripten_sleep(1). That way most of the unmodified Allegro examples seem to work.

The existing heartbeat() used for SDL2 does not work for this because al_wait_next_event() with the SDL driver relies on some timer to wake up the unconditional wait (our SDL2 port when not using a timer is currently broken as we cannot receive any SDL events until a timer is triggered).

So, basically this adds another hack on top of the SDL2 hack, but it works rather well: I got 2/3 demos working and after compiling freetype, ttf and ogg/vorbis to javascript I believe over 100/137 examples: http://allegro5.org/examples

The speed demo uses some of the more unusual event functions instead of al_get_next_event() and should be easy to fix but like with the normal SDL2 driver I don't know if there's a good method that will work in all possible cases.

Audio is all messed up, haven't investigated that yet.

If someone wants to reproduce, this is the script I used to configure Allegro to compile the examples:
```bash
# we need to call emsdk_env.sh to set up all the emscripten paths
. /home/elias/stuff/emscripten/emsdk/emsdk_env.sh
mkdir build_emscripten
cd build_emscripten

USE_FLAGS=(
    -s USE_FREETYPE=1
    -s USE_VORBIS=1
    -s USE_OGG=1
    -s USE_LIBJPEG=1
    -s USE_SDL=2
    -s USE_LIBPNG=1
    -s FULL_ES2=1
    -s ASYNCIFY
    -s TOTAL_MEMORY=2147418112
    -O3
    )

emscripten=(
    emcmake cmake ../../allegro/git
    -D CMAKE_BUILD_TYPE=Debug
    -D ALLEGRO_SDL=ON
    -D SHARED=OFF
    -D WANT_MONOLITH=ON
    -D WANT_ALLOW_SSE=OFF
    -D WANT_DOCS=OFF
    -D WANT_TESTS=OFF
    -D WANT_OPENAL=OFF
     # not sure why these are not found automatically with emcmake, so
     # we use a  hack - we manually specify the cache path, so cmake
     # will fail on first run in the compile test but populate the cache
     # (from the -s options) and on second run it all works
    -D SDL2_INCLUDE_DIR=$EM_CACHE/include
    -D PNG_PNG_INCLUDE_DIR=$EM_CACHE/include
    -D PNG_LIBRARY=$EM_CACHE/wasm/libpng.a
    -D JPEG_INCLUDE_DIR=$EM_CACHE/include
    -D JPEG_LIBRARY=$EM_CACHE/wasm/libjpeg.a
    -D FREETYPE_INCLUDE_DIRS=$EM_CACHE/include
    -D FREETYPE_LIBRARY=$EM_CACHE/wasm/libfreetype.a
    -D VORBIS_INCLUDE_DIR=$EM_CACHE/include/vorbis
    -D VORBIS_LIBRARY=$EM_CACHE/wasm/libvorbis.a
    -D VORBISFILE_LIBRARY=$EM_CACHE/wasm/libvorbis.a
    -D OGG_INCLUDE_DIR=$EM_CACHE/include
    -D OGG_LIBRARY=$EM_CACHE/wasm/libogg.a
    -D CMAKE_C_FLAGS="${USE_FLAGS[@]}"
    -D CMAKE_CXX_FLAGS="${USE_FLAGS[@]}"
    -D CMAKE_EXE_LINKER_FLAGS="${USE_FLAGS[@]} --preload-file data"
     # does not actually work (but can edit the toolchain file)
    -D CMAKE_EXECUTABLE_SUFFIX=".html"
   )

"${emscripten[@]}" # will fail if ports are not downloaded yet
"${emscripten[@]}" # second time it should succeed

# set a link to the data dir, that way emcc can pack it up for each example
( cd examples; ln -s ../../../allegro/git/examples/data )
# create empty data dir in case of the speed demo
mkdir demos/speed/data
touch demos/speed/data/nothing.txt

emmake make
```

### DIFF
--- a/src/debug.c
+++ b/src/debug.c
@@ -303,14 +303,11 @@ void _al_trace_suffix(const char *msg, ...)
       return;
    }
 
-#ifdef __EMSCRIPTEN__
-   printf("%s", static_trace_buffer);
-#endif
 #ifdef ALLEGRO_ANDROID
    (void)__android_log_print(ANDROID_LOG_INFO, "allegro", "%s",
       static_trace_buffer);
 #endif
-#ifdef ALLEGRO_IPHONE
+#if defined(ALLEGRO_IPHONE) || defined(__EMSCRIPTEN__)
    fprintf(stderr, "%s", static_trace_buffer);
    fflush(stderr);
 #endif
--- a/src/debug.c
+++ b/src/debug.c
@@ -176,7 +176,7 @@ static void open_trace_file(void)
       if (s)
          trace_info.trace_file = fopen(s, "w");
       else
-#if defined(ALLEGRO_IPHONE) || defined(ALLEGRO_ANDROID)
+#if defined(ALLEGRO_IPHONE) || defined(ALLEGRO_ANDROID) || defined(__EMSCRIPTEN__)
          /* iPhone and Android don't like us writing files, so we'll be doing
           * something else there by default. */
          trace_info.trace_file = NULL;
@@ -303,6 +303,9 @@ void _al_trace_suffix(const char *msg, ...)
       return;
    }
 
+#ifdef __EMSCRIPTEN__
+   printf("%s", static_trace_buffer);
+#endif
 #ifdef ALLEGRO_ANDROID
    (void)__android_log_print(ANDROID_LOG_INFO, "allegro", "%s",
       static_trace_buffer);
--- a/src/events.c
+++ b/src/events.c
@@ -23,9 +23,6 @@
 
 
 #include <string.h>
-#ifdef __EMSCRIPTEN__
-#include <emscripten.h>
-#endif
 
 #include "allegro5/allegro.h"
 #include "allegro5/internal/aintern.h"
@@ -405,16 +402,14 @@ void al_wait_for_event(ALLEGRO_EVENT_QUEUE *queue, ALLEGRO_EVENT *ret_event)
 
    _al_mutex_lock(&queue->mutex);
    {
-      #ifdef __EMSCRIPTEN__
       while (is_event_queue_empty(queue)) {
-         emscripten_sleep(1);
+         #ifdef ALLEGRO_NO_THREADS
+         al_rest(1);
          heartbeat();
-      }
-      #else
-      while (is_event_queue_empty(queue)) {
+         #else
          _al_cond_wait(&queue->cond, &queue->mutex);
+         #endif
       }
-      #endif
 
       if (ret_event) {
          next_event = get_next_event_if_any(queue, true);
--- a/src/events.c
+++ b/src/events.c
@@ -403,7 +403,7 @@ void al_wait_for_event(ALLEGRO_EVENT_QUEUE *queue, ALLEGRO_EVENT *ret_event)
    _al_mutex_lock(&queue->mutex);
    {
       while (is_event_queue_empty(queue)) {
-         #ifdef ALLEGRO_NO_THREADS
+         #ifdef ALLEGRO_WAIT_EVENT_SLEEP
          al_rest(1);
          heartbeat();
          #else
--- a/src/events.c
+++ b/src/events.c
@@ -404,7 +404,7 @@ void al_wait_for_event(ALLEGRO_EVENT_QUEUE *queue, ALLEGRO_EVENT *ret_event)
    {
       while (is_event_queue_empty(queue)) {
          #ifdef ALLEGRO_WAIT_EVENT_SLEEP
-         al_rest(1);
+         al_rest(0.001);
          heartbeat();
          #else
          _al_cond_wait(&queue->cond, &queue->mutex);
--- a/src/events.c
+++ b/src/events.c
@@ -23,6 +23,9 @@
 
 
 #include <string.h>
+#ifdef __EMSCRIPTEN__
+#include <emscripten.h>
+#endif
 
 #include "allegro5/allegro.h"
 #include "allegro5/internal/aintern.h"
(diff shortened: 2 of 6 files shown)


src/debug.c
```diff
@@ -303,6 +303,9 @@ void _al_trace_suffix(const char *msg, ...)
       return;
    }
 
+#ifdef __EMSCRIPTEN__
```
SiegeLord, line 306: Does emscripten support stderr? Could we use the ALLEGRO_IPHONE code for this?

src/events.c
```diff
@@ -402,9 +405,16 @@ void al_wait_for_event(ALLEGRO_EVENT_QUEUE *queue, ALLEGRO_EVENT *ret_event)
 
    _al_mutex_lock(&queue->mutex);
    {
+      #ifdef __EMSCRIPTEN__
+      while (is_event_queue_empty(queue)) {
```
SiegeLord, line 409: I don't suppose al_wait_cond_until would just work here (no idea if it's implemented)? Do condvars even work on emscripten?

SiegeLord (commented):
Very neat!

src/debug.c
elias-pschernig: Yes, stderr goes to the javascript console only, while stdout goes to the that fancy text widget in the HTML pages. I guess the Allegro logs make more sense in the javascript console, allowing the HTML widget to be used for something more user centric. Changing it.

src/events.c
elias-pschernig: It depends. You can compile emscripten with pthreads enabled: https://emscripten.org/docs/porting/pthreads.html

I just couldn't get it to work myself, but I can try again. Without pthreads none of the threading/locking related functions do anything at all so al_get_next_event() just instantly deadlocks the browser in that case.

elias-pschernig: Hm, I tried with -pthread but then I hit: https://github.com/emscripten-ports/SDL2/issues/110

I'm using the SDL2 that comes prepackaged with emscripten right now so not really sure how to apply those fixes to try it. I'll see if I can fix the audio issue before I attempt that.

elias-pschernig:
The audio issues were caused by a bug in the SDL driver and are all fixed, here's e.g. ex_haiku in a web browser: http://allegro5.org/examples/examples/ex_haiku.html

I don't really have a better idea at this moment for the lack of threads - once -pthread is supported in emscripten-SDL2 we can get rid of the loop-with-sleep method introduced by this PR but for now it's a hack that seems to work.

SiegeLord:
Ok, so lets ignore threads. Does `al_rest` work instead of a direct call to `emscripten_sleep`? I checked SDL's source, and it seems `SDL_Delay` will call it, and `al_rest` uses `SDL_Delay`.

Overall, I'm just trying to get at some generic solution here. What seems to be possible here is a compilation switch to go from condvar-based wait to `al_rest`-based wait.

elias-pschernig:
Yes, it can be enabled programmatically in SDL2: https://wiki.libsdl.org/SDL_HINT_EMSCRIPTEN_ASYNCIFY

I think it means we'd need another switch to enable that hint, something like EMSCRIPTEN_ASYNCIFY. Not sure there is much use of the rest based loop outside of emscripten so that flag could enable both, or else a separate flag to enable that.

elias-pschernig:
Didn't look closely enough - EMSCRIPTEN_ASYNCIFY is on by default when compiling with async emscripten so only one option it needed to use the sleep loop. I name it ALLEGRO_NO_THREADS for now, but maybe that's a bit too general and something like ALLEGRO_WAIT_EVENT_SLEEP is a better name.

beoran:
ALLEGRO_WAIT_EVENT_SLEEP sound better, yes. Doesn't the sleep cause some sluggishness in event handling, though?

elias-pschernig:
> ALLEGRO_WAIT_EVENT_SLEEP sound better, yes. Doesn't the sleep cause some sluggishness in event handling, though?

It shouldn't, once there is an event we break out of the loop and the game can handle all the queued events. With javascript the web browser already calls the game code as often as required (60 times per second probably). The game is not supposed to use any blocking calls at all. EMSCRIPTEN_ASYNCIFY is just a bit of a hack provided by the emscripten developers to make programs like the Allegro examples work who do use waits regardless. The better way to port a game to emscripten would be to just not use al_wait_for_event() at all but it would add a bit of porting work (and we could not run any of our examples).

elias-pschernig:
Renamed the #define now. Not sure about documentation, I could add a README_emscripten with explanations. Or a README_SDL with a section on how to make it work with emscripten. But then I feel SDL/emscripten is not really a fully supported port so maybe just provide the above bash script in misc/. Or else just have it in the wiki.

SiegeLord:
I guess we don't have a README_sdl, do we, haha. We should certainly document it somewhere, but it doesn't matter to me where (it's certainly the case that README_xxx is the most durable location though).

elias-pschernig:
I added a README_sdl.txt (btw. should we convert all those README*.txt to .md and include as a chapter in the online documentation?). And I put a section on emscripten in there.

beoran (approved):

SiegeLord:
Very awesome, I merged it as https://github.com/liballeg/allegro5/commit/089b9ac3a8584d052a69a7ce2b0aed71e0f9cec4 with some changes to the readme based on our discussion on IRC.

PR #1224, opened 2021-02-14
fix bug in SDL voice driver

elias-pschernig:
We pass SDL_AUDIO_ALLOW_FORMAT_CHANGE to SDL_OpenAudioDevice (which is a good idea as only one specific format may be available on the target device) but then completely ignore format changes leading to broken audio. This PR fixes the issue.

(This was the issue that caused audio in emscripten to be broken which uses the SDL driver.)


beoran:
LGTM :+1:

PR #1225, opened 2021-02-23
add default files to some audio examples

elias-pschernig:
This is a very minor PR. Some of our audio examples won't work without providing a file on the command line, even though most of them use a default file in that case. I added default files to 5 more:

- ex_acodec_multi
- ex_audio_simple
- ex_kcm_direct
- ex_mixer_chain
- ex_stream_file


SiegeLord:
Nice! Seems good on the consistency and the usability front.

PR #1226, opened 2021-02-28
Shooter

elias-pschernig:
This adds the old A4 demo to A5. The first commit simply copies the A4 folder. The second commit makes it compile as A5 (using the same a4_aux.c from the skater demo) and adds it to the build system.


SiegeLord:
This needed a few small fixes (05248497a19e6217746f5579645e57bfd7ee383e). We should probably fix up the text etc to make it more up to date. And maybe update the API to be more A5-idiomatic.

PR #1228, opened 2021-03-12
DirectX MSAA fix for flicker on startup in windowed and fullscreen mode

EdgarReynaldo:


EdgarReynaldo:
NOTE : This works with my NVIDIA GPU and DirectX MSAA in windowed mode, but fullscreen doesn't enable MSAA, and the onboard intel fails in both windowed and fullscreen modes. So basically it's not ready yet.

EdgarReynaldo:
ALLEGRO_FULLSCREEN_WINDOW works to enable MSAA but again, only on my NVIDIA GPU

SiegeLord:
I looked into this a bit. First, I've no idea what this PR is trying to accomplish, and it doesn't work for me:

- For `FULLSCREEN_WINDOWED`, it removes the flickering but it doesn't actually create a multisampled display
- For `FULLSCREEN` it increases flickering like 10x

Then, I looked at what was happening, and noticed a bug in `_al_d3d_generate_display_format_list`, in that it was enumerating the multisampling quality levels, but populating a bogus quality in the display options (which was then passed to D3D, and causing it to fail). Here's the fix which appears to help in my case (could you test it?):

```diff
diff --git a/src/win/d3d_display_formats.cpp b/src/win/d3d_display_formats.cpp
index 14f749a3c..9081a8b7a 100644
--- a/src/win/d3d_display_formats.cpp
+++ b/src/win/d3d_display_formats.cpp
@@ -134,7 +134,7 @@ void _al_d3d_generate_display_format_list(void)
                if (k > 1) {
                   eds->settings[ALLEGRO_SAMPLE_BUFFERS] = 1;
                   // TODO: Is it ok to use the quality level here?
-                  eds->settings[ALLEGRO_SAMPLES] = k;
+                  eds->settings[ALLEGRO_SAMPLES] = k - 1;
                }
             }
          }
```

But stepping back a bit... D3D9 quality levels do not map to number of samples of taken, so the entire API is completely broken from Allegro's point of view. Allegro user requests a certain number of samples, but the code selects the quality close to that number of samples; this won't work in general, however, as quality has no fixed relation to the number of samples. E.g. here's a table I found for some old AMD card:

| Quality | Mode        |
|---------|-------------|
| 0       | 2x MSAA     |
| 1       | 2f4x EQAA   |
| 2       | 4x MSAA     |
| 3       | 2f8x EQAA   |
| 4       | 4f8x EQAA   |
| 5       | 4f16x EQAA  |
| 6       | 8x MSAA     |
| 7       | 8f16x EQAA  |

So... yeah. Assuming the fix above works, we can fix the flickering, but the selected multi-sample quality will be quite random. We'd need to create a new API for MSAA, to fully support this.

SiegeLord:
I committed a version of my change in https://github.com/liballeg/allegro5/commit/b1f96f7bf0a6221b0e6110358971f99bba96f237. I don't think this PR is necessary anymore.

PR #1230, opened 2021-03-20
Regular exit added

BillKek:
Regular exit added. ( Cannot exit from App without magic keystroke "Esc" )
Tested OK on Windows with MSYS2/Mingw64.


SiegeLord:
Thanks!

PR #1238, opened 2021-04-10
List audio devices

aldrikboy:
Add audio device enumeration to the dsound, sdl, alsa and pulseaudio backends (not implemented: openal, kcm, aqueue, oss).

Example usage:

	int count = al_get_audio_device_count(); // returns -1 for unsupported backends
	printf("audio device count: %d\n", count);
	
	for (int i = 0; i < count; i++)
	{
		ALLEGRO_AUDIO_DEVICE* device = al_get_audio_device(i);
		
		printf("audio device: %s\n", al_get_audio_device_name(device));
	}

the device identifier is also stored so audio device selection could easily be implemented.

### DIFF
--- a/addons/audio/allegro5/allegro_audio.h
+++ b/addons/audio/allegro5/allegro_audio.h
@@ -359,7 +359,7 @@ ALLEGRO_KCM_AUDIO_FUNC(size_t, al_get_audio_depth_size, (ALLEGRO_AUDIO_DEPTH con
 ALLEGRO_KCM_AUDIO_FUNC(void, al_fill_silence, (void *buf, unsigned int samples,
       ALLEGRO_AUDIO_DEPTH depth, ALLEGRO_CHANNEL_CONF chan_conf));
 
-ALLEGRO_KCM_AUDIO_FUNC(int, al_get_audio_device_count, (void));
+ALLEGRO_KCM_AUDIO_FUNC(int, al_get_num_audio_devices, (void));
 ALLEGRO_KCM_AUDIO_FUNC(ALLEGRO_AUDIO_DEVICE *, al_get_audio_device, (int index));
 ALLEGRO_KCM_AUDIO_FUNC(char *, al_get_audio_device_name, (ALLEGRO_AUDIO_DEVICE * device));
 
--- a/addons/audio/allegro5/allegro_audio.h
+++ b/addons/audio/allegro5/allegro_audio.h
@@ -173,6 +173,10 @@ typedef struct ALLEGRO_MIXER ALLEGRO_MIXER;
  */
 typedef struct ALLEGRO_VOICE ALLEGRO_VOICE;
 
+/* Type: ALLEGRO_AUDIO_DEVICE
+ */
+typedef struct ALLEGRO_AUDIO_DEVICE ALLEGRO_AUDIO_DEVICE;
+
 
 #if defined(ALLEGRO_UNSTABLE) || defined(ALLEGRO_INTERNAL_UNSTABLE) || defined(ALLEGRO_KCM_AUDIO_SRC)
 /* Type: ALLEGRO_AUDIO_RECORDER
@@ -354,6 +358,10 @@ ALLEGRO_KCM_AUDIO_FUNC(size_t, al_get_audio_depth_size, (ALLEGRO_AUDIO_DEPTH con
 
 ALLEGRO_KCM_AUDIO_FUNC(void, al_fill_silence, (void *buf, unsigned int samples,
       ALLEGRO_AUDIO_DEPTH depth, ALLEGRO_CHANNEL_CONF chan_conf));
+
+ALLEGRO_KCM_AUDIO_FUNC(int, al_get_audio_device_count, (void));
+ALLEGRO_KCM_AUDIO_FUNC(ALLEGRO_AUDIO_DEVICE *, al_get_audio_device, (int index));
+ALLEGRO_KCM_AUDIO_FUNC(char *, al_get_audio_device_name, (ALLEGRO_AUDIO_DEVICE * device));
 
 /* Simple audio layer */
 ALLEGRO_KCM_AUDIO_FUNC(bool, al_reserve_samples, (int reserve_samples));
--- a/addons/audio/alsa.c
+++ b/addons/audio/alsa.c
@@ -878,14 +878,14 @@ static void alsa_deallocate_recorder(ALLEGRO_AUDIO_RECORDER *r)
    snd_pcm_close(data->capture_handle);
 }
 
-void _device_list_dtor(void* value, void* userdata)
+static void _device_list_dtor(void* value, void* userdata)
 {
    ALLEGRO_AUDIO_DEVICE* device = (ALLEGRO_AUDIO_DEVICE*)value;
    al_free(device->name);
    al_free(device->identifier);
 }
 
-_AL_LIST* alsa_get_devices()
+static _AL_LIST* alsa_get_devices()
 {
    if (!device_list) {
       device_list = _al_list_create();
--- a/addons/audio/alsa.c
+++ b/addons/audio/alsa.c
@@ -41,6 +41,7 @@ do {                                                                  \
 static snd_output_t *snd_output = NULL;
 static char *default_device = "default";
 static char *alsa_device = NULL;
+static _AL_LIST* device_list;
 
 // TODO: Setting this to 256 causes (extreme, about than 10 seconds)
 // lag if the alsa device is really pulseaudio.
@@ -140,6 +141,8 @@ Error:
    other processes to use the device */
 static void alsa_close(void)
 {
+   _al_list_destroy(device_list);
+
    if (alsa_device != default_device)
       al_free(alsa_device);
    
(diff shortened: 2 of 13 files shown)


addons/audio/allegro5/allegro_audio.h
```diff
@@ -355,6 +359,10 @@ ALLEGRO_KCM_AUDIO_FUNC(size_t, al_get_audio_depth_size, (ALLEGRO_AUDIO_DEPTH con
 ALLEGRO_KCM_AUDIO_FUNC(void, al_fill_silence, (void *buf, unsigned int samples,
       ALLEGRO_AUDIO_DEPTH depth, ALLEGRO_CHANNEL_CONF chan_conf));
 
+ALLEGRO_KCM_AUDIO_FUNC(int, al_get_audio_device_count, (void));
```
pedro-w, line 362: Bikeshedding I know, but should this be `al_get_num_audio_devices` to be consistent with `al_get_num_video_adapters`, `al_get_num_joysticks` and `al_get_num_display_modes` ?
OTOH, `al_get_channel_count` but that's the only one, and there isn't an `al_get_channel` so it's not quite the same.

src/misc/list.c
```diff
@@ -593,7 +593,7 @@ _AL_LIST_ITEM* _al_list_at(_AL_LIST* list, size_t index)
 
       _AL_LIST_ITEM* item = list->root->prev;
 
-      index = list->size - index;
+      index = list->size - index - 1;
```
pedro-w, line 596: Good spot, I wonder how this has not been fixed before!

addons/audio/allegro5/allegro_audio.h
aldrikboy: I agree, should've checked for similar functions

pedro-w:
This is great and when it's merged I will (try to) implement it for MacOS.

Please could you also update the documentation?

aldrikboy:
I am unable to get the docs building on my machine, could you please check if my changes are displaying correctly in the docs?

pedro-w:
Almost! The `make_doc` utility needs a special format comment in the source code so it can extract the parameters and link to source. For a function it just needs to be `Function: ` followed by the name of the function. See for example
https://github.com/liballeg/allegro5/blob/e1a0e7b9195c1cfe644520e5664efeba4d696a65/addons/audio/audio.c#L125-L129
Without that you just get
Screenshot 2021-04-11 at 13 51 54

and the source link goes to the top of the docs page, not the source.

pedro-w:
Nailed it!
@SiegeLord  it would be great if this could be merged

addons/audio/alsa.c
```diff
@@ -875,6 +878,94 @@ static void alsa_deallocate_recorder(ALLEGRO_AUDIO_RECORDER *r)
...
+
+      int rcard = -1;
+      while(snd_card_next(&rcard) == 0) {
+         if (rcard < 0) return device_list;
+
+         snd_ctl_t *handle;
+         char str[64];
+
+         sprintf(str, "hw:%i", rcard);
+         if (snd_ctl_open(&handle, str, 0) < 0) {
```
SiegeLord, line 903: Why do we bail out here instead of doing a continue?

addons/audio/alsa.c
```diff
@@ -875,6 +878,94 @@ static void alsa_deallocate_recorder(ALLEGRO_AUDIO_RECORDER *r)
...
+            return device_list;
```
SiegeLord, line 904: For these early returns etc, it's useful to add an ALLEGRO_DEBUG("snd_ctl_open failed %d with error code %d\n", rcard, error_code);

addons/audio/alsa.c
```diff
@@ -875,6 +878,94 @@ static void alsa_deallocate_recorder(ALLEGRO_AUDIO_RECORDER *r)
...
+         }
+
+         snd_ctl_card_info_t *card_info;
+         snd_ctl_card_info_alloca(&card_info);
+         if (snd_ctl_card_info(handle, card_info) < 0) {
+            continue;
```
SiegeLord, line 910: Log here too.

addons/audio/alsa.c
```diff
@@ -875,6 +878,94 @@ static void alsa_deallocate_recorder(ALLEGRO_AUDIO_RECORDER *r)
...
+         }
+
+         snd_ctl_card_info_get_name(card_info);
+         snd_ctl_card_info_get_id(card_info);
+
+         int dev_num = -1;
+
+         while(1) {
+            if (snd_ctl_pcm_next_device(handle, &dev_num) < 0) {
+               break;
```
SiegeLord, line 920: Log here too.

addons/audio/alsa.c
```diff
@@ -875,6 +878,94 @@ static void alsa_deallocate_recorder(ALLEGRO_AUDIO_RECORDER *r)
...
+            }
+
+            if (dev_num < 0) break;
```
SiegeLord, line 923: Log here too.

addons/audio/alsa.c
```diff
@@ -875,6 +878,94 @@ static void alsa_deallocate_recorder(ALLEGRO_AUDIO_RECORDER *r)
...
+               
+            char* identifier = 0;
+            char* name = 0;
+            while(*n != NULL)
+            {
+               identifier = snd_device_name_get_hint(*n, "NAME");
+               name = snd_device_name_get_hint(*n, "DESC");
+                  
+               char* ioid = snd_device_name_get_hint(*n, "IOID");
+               if (ioid == NULL || strcmp(ioid, "Output") == 0) break;
```
SiegeLord, line 938: Log here too.

addons/audio/alsa.c
```diff
@@ -875,6 +878,94 @@ static void alsa_deallocate_recorder(ALLEGRO_AUDIO_RECORDER *r)
...
+               n++;
+            }
+
+            if (!identifier || !name) continue;
```
SiegeLord, line 942: Log here too.

addons/audio/audio.c
```diff
@@ -196,6 +196,50 @@ static ALLEGRO_AUDIO_DRIVER_ENUM get_config_audio_driver(void)
...
+   if (_al_kcm_driver) {
+      if (_al_kcm_driver->get_devices) {
+         _AL_LIST* audio_devices = _al_kcm_driver->get_devices();
+
+         if (index >= 0 && index < _al_list_size(audio_devices)) {
+            return _al_list_item_data(_al_list_at(audio_devices, index));
+         }
+      }
+      else {
+         return 0;
```
SiegeLord, line 229: NULL is preferable to 0.

addons/audio/audio.c
```diff
@@ -196,6 +196,50 @@ static ALLEGRO_AUDIO_DRIVER_ENUM get_config_audio_driver(void)
...
+   }
+
+   return 0;
+}
+
+/* Function: al_get_audio_device_name
+ */
+char* al_get_audio_device_name(ALLEGRO_AUDIO_DEVICE * device)
+{
+   return device ? device->name : 0;
```
SiegeLord, line 240: NULL is preferable to 0.

addons/audio/audio.c
```diff
@@ -196,6 +196,50 @@ static ALLEGRO_AUDIO_DRIVER_ENUM get_config_audio_driver(void)
...
+         return 0;
+      }
+   }
+
+   return 0;
+}
+
+/* Function: al_get_audio_device_name
+ */
+char* al_get_audio_device_name(ALLEGRO_AUDIO_DEVICE * device)
```
SiegeLord, line 238: const char* and const ALLEGRO_AUDIO_DEVICE*, probably.

addons/audio/audio.c
```diff
@@ -196,6 +196,50 @@ static ALLEGRO_AUDIO_DRIVER_ENUM get_config_audio_driver(void)
...
+         return -1;
+      }
+   }
+
+   return 0;
+}
+
+/* Function: al_get_audio_device
+ */
+ALLEGRO_AUDIO_DEVICE* al_get_audio_device(int index)
```
SiegeLord, line 218: const ALLEGRO_AUDIO_DEVICE*, I assume?

addons/audio/alsa.c
```diff
@@ -875,6 +878,94 @@ static void alsa_deallocate_recorder(ALLEGRO_AUDIO_RECORDER *r)
...
+            int success = snd_device_name_hint(dev_num, "pcm", &hints);
+            if (success < 0) continue;
+            char**n = (char**)hints;
+               
+            char* identifier = 0;
+            char* name = 0;
+            while(*n != NULL)
+            {
+               identifier = snd_device_name_get_hint(*n, "NAME");
+               name = snd_device_name_get_hint(*n, "DESC");
```
SiegeLord, line 935: Why is DESC treated as a name, rather than NAME?

addons/audio/alsa.c
SiegeLord, line 938: Why are we dropping non-Output devices, we do support recording, after all.

addons/audio/pulseaudio.c
```diff
@@ -73,16 +75,40 @@ static unsigned int get_buffer_size(const ALLEGRO_CONFIG *config)
...
+      device_list = _al_list_create();
+   }
+
    pa_sink_state_t *ret = userdata;
    if (!i)
       return;
    *ret = i->state;
+
+   int iden_len = strlen(i->name) + 1;
+   int name_len = strlen(i->description) + 1;
```
SiegeLord, line 103: Same question about description vs name.

SiegeLord (changes requested):
This looks good, thanks. I'm mostly concerned about the control flow in the alsa code causing silent failure and I'm confused about the name vs description distinction this code is making.

addons/audio/alsa.c
pedro-w: Also: if this continue branch is taken, will we leak the memory allocated by `snd_device_name_hint` (line 926), since it will skip the `snd_device_name_free_hint` (line 957)

addons/audio/alsa.c
pedro-w: For ALSA,  the 'name' is a (presumably unique) system id, and 'desc' is a friendly name, for example I have name=`alsa_output.pci-0000_00_05.0.analog-stereo` and desc=`Built-in Audio Analog Stereo`
So I think the friendly name is the one that the user should see.
I haven't checked other platforms though,

SiegeLord:
This got merged via #1263.

PR #1254, opened 2021-06-16
Allow space for string null terminator in sscanf

pedro-w:
Causes an error if using ASan


SiegeLord:
Bugs in the test driver itself? Astonishing.

Thanks for the fix!

pedro-w:
It doesn't get triggered by the existing test scripts AFAICS. I just came across it because I was building/running Allegro with `-fsanitize=address` (to sort out that bitmap-loading stuff) and it came up with a few hits.

PR #1255, opened 2021-06-16
Fix size calculation for memmove

pedro-w:
Previously was moving one element too many. Found using the address sanitizer on scan_examples.c


SiegeLord:
Thanks!

PR #1256, opened 2021-06-16
Fix memory leak

pedro-w:
Don't need to call xstrdup as aatree already does it.

### DIFF
--- a/docs/scripts/make_index.c
+++ b/docs/scripts/make_index.c
@@ -10,15 +10,6 @@
 #include "dawk.h"
 #include "aatree.h"
 
-static char *xstrdup(const char *s)
-{
-   size_t n = strlen(s);
-   char *p = malloc(n + 1);
-   memcpy(p, s, n);
-   p[n] = '\0';
-   return p;
-}
-
 static void print_link(const char *value)
 {
    d_printf("* [%s]\n", value);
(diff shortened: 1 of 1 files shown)


SiegeLord:
I don't quite follow the logic of this change. I don't see where in aatree.c we're duplicating the key/value. If aatree is owning pointers, I'd expect us to write a `_al_aa_free` w/ a custom key/value destructor.

pedro-w:
When calling `aa_insert` it recurses but will eventually either call xstrdup [here](https://github.com/liballeg/allegro5/blob/5595375931ba405c5433d3c48a359126884a5d10/docs/scripts/aatree.c#L75) if the key exists, or call `aa_singleton` [here](https://github.com/liballeg/allegro5/blob/5595375931ba405c5433d3c48a359126884a5d10/docs/scripts/aatree.c#L51-L60) if it doesn't.
So, it always makes a copy of the value and either copies the key or uses the existing key. Therefore calling `aa_insert` with a key and value does _not_ transfer ownership of that key and value to the Aatree.
This means that when we call `aa_insert` from make_index.c we don't need to duplicate the key/value just before:
https://github.com/liballeg/allegro5/blob/5595375931ba405c5433d3c48a359126884a5d10/docs/scripts/make_index.c#L54-L56
Does that make sense?

SiegeLord:
Ok, that makes sense. Thanks for the explanation and the fix!

Issue #1260, opened 2021-07-06
OSX threads randomly crash


allefant:
I was stuck with a Mac for two weeks and tried to compile one of my games but it would randomly crash. The backtrace always pointed to osx_thread_init in Allegro but otherwise was completely random, could work for a few seconds or half an hour. This simple test case seems to reproduce it for me:

```c
#include <allegro5/allegro.h>
#include <stdio.h>
#include <time.h>

static void* aproc(ALLEGRO_THREAD * thread, void * arg) {
    usleep(10000);
    return NULL;
}

int main() {
    al_init();

    ALLEGRO_THREAD *t[10000];

    for (int i = 0; i < 10000; i++) {
        t[i] = al_create_thread(aproc, NULL);
        al_start_thread(t[i]);
    }

    for (int i = 0; i < 10000; i++) {
        al_join_thread(t[i], NULL);
    }
    return 0;
}
```

Compiled with: `clang a.c -lallegro_monolith-debug -g`

And backtrace:
```
Eliass-MacBook-Pro:test eliaspschernig$ lldb ./a.out
(lldb) target create "./a.out"
Current executable set to '/Users/eliaspschernig/temp/test/a.out' (x86_64).
(lldb) r
Process 49743 launched: '/Users/eliaspschernig/temp/test/a.out' (x86_64)
a.out(49743,0x70000ff86000) malloc: pointer 0x101610080 being reallocated was not allocated
a.out(49743,0x70001010f000) malloc: Double free of object 0x10120d3b0
a.out(49743,0x70001010f000) malloc: *** set a breakpoint in malloc_error_break to debug
a.out(49743,0x70000ff86000) malloc: *** set a breakpoint in malloc_error_break to debug
Process 49743 stopped
* thread #13, stop reason = signal SIGABRT
    frame #0: 0x00007fff7094833a libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
->  0x7fff7094833a <+10>: jae    0x7fff70948344            ; <+20>
    0x7fff7094833c <+12>: movq   %rax, %rdi
    0x7fff7094833f <+15>: jmp    0x7fff70942629            ; cerror_nocancel
    0x7fff70948344 <+20>: retq   
Target 0: (a.out) stopped.
(lldb) bt
* thread #13, stop reason = signal SIGABRT
  * frame #0: 0x00007fff7094833a libsystem_kernel.dylib`__pthread_kill + 10
[...]

PR #1261, opened 2021-07-07
fix osx crash

allefant:
This fixes https://github.com/liballeg/allegro5/issues/1260 - I have no idea what _osx_threads was meant to be used for but it is not currently being used and obviously had no guards against concurrent writes, causing the crash.


SiegeLord:
Looks fine to me!

PR #1265, opened 2021-07-18
Fix the issue in _al_list_at function, when index is greater or equal…

lorry-lee:
… list->size / 2, it passed 1 extra element.

### DIFF
--- a/src/misc/list.c
+++ b/src/misc/list.c
@@ -595,7 +595,7 @@ _AL_LIST_ITEM* _al_list_at(_AL_LIST* list, size_t index)
 
       index = list->size - index - 1;
 
-      while (index--)
+      while (--index)
          item = item->prev;
 
       return item;


SiegeLord:
I just added some actual tests for list.c, but they appear to pass without this change. When I add in this change, I get an infinite loop. Are you sure this is right? To try the tests yourself, pull the latest master branch and run: `make run_standalone_tests`.

lorry-lee:
@SiegeLord I've tried it, well on windows with visual studio, the test failed without my change:

      // Here it already got the wrong item returned when i == 0 because of the issue I'm fixing
      item = _al_list_at(list, i);
      // This throw exception for me because of accessing invalid memory
      assert(*(int*)_al_list_item_data(item) == xs[i]);

With my change, there is no problem of running the test.

Also with a simple test I did on my side, without my change it doesn't work, and it worked after I apply my change:

   int xs[11] = {0,1,2,3,4,5,6,7,8,9,10};

   _AL_LIST* list = _al_list_create();

   for (int i = 0; i < 11; ++i)
      _al_list_push_back(list, &xs[i]);

   _AL_LIST_ITEM* item = _al_list_at(list, 7);
   int value = (*(int*)_al_list_item_data(item));
   assert(7 == value);

   _al_list_destroy(list);

pedro-w:
Didn't some of this code get changed recently, maybe with the audio driver
listing? Sorry haven't got any access at the moment so can't check.

pedro-w:
I think it was already fixed by https://github.com/liballeg/allegro5/pull/1263/commits/c2c9c8b288f84cc4619d4d9c32f4102be5478cae in a different way. Do you agree?

lorry-lee:
I agree it's the same fix.

Maybe I misunderstood something, but with my fix, I think it should pass, which is the same result as my local test.

I'm not sure what is/should be the status without my fix.

SiegeLord:
I checked again, and I think the code is correct as is (with the previous fix). Thanks either way!

PR #1267, opened 2021-07-18
android: Update Gradle to 5.0

dos1:
This brings OpenJDK 11 support in.

I had troubles with Gradle versions newer than that, so decided to not jump higher as OpenJDK 11 is a version that's still supported - as opposed to OpenJDK 8 that was required with Gradle 4.6 which is end-of-life and unavailable in several distros already.


SiegeLord:
Everything appears to work well, thanks!

PR #1268, opened 2021-07-23
SDL: touch: Ignore touch events from indirect touch devices

dos1:
There's no equivalent API in Allegro for them yet.


SiegeLord:
Thanks!

PR #1269, opened 2021-07-23
SDL: Use SDL_PeepEvents for event handling

dos1:
This should improve performance on platforms where SDL_PumpEvents
is expensive, like Emscripten.

### DIFF
--- a/src/sdl/sdl_system.c
+++ b/src/sdl/sdl_system.c
@@ -21,6 +21,8 @@ ALLEGRO_DEBUG_CHANNEL("SDL")
 
 static ALLEGRO_SYSTEM_INTERFACE *vt;
 
+#define ALLEGRO_SDL_EVENT_QUEUE_SIZE 8
+
 #ifdef DEBUGMODE
 #define _E(x) if (type == x) return #x;
 static char const *event_name(int type)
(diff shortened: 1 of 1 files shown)


dos1:
Updated to get rid of C99ism.

src/sdl/sdl_system.c
```diff
@@ -73,54 +75,58 @@ static char const *event_name(int type)
...
-                  _al_sdl_keyboard_event(&event);
-                  break;
-               case SDL_WINDOWEVENT_CLOSE:
-               case SDL_WINDOWEVENT_RESIZED:
-                  _al_sdl_display_event(&event);
-                  break;
-            }
+   SDL_PumpEvents();
+   int n = 0, i;
+   while ((n = SDL_PeepEvents(events, ALLEGRO_SDL_EVENT_QUEUE_SIZE, SDL_GETEVENT, SDL_FIRSTEVENT, SDL_LASTEVENT))) {
```
SiegeLord, line 83: This should be > 0, if I'm reading the documentation correctly.

dos1: Good catch! Updated.

SiegeLord:
Thanks!

PR #1270, opened 2021-07-24
Fix build failure with sanitizers enabled

tehsausage:
Without this change the build fails with GCC's sanitizers enabled (probably ubsan) due to calling strncpy with a NULL source parameter:

```
[...]/allegro5/docs/scripts/dawk.c:247:10: runtime error: null pointer passed as argument 2, which is declared to never be null
    #0 0x558d2fc7fda0 in d_match [...]/allegro5/docs/scripts/dawk.c:247
    #1 0x558d2fc7d698 in main [...]/allegro5/docs/scripts/make_html_refs.c:34
    #2 0x7f97781ec7ec in __libc_start_main ../csu/libc-start.c:332
    #3 0x558d2fc7d389 in _start ([...]/allegro5-native-Debug/docs/make_html_refs+0x1c389)
```

Reproduction: build with gcc flag `-fsanitize=undefined`, and environment variable `UBSAN_OPTIONS=print_stacktrace=1,halt_on_error=1,abort_on_error=1` set.

### DIFF
--- a/docs/scripts/dawk.c
+++ b/docs/scripts/dawk.c
@@ -244,7 +244,8 @@ bool d_match(dstr line, const char *regex)
 
    for (i = 0; i < MAX_MATCH; i++) {
       if (trex_getsubexp(re->reg, i, &match)) {
-         strncpy(d_submatches[i], match.begin, match.len);
+         if (match.begin)
+            strncpy(d_submatches[i], match.begin, match.len);
          d_submatches[i][match.len] = '\0';
       }
       else {


pedro-w (commented):
This would certainly fix the error but the original doesn't look right to me - for `strncpy`, isn't the `len` argument supposed to be the size of the _destination_? If you know the length to copy already (ie. `match.len`) it would be better to use `memcpy` And (I haven't checked) but somewhere we need to be sure that `match.len` isn't greater than the size of `d_submatches[i] - 1`

tehsausage:
Considering that asan isn't giving me grief about buffer overflows, its fair to say that isn't an issue right now. I think someone would have to drop a 3000 character line in to one of the documentation files in order to make anything break.

dos1:
@pedro-w Looking at that code it seems to me like it's using strncpy correctly:

```
The strncpy() function is similar [to strcpy], except that at most n bytes of src are copied.
```

When using `MAX_MATCH` instead of `match.len` it would copy everything till the end of the string, not just the matching substring. Seems like `match.len` should always be less than `MAX_MATCH`. You could use `memcpy` instead indeed, but from quick glance at the current code it appears to me to be correct nevertheless.

pedro-w:
I do agree entirely. I should have said more clearly, I didn't mean that the code (with @tehsausage patch) is buggy, just that strncpy isn't the simplest function that would do the job here (and I like simple!) As long as we're sure that this can't overflow the destination, I'm OK with it.

I did hope that `memcpy(..., NULL, 0)` would be valid (albeit a no-op) and then we could skip the test. But it seems that Posix says that's undefined behaviour, according to [this](https://www.imperialviolet.org/2016/06/26/nonnull.html)

By the way, is it the case that `match.begin` is only NULL, when `match.len` is zero, i.e. it's an empty capture? We're not just ignoring some error condition elsewhere are we?

SiegeLord:
> By the way, is it the case that match.begin is only NULL, when match.len is zero, i.e. it's an empty capture? We're not just ignoring some error condition elsewhere are we?

I glanced at the code, I think it is the case.

Let's merge this. Thanks!

PR #1275, opened 2021-08-07
linux: Drop old and unused files

dos1:


SiegeLord:
Weird! Well, they're truly unused, no mention of them at all in `FileList.cmake`. Seems harmless to get rid of these.

beoran:
These were for running directly on the kernel without X11 nor Wayland. But i guess that plan never panned out.

SiegeLord:
The WIP KMS driver doesn't use these, so I'm not sure what use these have. We can always re-add them if they end up being necessary.

PR #1277, opened 2021-08-13
Add audio identification / al_identify_sample()

tehsausage:
This adds the function al_identify_sample() and related function al_register_sample_identifier(), which adjusts the behavior of al_load_sample() and al_load_audio_stream() to function the same way as al_load_bitmap() and al_load_video(), by attempting to detect the file format before considering the file extension.

Some of the identifier functions, specifically for some DUMB formats and the heuristic MP3 detection, have real potential to break code that may currently be using custom audio formats registered with al_register_sample_loader() and al_register_audio_stream_loader(), due to the potential for false-positives. Particularly uncompressed mod/tracker formats are very prone to being mis-detected as MP3 due to repetitive byte patterns.

The MP3 identifier function uses an algorithm that is similar to minimp3's mp3dec_detect family of functions, but has the advantage of only loading a fixed portion of the file contents (8KB), and typically runs about 100x faster than calling mp3dec_detect().

_This contains some code copy-pasted verbatim from minimp3, to avoid dependencies on non-public interface parts of minimp3, however it is CC0 licensed (https://github.com/lieff/minimp3/blob/master/LICENSE) and I invoke my right to copy this code and re-release it under Allegro's license._

A new pair of functions, al_identify_video() and al_identify_video_f(), are also added to the video addon for interface parity with image and audio. They function as expected and only support detecting the file format ".ogv".

Outdated al_load_bitmap() documentation is also fixed to correctly describe how format detection actually works.

Test program with results: https://gist.github.com/tehsausage/1d1f17344ed474e8c8fe0005a0c277d6

### DIFF
--- a/addons/acodec/acodec.c
+++ b/addons/acodec/acodec.c
@@ -24,25 +24,23 @@ bool al_init_acodec_addon(void)
 
    ret &= al_register_sample_loader(".wav", _al_load_wav);
    ret &= al_register_sample_saver(".wav", _al_save_wav);
-   ret &= al_register_audio_stream_loader(".wav", _al_load_wav_audio_stream);
-
    ret &= al_register_sample_loader_f(".wav", _al_load_wav_f);
    ret &= al_register_sample_saver_f(".wav", _al_save_wav_f);
+   ret &= al_register_audio_stream_loader(".wav", _al_load_wav_audio_stream);
    ret &= al_register_audio_stream_loader_f(".wav", _al_load_wav_audio_stream_f);
+   ret &= al_register_sample_identifier(".wav", _al_identify_wav);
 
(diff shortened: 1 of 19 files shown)


SiegeLord:
This is awesome, thanks! The code looks great; my only question is whether you think the copyright status of your test files is convenient enough to be able to use it in Allegro's automated tests? If so, I'd prefer to add tests for this at some later point.

tehsausage:
Mostly very copyrighted and not easily acquirable. It probably wouldn't be too hard to put together a set of small test files using ffmpeg, for the non-DUMB formats; but finding and exporting something from a dozen different tracker programs would be a lot of work, and I think some of the formats are even proprietary ones used in old game engines.

Issue #1276, opened 2021-08-08
No 5.2.7 build on launchpad for Xenial 16.04


lockie:
Hello! I'm using liballegro Xenial build from [the launchpad ppa](https://launchpad.net/~allegro) to make sure my game build runs on the widest range of old Linux distributions. Unfortunately, there is no newest 5.2.7 build for Xenial due to [some kind of liballegro_dialog related package bug](https://launchpad.net/~allegro/+archive/ubuntu/5.2/+build/21726297); the build log says
```
dh_install: liballegro-dialog5.2 missing files: usr/lib/*/liballegro_dialog.so.*
```
I suspect this is related to moving to GTK3 and this other line in the log:
```
- Checking for module 'gtk+-3.0'
--   No package 'gtk+-3.0' found
```
I have Xenial-based system handy, and it seems that it has the package called `libgtk-3-dev` which most probably will solve the issue.

SiegeLord:
I tried that, but got an error:

```
../lib/liballegro_dialog.so.5.2.7: undefined reference to `gtk_menu_popup_at_widget'
```

Maybe the Xenial GTK3 is too old. Perhaps we could use a different function? Can you reproduce this error on your Xenial system?

lockie:
Strangely enough, both on my Xenial-based Mint laptop and under `ubuntu:xenial` docker container, I was able to compile `liballegro_dialog` from 5.2.7.0 sources just fine, only having the warning about that function that reads
```
/tmp/allegro-5.2.7.0/addons/native_dialog/gtk_menu.c: In function 'do_show_popup_menu':
/tmp/allegro-5.2.7.0/addons/native_dialog/gtk_menu.c:390:7: warning: implicit declaration of function 'gtk_menu_popup_at_widget' [-Wimplicit-function-declaration]
       gtk_menu_popup_at_widget(args->menu->extra1, menu,  GDK_GRAVITY_SOUTH_WEST,
       ^
```
The `libgtk-3-dev` package version on both systems is `3.18.9-1ubuntu3.3`.
I wonder whether Launchpad build environment is somehow different from pristine Xenial install which exists in docker container.

dos1:
If you got that warning, it means that build is likely broken anyway. Have you tried actually running anything that uses allegro-dialog on xenial?

lockie:
@dos1 Oops, you're right: trying to compile ex_menu example from Allegro source tree, I'm getting
```
[...]

PR #1279, opened 2021-08-19
Fix native dialogs addon compatibility with older GTK+ versions

lockie:
This is part of the fix for #1276 . Having this commit cherry-picked on top of `5.2.7` branch, I was able to build liballegro_dialog and ex_menu, the latter working as expected.

### DIFF
--- a/addons/native_dialog/gtk_menu.c
+++ b/addons/native_dialog/gtk_menu.c
@@ -393,7 +393,7 @@ static gboolean do_show_popup_menu(gpointer data)
                               GDK_GRAVITY_NORTH_WEST,
                               NULL);
 #else
-      gtk_menu_popup(args->menu->extra1, menu, NULL, NULL, NULL, 1, 0);
+      gtk_menu_popup(args->menu->extra1, NULL, NULL, NULL, NULL, 1, 0);
 #endif
 
    if (!position_called) {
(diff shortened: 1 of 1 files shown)


addons/native_dialog/gtk_menu.c
```diff
@@ -387,9 +387,14 @@ static gboolean do_show_popup_menu(gpointer data)
 
    bool position_called = false;
    if (menu)
+      /* gtk_menu_popup_at_widget only exists in gtk newer than 3.22 */
+#if GTK_CHECK_VERSION(3, 22, 0)
       gtk_menu_popup_at_widget(args->menu->extra1, menu,  GDK_GRAVITY_SOUTH_WEST,
                               GDK_GRAVITY_NORTH_WEST,
                               NULL);
+#else
+      gtk_menu_popup(args->menu->extra1, menu, NULL, NULL, NULL, 1, 0);
```
SiegeLord, line 396: I tried this, and this actually completely locked up ex_menu. I got it to sort of work if the second argument was set to NULL, but I note that even on newer GTKs this code appears to be completely broken.

Let's change the second argument to NULL and call it a day, we'll fix the underlying issue at some later point.

lockie: Weird, it worked on my machine with no hiccup. Anyway, I'm no expert on GTK programming :sweat_smile:

PR #1286, opened 2021-10-23
Make al_get/set_window_position work on Retina displays for OSX.

### LINKED ISSUE #923: al_get_monitor_info returns wrong values on Retina displays
On macOS, `al_get_monitor_info` returns the wrong values when called on a Retina display, specifically it seems like `x2` and `y2` are half of what they're supposed to be.  As miniSphere uses this function to determine window scaling and position, this results in game windows which are too small and off-center:

screen_shot_2018-08-04_at_11 17 24_am

As Allegro is otherwise Retina-aware, `al_get_monitor_info` should return the native resolution rather than the one the system is faking.

SiegeLord:
Fixes #923

### DIFF
--- a/README_macosx.txt
+++ b/README_macosx.txt
@@ -42,10 +42,12 @@ two lines to the Info.plist using a text editor:
 
 If you are making an unbundled app this feature appears to be enabled by
 default, but it is not recommended to rely on this working: make a bundled
-app for the most predictable performance.
+app for the most predictable behavior.
 
-When changing the display DPI or moving the window between two displays with
-different DPIs, Allegro displays behave in the following way:
+Allegro uses a pixel-based coordinate system, meaning that high-DPI displays
+will be larger than is reported by the OS. When changing the display DPI or
+moving the window between two displays with different DPIs, Allegro displays
+behave in the following way:
 
 - If the ALLEGRO_DISPLAY was created with the ALLEGRO_RESIZABLE flag it will
   send an ALLEGRO_DISPLAY_RESIZE event. This will have the effect of your app's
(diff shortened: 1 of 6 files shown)


connorjclark:
I have the new Mac M1 which has a retina display, would it help to have a second pair of eyes test this PR?

SiegeLord:
Thanks, but not right now. Maybe when it's closer to being done, right now it's not doing a correct thing, so no point in discovering what's wrong :).

SiegeLord:
@pedro-w I think I got a workable solution to the monitor info/window position issues. Here's the idea:

1. Find the largest scale factor amongst the connected monitors, use this as a "global scale factor"
2. Arrange all the monitors on a single plane, scaling their origins using the global scale factor.
3. Transform to/from this plane when calling get/set window position.

It looks something like this:

osx_scaling

The encoding I choose is very simplistic in that adds dead zones between monitors if there's heterogeneity in scales. In the simple example above, it can be avoided, but for more complex multi-monitor setups I don't think it can be. I prefer the simple route here.

pedro-w:
So you've got a uniform space on which to position the windows? I like the idea, it is appealingly simple. Unfortunately I still have no hardware to test it myself 😞 
What happens if you position a window in the dead zone?

SiegeLord:
>  What happens if you position a window in the dead zone?

It silently fails.

fatcerberus:
Thanks for finally getting around to fixing this @SiegeLord, I can finally stop having to drag the neoSphere window back on-screen every single time I playtest a game (which got extremely tedious after a while) 🎉 

OS X DPI scaling really seems insanely overengineered.  Say what you will about spotty DPI awareness on Windows, but at least the OS always gives you physical pixels to work with.

PR #1287, opened 2021-10-25
Update video.c

Jaysonjson:
al_init_video_addon() did not update the bool "video_inited", therefore making it always stay on the default given value (false) which gives al_is_video_addon_initialized() a wrong result

Found it when al_init_video_addon gave me a 1 at debuggin and al_is_video_addon_initalized a 0,
Tested on Win11 and Arch Linux (If somehow, for some unknown reason, video_inited changes using magic powder)


SiegeLord:
Thanks!

PR #1289, opened 2021-11-11
Fix _al_unix_get_time()

check-switch-26:
Fix _al_unix_get_time()
Changing the system clock would cause discontinuities in the return value.

Use clock_gettime() with CLOCK_MONOTONIC instead of gettimeofday() to solve the problem.
(Bonus: no longer using gettimeofday(), which has been deprecated)

Tested on Ubuntu 18.04.5 LTS


SiegeLord:
Thanks!

PR #1290, opened 2021-11-11
Update allegro_font.h

check-switch-26:
Fix declaration of al_draw_multiline_textf() as a PRINTFUNC

Tested with gcc 7.5.0


SiegeLord:
Thanks!

PR #1291, opened 2021-11-11
Fix DDS compressed image loading

check-switch-26:
Compressed images with dimensions that are not a multiple of the block size were not loaded correctly.  This is fixed by rounding up the pitch and line count.

Tested on Windows 10 with OpenGL display.

### DIFF
--- a/addons/image/dds.c
+++ b/addons/image/dds.c
@@ -143,8 +143,8 @@ ALLEGRO_BITMAP *_al_load_dds_f(ALLEGRO_FILE *f, int flags)
 
    bitmap_data = lr->data;
 
-   for (ii = 0; ii < h / block_height; ii++) {
-      size_t pitch = (size_t)(w / block_width * block_size);
+   for (ii = 0; ii < (h + block_height - 1) / block_height; ii++) {
+      size_t pitch = (size_t)((w + block_width - 1) / block_width * block_size);
       num_read = al_fread(f, bitmap_data, pitch);
       if (num_read != pitch) {
          ALLEGRO_ERROR("DDS file too short.\n");


SiegeLord:
Are such bitmaps valid? I was under the impression that DDS bitmaps must have width/height be a multiple of 4. Where did you get such a bitmap?

I'm primarily concerned that Allegro is probably making assumptions about the bitmap size being an exact multiple of block size in many more locations than this.

check-switch-26:
> Are such bitmaps valid? I was under the impression that DDS bitmaps must have width/height be a multiple of 4. Where did you get such a bitmap?

I compressed images with [Compressonator](https://github.com/GPUOpen-Tools/compressonator)
I am certain that width/height are not required to be a multiple of 4.  Consider:

- "For non-power-of-two images that aren't a multiple of 4 in size, the other colors of the 4x4 block are taken to be black." (https://www.khronos.org/opengl/wiki/S3_Texture_Compression)
- MS DirectXTex code does the same rounding up of block "blocks wide" and "blocks high" (https://github.com/microsoft/DirectXTex/blob/e9d6e9c79796d5a4f07876359d67ef898bdea203/DDSTextureLoader/DDSTextureLoader9.cpp#L449)
- I can open my oddly-sized DDS files in a number of image viewers without problem.
 
> I'm primarily concerned that Allegro is probably making assumptions about the bitmap size being an exact multiple of block size in many more locations than this.

Based on my review of `_al_ogl_create_bitmap()`, plus my experience with using oddly-sized DDS images with the proposed commit, I'm pretty confident this is all handled correctly for OpenGL.

`d3d_create_bitmap()` appears to be doing the right thing.  I don't have an environment for building Allegro for D3D support, so I don't have a way to test.

SiegeLord:
I tested this a bit, and am convinced that this is correct. Thanks!

PR #1292, opened 2021-11-11
Improve TT font rendering when using rotation transforms

check-switch-26:
TrueType rendering with a rotation transform results in "jaggies" on the outside portions of glyphs.

Glyph bitmap rendering now includes the 1-pixel border around the glyph.  Also, border changed to "transparent white" for ALLEGRO_NO_PREMULTIPLIED_ALPHA.

See attached files for "before" and "after" examples using DejaVuSans.ttf and a rotation transform of ALLEGRO_PI / 32.0

**Before Fix - no premultiplied alpha**
Before Fix - no premultiplied alpha
**Before Fix - premultiplied alpha**
Before Fix - premultiplied alpha
**After Fix - no premultiplied alpha**
After Fix - no premultiplied alpha
**After Fix - premultiplied alpha**
After Fix - premultiplied alpha

Tested on Windows 10 and on Ubuntu 18.04.5 LTS, both using an OpenGL 3.0 display.

### DIFF
--- a/addons/ttf/ttf.c
+++ b/addons/ttf/ttf.c
@@ -301,13 +301,32 @@ static unsigned char *alloc_glyph_region(ALLEGRO_TTF_FONT_DATA *data,
          return NULL;
       }
 
-      /* Clear the data so we don't get garbage when using filtering
-       * FIXME We could clear just the border but I'm not convinced that
-       * would be faster (yet)
-       */
-      for (i = 0; i < lock_rect.h; i++) {
-          ptr = (char *)(data->page_lr->data) + (i * data->page_lr->pitch);
-          memset(ptr, 0, lock_rect.w * 4);
+      if (data->flags & ALLEGRO_NO_PREMULTIPLIED_ALPHA) {
+         /* Fill in border with "transparent white"
+          * to make anti-aliasing work well when rotation
+          * and/or scaling transforms are in effect
+          * FIXME We could clear just the border
+          */ 
+         for (i = 0; i < lock_rect.h; i++) {
+            ptr = (char *)(data->page_lr->data) + (i * data->page_lr->pitch);
+            int j;
+            for (j = 0; j < lock_rect.w; ++j) {
+               *ptr++ = 255;
+               *ptr++ = 255;
+               *ptr++ = 255;
+               *ptr++ = 0;
+            }
+         }
+      }
+      else {
+         /* Clear the data so we don't get garbage when using filtering
+          * FIXME We could clear just the border but I'm not convinced that
+          * would be faster (yet)
+          */
+         for (i = 0; i < lock_rect.h; i++) {
+            ptr = (char *)(data->page_lr->data) + (i * data->page_lr->pitch);
+            memset(ptr, 0, lock_rect.w * 4);
+         }
       }
    }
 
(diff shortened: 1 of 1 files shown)


addons/ttf/ttf.c
```diff
@@ -526,6 +551,7 @@ static bool ttf_get_glyph_worker(ALLEGRO_FONT const *f, int prev_ft_index, int f
 
    if (glyph->page_bitmap) {
       info->bitmap = glyph->page_bitmap;
+	   /* the adjustments below remove the 1-pixel border from the glyph */
```
SiegeLord, line 554: Could we remove these adjustments so that the glyph info contains the border?

SiegeLord (changes requested):
The screenshots look amazing. I think this is good, but I want to see if we could stick these offsets into the ALLEGRO_GLYPH values.

check-switch-26:
Removing all those adjustments would be nice because the adjustments in render_glyph() could also be removed.

However, `struct ALLEGRO_GLYPH` is exposed to the API (although it is UNSTABLE).
https://liballeg.org/a5docs/trunk/font.html#allegro_glyph

An intermediate routine could be created to massage the data returned via `al_get_glyph()` to preserve the existing interface.

On the other hand, someone using `al_get_glyph()` would probably like to use the border as well for rendering, so changing the API there would be an improvement even though it would break existing code.  The change would have to be carefully documented, and maybe the struct name should be changed to force a compile error on existing code.

I'm happy to make and test either improvement.  I think it should be a separate commit, though, in order to keep the code changes clear.  Please advise on how you'd like me to proceed.

addons/ttf/ttf.c
check-switch-26: Removing all those adjustments would be nice because the adjustments in `render_glyph()` could also be removed.

However, `struct ALLEGRO_GLYPH` is exposed to the API (although it is UNSTABLE).
https://liballeg.org/a5docs/trunk/font.html#allegro_glyph

An intermediate routine could be created to massage the data returned via `al_get_glyph()` to preserve the existing interface.

On the other hand, someone using `al_get_glyph()` would probably like to use the border as well for rendering, so changing the API there would be an improvement even though it would break existing code. The change would have to be carefully documented, and maybe the struct name should be changed to force a compile error on existing code.

I'm happy to make and test either improvement. I think it should be a separate commit, though, in order to keep the code changes clear. Please advise on how you'd like me to proceed.

SiegeLord: I don't think the worries about backwards compatibility are warranted. The way you'd generically use `ALLEGRO_GLYPH` is to use the `offset_x` etc just like the `render_glyph` does. By adding the border, the only thing we'd change is the values of the glyph parameters, but the code that uses would both remain the same and retain the same output (when no transformations are used).

Does that make sense?

check-switch-26: I understand your explanation, but I respectfully disagree.  Someone calling `al_get_glyph()` and then rendering instead of simply calling `al_draw_glyph()` is likely to be using the info in `ALLEGRO_GLYPH` to perform additional work.

Although not explicitly stated, it is implied that the info in `ALLEGRO_GLYPH` is the same as that returned by `al_get_glyph_dimensions()`.  Unexpectedly changing that correlation is likely to break existing code.  Here are three non-contrived examples:

- _Centering_ and  _Right justification_ - these calculations will be wrong and result in incorrect glyph placement
- _Inverse_ or _Cursor_ plotting - the resulting boxes will be expanded by 2 pixels in both directions
- _Scaling_ - attempting to scale a glyph to fill a given space will use the wrong width and height

check-switch-26: Another consideration:

Changing the `ALLEGRO_GLYPH` values returned by `al_get_glyph()` for TTF fonts would make out of sync from the values returned for bitmap and bmfonts.  In my opinion, the `al_get_glyph()` API would end up being rather complicated.

SiegeLord: Alright, you make reasonable points. We'll keep it the way you have it, and if we regret it, we can add a flag.

Issue #1212, opened 2021-01-03
Expose sRGB <-> linear conversion functions


SiegeLord:
They already exist in the Color addon, just need an API. These things are useful if you're doing color interpolation etc.

Issue #1211, opened 2021-01-03
Add OKLAB  to the color addon


SiegeLord:
See https://bottosson.github.io/posts/oklab/. It's a nice uniform

allefant:
That looks really nice - I don't think it existed when I last played with colors (and added the CIELAB functions to Allegro).

PR #1294, opened 2021-11-16
add linear and oklab color spaces to the color addon

allefant:
This addresses issues #1211 and #1212.

I'm also working on an ex_color3 example which will use the new functions, but that will be a separate PR.


allefant:
Actually might as well have it in the same PR!

It's quite interesting, the blue->white gradient from the original description of course looks beset with Oklab. As does the blue->yellow one. The red->green is identical to CIE Lab. But then for the black->pink Oklab completely fails and the left half is just black, with both plain RGB and CIE Lab looking better. Guess we have to wait for Oklab2 to *really* create a uniform color space, heh :)

allefant:
Found a nice review of oklab and it may explain the darkness in ex_gradient: https://raphlinus.github.io/color/2021/01/18/oklab-critique.html

SiegeLord:
Nice!

PR #1296, opened 2021-11-30
Bug in Windows OpenGL display creation

HerrNamenlos123:
There's an infinite loop which ensures to focus the window when being created, which is apparently only strictly needed in fullscreen mode.

Sometimes the window simply cannot be focused, in which case the window creation fails after 10 seconds. This happens sporadically and is not reproducable, but for me it happens every time when launching the executable with a WiX installer after installation has finished. (Seems to lock the foreground?)

The interval of failure also seems to be dependent on the speed of the computer. It fails more often in the virtual machine which is very slow, but it has also failed on my main dev pc at some point. By starting an allegro application at the same time from 150 threads i was able to reproduce it sometimes.

Since focusing isn't strictly necessary when not in fullscreen mode, this simple fix makes it work for me, no more crashes (window is still being focused tho).

### DIFF
--- a/src/win/wgl_disp.c
+++ b/src/win/wgl_disp.c
@@ -1241,7 +1241,7 @@ static void display_thread_proc(void *arg)
       DWORD lock_time;
       HWND wnd = win_disp->window;
 
-      if (disp->flags & ALLEGRO_FULLSCREEN == 0) { // Not fullscreen mode
+      if ((disp->flags & ALLEGRO_FULLSCREEN) == 0) { // Not fullscreen mode
          ShowWindow(wnd, SW_SHOWNORMAL);
          SetForegroundWindow(wnd);
          UpdateWindow(wnd);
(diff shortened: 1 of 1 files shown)


pedro-w:
I was surprised to see how complicated 'setting the foreground window' is on Windows!  A lot of the complication seems to come from Windows trying to prevent the annoyance of apps jumping to the front while the user is working on another window. What you've done looks good, since it's only really necessary to avoid the unresponsive fullscreen situation.
However, I wonder if it would be better to re-arrange the surrounding code to split it into two clear blocks for fullscreen and windowed; at the moment it checks 3 separate times for window status in that section.

HerrNamenlos123:
I have cleaned up the conditions in the function now, i have not tested it extensively, but it seems to work.
I am testing it with non-fullscreen windows and everything works, someone should still test it with fullscreen windows.

SiegeLord:
Thanks! There are actually two different fullscreen modes, one is the "fullscreen window", commonly known as "borderless window" these days, and true fullscreen. This patch broke the borderless variant, but the fix was simple. Otherwise, it seems to work fine.

I merged this as https://github.com/liballeg/allegro5/commit/2a8b456657a9514d308b874c05d08581ce0d0cbe.

PR #1308, opened 2022-02-23
Incorrect error handling in PHYSFS_close

j-w-c-b:
Per https://www.icculus.org/physfs/docs/html/physfs_8h.html#a6822f8ff10073e855a1c3a6485b882f2 the return value from PHYSFS_close() is `nonzero on success, zero on error`


SiegeLord:
Wow, that's unfortunate! This isn't the only one too, the `PHYSFS_flush` and `PHYSFS_seek` appear to be broken too.

SiegeLord:
Actually, nevermind... those are fine. Thanks for the change!

PR #1311, opened 2022-02-25
file_slice.c: Add al_fopen_slice() mode to prevent seek on close.

WickedSmoke:
Add 'n' (No seek) mode and 's' (Seek on close).

Ideally, only 's' should exist to enable the extra action, but for
compatibility with the Allegro 5.2 behavior the 'n' mode is needed to
disable it.

 - Removed used of strstr() to optimize handling of mode string.


SiegeLord:
Thanks!

PR #1313, opened 2022-03-05
Allow configuring bitmap wrapping.

SiegeLord:


beoran:
Let me take a look...

src/tri_soft.c
```diff
@@ -682,12 +683,12 @@ static void triangle_stepper(uintptr_t state,
...
    }
 }
-
+#include <stdio.h>
 /*
 This one will check to see what exactly we need to draw...
 I.e. this will call all of the actual renderers and set the appropriate callbacks
 */
-void _al_triangle_2d(ALLEGRO_BITMAP* texture, ALLEGRO_VERTEX* v1, ALLEGRO_VERTEX* v2, ALLEGRO_VERTEX* v3)
+void _al_triangle_2d(ALLEGRO_BITMAP* texture, ALLEGRO_VERTEX* v1, ALLEGRO_VERTEX* v2, ALLEGRO_VERTEX* v3, bool prim_addon)
```
beoran, line 691: This is the boolean flag anti-pattern. It might be better and more performant to have two different functions for the two different cases, especially seeing we are always calling this with a true or false constant.

src/tri_soft.c
```diff
@@ -712,9 +713,16 @@ void _al_triangle_2d(ALLEGRO_BITMAP* texture, ALLEGRO_VERTEX* v1, ALLEGRO_VERTEX
    }
 
    if (texture) {
+      ALLEGRO_BITMAP_WRAP wrap_u, wrap_v;
+      _al_get_bitmap_wrap(texture, &wrap_u, &wrap_v);
+
+      // XXX: Why are we using repeat for regular bitmaps by default?
```
beoran, line 719: Indeed, why?

beoran (commented):
Thanks for your hard work on this. I have two comments I would like you to address if possible.

src/tri_soft.c
SiegeLord: I think boolean flags have their place, although in this case due to the oddity with the repeat/clamp (see the other comment), we turn out not to need this at all.

src/tri_soft.c
SiegeLord: I added more context. It's a bug I don't care to fix.

beoran: Thanks for expanding that comment.

beoran:
Lgtm!

SiegeLord:
Thanks for looking this over.

PR #1314, opened 2022-03-06
Improve emscripten instructions for building demos

connorjclark:
- Formatted command to be copy/paste-friendly
- Turns out you don't need to set all those flags for the ports emscripten manages. However, `SDL2_INCLUDE_DIR` was necessary
- added note about $EM_CACHE (I just setup emsdk and this variable was never set for me...)
- added command for copying the data folder used by demos

### DIFF
--- a/README_sdl.txt
+++ b/README_sdl.txt
@@ -46,14 +46,12 @@ or equivalent as provided by emscripten.
 mkdir build_emscripten
 cd build_emscripten
 
-3. The "--preload-file data" option below includes a folder named "data"
-with each binary, so we make sure each example and demo has such a
-folder:
+If you're going to build a demo, you will need to manually create a file:
 
-mkdir examples
-cp -r ../examples/data examples/data
+mkdir -p demos/speed/data
+touch demos/speed/data/nothing.txt
 
-4. Configure CMake, using emcmake.
+3. Configure CMake, using emcmake.
 
 ```
 USE_FLAGS=(
@@ -95,7 +93,7 @@ its ports system.
 To compile your own game adjust as necessary. You can use the
 lib/liballegro_monolith-static.a library.
 
-5. Compile the library and examples.
+4. Compile the library and examples.
 
 make
 
@@ -104,10 +102,12 @@ may want to compile individual examples, e.g.
 
 make ex_draw_bitmap
 
-6. To run the examples, navigate to the examples folder. At this point it is
+Note: If you make any changes to any data/ files, you will need to manually delete the generated .html
+file in <build dir>/examples to force emscripten to regenerate the .data file.
+
+5. To run the examples, navigate to the examples folder. At this point it is
 easiest to start a local webserver, and then navigate to the examples using a
 web browser. E.g. you could use the Python's web server module which prints out
 a URL you can open:
 
-cd examples
 python3 -m http.server
--- a/README_sdl.txt
+++ b/README_sdl.txt
@@ -50,11 +50,12 @@ cd build_emscripten
 with each binary, so we make sure each example and demo has such a
 folder:
 
-mkdir -p demos/speed/data
-demos/speed/data/nothing.txt
+mkdir examples
+cp -r ../examples/data examples/data
 
 4. Configure CMake, using emcmake.
 
+```
 USE_FLAGS=(
     -s USE_FREETYPE=1
     -s USE_VORBIS=1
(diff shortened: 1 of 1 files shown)


SiegeLord:
I tried your version, but it wouldn't find png and other dependencies without those extra flags. My SDK version is 2.0.14, FWIW.

It's weird that we need to copy the bitmap data manually, it should be enough to do `make copy_example_data ex_draw_bitmap`, but it looks like there's a dependency missing in CMake to make sure that happens automatically/it's correctly synced.

Overall, though, this looks great, thanks!

connorjclark:
I also got an error about no libpng being found (with the original command and the one in the PR). It went away once I did this (in the emsdk folder):

```
python3 upstream/emscripten/embuilder.py build sdl2_image_png
```

I left it out of the PR because I cleared the emsdk cache folder and tried everything again from zero, and that time it just worked 🤷 If this command fixes it for you I'll add to the PR.

> My SDK version is 2.0.14, FWIW.

I just downloaded the SDK so I have the newest, 3.1.6.

> It's weird that we need to copy the bitmap data manually, it should be enough to do make copy_example_data ex_draw_bitmap, but it looks like there's a dependency missing in CMake to make sure that happens automatically/it's correctly synced.

https://github.com/connorjclark/allegro5/blob/f41f49d65c89fdb214b4dd49172b5adfbf16c4ec/examples/CMakeLists.txt#L271-L271

I see that some targets specify `copy_example_data` as a dependency (the demos), but for the examples there is just this function call. I don't know CMAKE that well but I assume that the examples build targets need to specify this dependency explicitly?

SiegeLord:
Ok, I re-downloaded my SDK, and your instructions appeared to work fine. I didn't need to set `EM_CACHE`, but probably doesn't hurt to leave that in. Let me suggest two changes and then this is good to go.

README_sdl.txt
```diff
@@ -50,11 +50,12 @@ cd build_emscripten
 with each binary, so we make sure each example and demo has such a
 folder:
 
-mkdir -p demos/speed/data
```
SiegeLord, line 53: This needs to be reverted as this was actually correct: the speed demo has no data directory, so it won't build without this.

For examples, instead of this change, could you add `make copy_example_data` before `make ex_draw_bitmap` below? I'll look into fixing the cmake dependency later so this happens automatically.

README_sdl.txt
```diff
@@ -119,4 +109,5 @@ easiest to start a local webserver, and then navigate to the examples using a
 web browser. E.g. you could use the Python's web server module which prints out
 a URL you can open:
 
+cd examples
```
SiegeLord, line 112: Let's leave this off as well, as running the server in the build directory lets you play the demos as well.

SiegeLord (changes requested):

SiegeLord:
Maybe with https://github.com/liballeg/allegro5/pull/1317 we won't need to explicit `make copy_example_data`.

README_sdl.txt
```diff
@@ -114,7 +102,10 @@ may want to compile individual examples, e.g.
 
 make ex_draw_bitmap
 
-6. To run the examples, navigate to the examples folder. At this point it is
+Note: If you make any changes to any data/ files, you will need to manually delete the generated .html
+file in <build dir>/examples to force emscripten to regenerate the .data file.
+
```
connorjclark, line 107: This really confused me for a bit, so I added a note about it here.

connorjclark:
> Maybe with #1317 we won't need to explicit make copy_example_data.

Yup, this works!

README_sdl.txt
connorjclark: I can tell that this timer is not functioning correctly (I added a print statement in the loop and `al_get_timer_count(inc_counter)` is alway zero): https://github.com/connorjclark/allegro5/blob/f41f49d65c89fdb214b4dd49172b5adfbf16c4ec/demos/speed/main.c#L64-L64

could it be because it isn't registered to an event queue?

connorjclark:
FYI for future reference, in the GH UI the merge button has an option to "squash" all the commits and just take the PR title as the commit message. There should be an option in the repo settings to make this the default behavior. I think this is a pretty common setting (no idea why it isn't made the default...) as it separates the developing commit history of a PR from the main branch.

README_sdl.txt
SiegeLord: That demo is written very unidiomatically, as it's a pretty basic sourceport of the original written for Allegro 4. The issue is that it wasn't calling a "heartbeat" function everywhere. I added a small hack to get it half working, but it really needs a rewrite: https://github.com/liballeg/allegro5/commit/42fe8739a6dd286e84052ba927103f4edcdbb22b

PR #1318, opened 2022-03-07
Update ex_palette to build with emscripten

### LINKED ISSUE #1315: Can't get shader examples to run via emscripten
I wanted to see `ex_palette` running in the browser, but I get this error after building locally and opening in the browser:

```
Uncaught TypeError: Failed to execute 'attachShader' on 'WebGLRenderingContext': parameter 2 is not of type 'WebGLShader'
```

I noticed all the shader examples are not presented in the web demo page: https://liballeg.org/examples_graphics.html

I understand that some features of OpenGL are not able to be emulated through WebGL, but as long as you are mindful of the differences in feature sets shouldn't it be possible to build (with emscripten) an allegro5 app that uses shaders?

EDIT

I overlooked a few examples that indeed are working on the demo page, such as https://allegro5.org/examples/examples/ex_shader.html . I guess there's something about `ex_palette` that fails under webgl.

connorjclark:
Fixes #1315

Demo: https://tedious-porter.surge.sh/ex_palette.html

- WebGL only supports non-const array index access in shaders, which makes the approach to palette swapping done by `ex_palette` not portable via emscripten. Instead of using a uniform to store the palette, and updating/interpolating it on the CPU every cycle, I changed the program to instead pre-generate a texture with all the colors and added uniforms to do the interpolation on the GPU.
- A (smaller) necessary change was to specify the precision for the fragment shader–webgl requires this
- Added a neat feature–press P and the palette texture will be rendered (scaled up)
- Extracted the fragment shader source to a data file, and replaced the vertex shader with the provided allegro default

Please note I cannot build natively on my Mac (Unrelated to these changes, I can't get the build to work, first error being: `error: use of undeclared identifier 'ALLEGRO_TIMEOUT_SDL'`). I can attempt to build it on my Windows machine tomorrow (unless a reviewer shares that it worked for them).

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -81,10 +81,11 @@ set(DATA_SHADERS
     ex_prim_shader_pixel.hlsl
     ex_prim_shader_vertex.glsl
     ex_prim_shader_vertex.hlsl
-    ex_shader_multitex_pixel.glsl
-    ex_shader_multitex_pixel.hlsl
     ex_prim_wrap_pixel.glsl
     ex_prim_wrap_pixel.hlsl
+    ex_shader_multitex_pixel.glsl
+    ex_shader_multitex_pixel.hlsl
+    ex_shader_palette_pixel.glsl
     ex_shader_pixel.glsl
     ex_shader_pixel.hlsl
     ex_shader_vertex.glsl
@@ -212,7 +213,7 @@ if(SUPPORT_OPENGL AND NOT SUPPORT_OPENGLES)
     example(ex_opengl_pixel_shader ${IMAGE})
 endif()
 if(SUPPORT_OPENGL AND NOT IPHONE AND NOT ALLEGRO_RASPBERRYPI)
-    example(ex_palette ${IMAGE} ${COLOR} ${DATA_IMAGES})
+    example(ex_palette ${IMAGE} ${COLOR} ${DATA_IMAGES} ${DATA_SHADERS})
 endif()
 
 example(ex_font ${FONT} ${IMAGE} ${DATA_IMAGES})
--- a/examples/ex_palette.c
+++ b/examples/ex_palette.c
@@ -96,16 +96,6 @@ int main(int argc, char **argv)
    background = al_load_bitmap("data/bkg.png");
    /* Continue even if fail to load. */
 
-   // al_set_new_bitmap_flags(ALLEGRO_MEMORY_BITMAP);
-   // al_set_new_bitmap_flags(ALLEGRO_VIDEO_BITMAP);
-   // al_set_new_bitmap_format(ALLEGRO_PIXEL_FORMAT_ARGB_8888);
-   // al_set_new_bitmap_format(ALLEGRO_PIXEL_FORMAT_ANY_32_NO_ALPHA);
-   al_set_new_bitmap_format(ALLEGRO_PIXEL_FORMAT_ANY);
-   pal_bitmap = al_create_bitmap(255, 1);
-   al_set_target_bitmap(pal_bitmap);
-
-   al_set_target_backbuffer(display);
-
    /* Create 7 palettes with changed hue. */
    for (j = 0; j < 7; j++) {
       for (i = 0; i < 256; i++) {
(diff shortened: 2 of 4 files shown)


examples/CMakeLists.txt
```diff
@@ -81,10 +81,11 @@ set(DATA_SHADERS
...
     ex_prim_shader_vertex.glsl
     ex_prim_shader_vertex.hlsl
-    ex_shader_multitex_pixel.glsl
-    ex_shader_multitex_pixel.hlsl
     ex_prim_wrap_pixel.glsl
     ex_prim_wrap_pixel.hlsl
+    ex_shader_multitex_pixel.glsl
+    ex_shader_multitex_pixel.hlsl
+    ex_shader_palette_pixel.glsl
     ex_shader_pixel.glsl
```
connorjclark, line 89: I had a really old branch, so when I rebased I had a conflict here (it was simple). Anyway, before the rebase I saw these files were alphabetical (in my older checkout), and so after the rebase I again sorted alphabetically in my IDE. Looking at the diff now, obv it wasn't alpha sorted in master... anyway, I can revert the order changes if requested.

examples/ex_palette.c
```diff
@@ -115,45 +94,30 @@ int main(int argc, char **argv)
       }
    }
 
+   al_set_new_bitmap_format(ALLEGRO_PIXEL_FORMAT_ANY);
+   pal_bitmap = al_create_bitmap(255, 7);
```
connorjclark, line 98: OK, so I came across some very unexpected behavior here. Perhaps it is a bug? I'll explain.

I knew that allegro imposes a minimum texture size (16 for opengl. or maybe it is for any gfx backend?). So, originally I just had `16` here for the bitmap height, knowing I'd only use 7 rows and would need to divide by 15 in the fragment shader for the indexing (instead of 6, if the texture really were 7 pixels tall). All good, conceptually. But in practice, setting this value to `16` resulted in the entire screen being black.

I debugged via Spector.js and saw the texture in memory was padded in an unexpected way.

Here it is when you create the bitmap with height `7`:

image

And this is height `16`:

image

The bug is that I expected the additional padding that allegro would provide to go on the bottom and the right of the texture, leaving my "intended" texture in the top-left corner of the actual texture. The reality is this is reversed, resulting in the shader only ever getting black pixels (as the rest of the bitmap I'm not using it simply black).

I'm not sure if this misconception comes from my lack of familiarity with graphics pipelines (does padding in this manner make sense if you know more about how the system works?). And I'm sure changing this behavior is not a reasonable request. At least, it'd be nice if this behavior was documented [here](https://man.archlinux.org/man/community/allegro/al_create_bitmap.3.en#:~:text=If%20you%20try%20to%20create%20a%20bitmap%20smaller%20than%20this%2C%20this%20call%20will%20not%20fail%20but%20the%20returned%20bitmap%20will%20be%20a%20section%20of%20a%20larger%20bitmap%20with%20the%20minimum%20size.)

examples/ex_palette.c
```diff
@@ -82,31 +72,20 @@ int main(int argc, char **argv)
...
          float r, g, b, h, s, l;
-         r = (pal_hex[i] >> 16) / 255.0;
-         g = ((pal_hex[i] >> 8) & 255) / 255.0;
-         b = (pal_hex[i] & 255) / 255.0;
+         r = pal_hex[i] >> 16;
+         g = (pal_hex[i] >> 8) & 255;
+         b = pal_hex[i] & 255;
          
          al_color_rgb_to_hsl(r, g, b, &h, &s, &l);
-         if (j == 6) {
```
connorjclark, line 96: Oh, almost forgot about this. If you leave this code in, every odd value in the palette array will be all black! I didn't really dig into _why_ this was happening, I just deleted all the branches here and kept the interesting part, the hue shifting.

Looking at what I assume is the old version ( https://www.youtube.com/watch?v=Shk8GBCXQq4&ab_channel=EliasPschernig ) I do notice some differences (the corner sprites strobe, and there's a moment all the colors get really washed out) but IMO simpler is better here for this demo.

examples/ex_palette.c
```diff
@@ -193,31 +163,42 @@ int main(int argc, char **argv)
...
             al_draw_rotated_bitmap(bitmap,
                64, 64, s->x, s->y, s->angle, s->flags);
          }
          
          {
             float sc = 0.5;
-            al_set_shader_float_vector("pal", 3,
-               pals[(int)t % 20 > 15 ? 6 : 0], 256);
+            al_set_shader_float("pal_set_1", (int)t % 20 > 15 ? 6 : 0);
+            al_set_shader_float("pal_interp", 0);
```
connorjclark, line 186: I'd expect the corner sprites to swap palettes, but they remain the same colors... I'll have to look into this more tomorrow.

examples/ex_palette.c
connorjclark: Ah, I hadn't realized the hsl conversion functions expect rgb values to be 0-1, so this makes sense now.

Here it is with the original hsl shifting: https://unique-van.surge.sh/ex_palette.html

examples/ex_palette.c
connorjclark: It's because the 0th and. the 6th palettes were the same colors, both being hue shifted at equivalent values (0 and 360). I change the hue shifter to be not a factor of the period, and now it alters between two colors as expected.

connorjclark:
I was able compile allegro on my mac, and verified this works on OSX too.

examples/ex_palette.c
SiegeLord: I think what you're seeing here is that the texture is stored upside down. When the height is 7, it's padded on the bottom. When the height is 16 there's no padding, but the data is placed on the bottom because the texture is upside down.

SiegeLord:
This is great, thanks!

PR #1319, opened 2022-03-10
Fix emscripten build example for bash

connorjclark:
expanding an array in zsh (`echo "${USE_FLAGS}"` ) is different than in bash. zsh will expand the values as you expect, bash only the first item.

When I verified this command initially I was in my zsh terminal, and only later when trying to write a bash script using the same commands did I realize (but it took a long time) the commands weren't portable.

luckily both zsh and bash do the same sensible thing for `"${USE_FLAGS[*]}"`


SiegeLord:
Thanks!

PR #1322, opened 2022-03-26
Fix usage of SDL_Init for emscripten

### LINKED ISSUE #1321: How are examples built with emscripten such that sound works?
Although there is a working example here: https://allegro5.org/examples/examples/ex_saw.html I am not able to get a local build to work.

Following the instructions at https://github.com/liballeg/allegro5/blob/master/README_sdl.txt , I made a few tweaks during my attempts:

1) `-D WANT_OPENAL=ON` (seemed odd this would be disabled, as emscripten has an openal port)
2) `-I "$EM_CACHE/sysroot/include/AL"` (probably a hack.  altho emscripten has openal headers in its provided sysroot, they are within a folder and the compiled code expects them to not be: `#include <al.h>` vs `#include <AL/al.h>`

I can successfully build and load the HTML in the browser, but I only get `Generating saw wave...` in the output and nothing else.

connorjclark:
This patch fixes audio for emscripten.

I printed errors as SDL_GetError was called, and saw this:

image

Debugging told me the audio system was being initialized, but then shut down. Turns out that if any subsytem requested by SDL_Init cannot start, SDL will shutdown everything. So the haptic subsystem failing was problematic.

It fails because SDL configured the haptic subsytem to not build for emscripten (guess it's not supported).

This behavior in SDL_Init is a recent change. About 6 months ago SDL changed the init function to shutdown things when any requested subsystem failed to start: https://github.com/libsdl-org/SDL/commit/0e294e90ae6953777cc78b5d09dd5c8966dc96ab . Previously, it would just return -1, but allegro ignores the return code, so it didn't matter (not fully understanding why everything still worked even though every subsystem after the haptic would have not been initialized...shrug)

I confirmed that the audio demos now play sound in the browser when built locally, using the instructions found in the SDL readme.

Fixes #1321

### DIFF
--- a/src/sdl/sdl_system.c
+++ b/src/sdl/sdl_system.c
@@ -148,7 +148,7 @@ static ALLEGRO_SYSTEM *sdl_initialize(int flags)
    unsigned int sdl_flags = SDL_INIT_EVERYTHING;
 #ifdef __EMSCRIPTEN__
    // SDL currently does not support haptic feedback for emscripten.
-   sdl_flags -= SDL_INIT_HAPTIC;
+   sdl_flags &= ~SDL_INIT_HAPTIC;
 #endif
    if (SDL_Init(sdl_flags) < 0) {
       ALLEGRO_ERROR("SDL_Init failed: %s", SDL_GetError());
--- a/src/sdl/sdl_system.c
+++ b/src/sdl/sdl_system.c
@@ -144,7 +144,16 @@ static ALLEGRO_SYSTEM *sdl_initialize(int flags)
    ALLEGRO_SYSTEM_SDL *s = al_calloc(1, sizeof *s);
    s->system.vt = vt;
 
-   SDL_Init(SDL_INIT_EVERYTHING);
+   // TODO: map allegro flags to sdl flags.
+   unsigned int sdl_flags = SDL_INIT_EVERYTHING;
+#ifdef __EMSCRIPTEN__
+   // SDL currently does not support haptic feedback for emscripten.
+   sdl_flags -= SDL_INIT_HAPTIC;
+#endif
+   if (SDL_Init(sdl_flags) < 0) {
+      ALLEGRO_ERROR("SDL_Init failed: %s", SDL_GetError());
+      return NULL;
+   }
 
    _al_vector_init(&s->system.displays, sizeof (ALLEGRO_DISPLAY_SDL *));
 


connorjclark:
Aside: I was only able to start debugging when I added logging to SDL_SetError in my local emscripten ports build cache. I noticed that allegro doesn't provide a logging interface in its System driver, and only in one place does it ever call SDL_GetError (when making a window). Would it make sense to extend the system driver interface to include logging, and somehow wire SDL's logging infrastructure into allegro? Or is there already a way to get SDL logging from allegro that I'm overlooking?

src/sdl/sdl_system.c
```diff
@@ -144,7 +144,16 @@ static ALLEGRO_SYSTEM *sdl_initialize(int flags)
    ALLEGRO_SYSTEM_SDL *s = al_calloc(1, sizeof *s);
    s->system.vt = vt;
 
-   SDL_Init(SDL_INIT_EVERYTHING);
+   // TODO: map allegro flags to sdl flags.
+   unsigned int sdl_flags = SDL_INIT_EVERYTHING;
+#ifdef __EMSCRIPTEN__
+   // SDL currently does not support haptic feedback for emscripten.
+   sdl_flags -= SDL_INIT_HAPTIC;
```
SiegeLord, line 151: That looks very odd... I'd rather see this as `sdl_flags &= ~SDL_INIT_HAPTIC;`

SiegeLord:
Thanks for investigating this.

What do you mean by a logging interface? You want an Allegro function that one can call to add to a log? If SDL has something like `https://liballeg.org/a5docs/trunk/system.html#al_register_trace_handler`, then we could pass it a function that appends to logs (via `ALLEGRO_INFO` etc macros).

SiegeLord:
Something like:

```c
void log_to_allegro(void *userdata, int category, SDL_LogPriority priority, const char *message) {
  (void)userdata;
  const char* category_string;
  switch (category_string) {
    case SDL_LOG_CATEGORY_APPLICATION:
      category_string: "SDL_LOG_CATEGORY_APPLICATION",
      break;
    ...
  }
  switch (priority) {
    case SDL_LOG_PRIORITY_INFO:
      ALLEGRO_INFO("%s: %s\n", category_string, message);
      break;
    ...
  }
}

SDL_LogSetOutputFunction(log_to_allegro, NULL)
```

src/sdl/sdl_system.c
pedro-w: Is it possible to do this by trying to init with `SDL_INIT_EVERYTHING` and falling back to `SDL_INIT_EVERYTHING & ~SDL_INIT_HAPTIC` if it fails? The reason being that, if SDL ever does implement a haptic driver for Emscripten, it will work without having to change this code back.

connorjclark: @pedro-w Good idea. There is a Web api for [vibration](https://developer.mozilla.org/en-US/docs/Web/API/Vibration_API), so it's certainly possible that one day the haptic system will work. I'll try your backing-off strategy this evening.

connorjclark:
> What do you mean by a logging interface?

I don't have any particular ideas re: implementation, mainly just voicing an annoyance that whatever logging SDL is doing is not accessible to allegro / configurable. Sorry, I'm not being very constructive here :)

There does seem to be a way to set a callback function for logging in SDL: https://github.com/EddieRingle/SDL/blob/master/src/SDL_log.c#L361

src/sdl/sdl_system.c
connorjclark: There seemed to be some sound issues when I did the backoff approach. I'm not confident that SDL truly cleans up everything when it shutsdown its systems.

connorjclark, line 151:
```suggestion
   sdl_flags &= ~SDL_INIT_HAPTIC;
```

PR #1326, opened 2022-04-09
Allow SDL joysticks to be reconfigured after init

connorjclark:
Currently joysticks do not work with programs built with emscripten+SDL backend. You can verify this by trying to use a gamepad with the joystick events example [here](https://allegro5.org/examples/examples/ex_joystick_events.html).

The SDL joystick code reads the number of joysticks connected on init, keeping track of the number found in a `count` variable. If a new controller connects, SDL emits an event, but currently the event handling code _ignores all events_ if on init it found zero controllers.

On the web, the `gamepadconnected` event [only fires after the first input on the controller](https://developer.mozilla.org/en-US/docs/Web/API/Gamepad_API/Using_the_Gamepad_API). This means that on the web, SDL will have zero controllers on setup... unless you happen to be twiddling the controller inputs on setup. The example above initializes too quickly to see this behavior, but it was the first thing I noticed while debugging this issue in my (more setup-intensive) application: controllers only worked if I was twiddling the controller during setup.

The fix is:
1) Don't ignore events if allegro doesn't yet know about any controllers. After all, one of those events is "new controller connected"!
2) Call `sdl_reconfigure_joysticks` when a controller connects or disconnects

I confirmed this patch works for my application and for the above example:
image

### DIFF
--- a/src/sdl/sdl_joystick.c
+++ b/src/sdl/sdl_joystick.c
@@ -49,9 +49,6 @@ static SDL_Joystick *get_sdl(ALLEGRO_JOYSTICK *allegro)
 
 void _al_sdl_joystick_event(SDL_Event *e)
 {
-   if (count <= 0)
-      return;
-
    ALLEGRO_EVENT event;
    memset(&event, 0, sizeof event);
 
@@ -97,7 +94,7 @@ void _al_sdl_joystick_event(SDL_Event *e)
 static bool sdl_init_joystick(void)
 {
    count = SDL_NumJoysticks();
-   joysticks = calloc(count, sizeof * joysticks);
+   joysticks = count > 0 ? calloc(count, sizeof * joysticks) : NULL;
    int i;
    for (i = 0; i < count; i++) {
       joysticks[i].sdl = SDL_JoystickOpen(i);
@@ -131,6 +128,7 @@ static void sdl_exit_joystick(void)
    }
    count = 0;
    free(joysticks);
+   joysticks = NULL;
 }
 
 static bool sdl_reconfigure_joysticks(void)


beoran:
Thanks for this patch. I wonder what that count variable was supposed to do. Could you please confirm there are no unintended side effects?

connorjclark:
Nothing seems to have broken for me, with 0 joysticks or 1 connected. However, I only own a single gamepad so I can't test the more complex cases of multiple gamepads and them being reconfigured in an expected manner on connect/disconnects.

beoran:
That would be great, thanks. If you could buy or borrow a second joypad or joystick, that would even be better. They are relatively cheap these days.

SiegeLord:
I should be able to test with multiple controllers.

src/sdl/sdl_joystick.c
```diff
@@ -83,6 +82,7 @@ void _al_sdl_joystick_event(SDL_Event *e)
    }
    else if (e->type == SDL_JOYDEVICEADDED || e->type == SDL_JOYDEVICEREMOVED) {
       event.joystick.type = ALLEGRO_EVENT_JOYSTICK_CONFIGURATION;
+      sdl_reconfigure_joysticks();
```
connorjclark, line 85: looking at `ex_joystick_hotplugging` it seems the expectation is that programs should decide themselves to reconfigure. If so, should this line be removed (and any program that needs to just will call `al_reconfigure_joysticks` on this event themselves?)

connorjclark: Ah, so does `ex_joystick_events`. OK, this line probably doesn't belong here.

src/sdl/sdl_joystick.c
```diff
@@ -97,7 +94,7 @@ void _al_sdl_joystick_event(SDL_Event *e)
 static bool sdl_init_joystick(void)
 {
    count = SDL_NumJoysticks();
-   joysticks = calloc(count, sizeof * joysticks);
+   joysticks = count > 0 ? calloc(count, sizeof * joysticks) : NULL;
```
connorjclark, line 97: In my testing I got errors when going from 1 -> 0 controllers because although `joysticks` was being freed in `sdl_exit_joystick`, this line here would have a count of zero and I noticed `joysticks[0]` still had a pointer to (free'd) data. [`calloc`](https://en.cppreference.com/w/c/memory/calloc) has odd behavior when size is zero. To fix, manually set to NULL when size is zero.

Also, set `joysticks` to NULL after freeing for good measure.

connorjclark:
I confirmed that going from 0->1 and 1->0 controllers works as expected in the demos. @SiegeLord I've uploaded emscripten builds for your convenience:

https://comfortable-badge.surge.sh/ex_joystick_events.html

https://comfortable-badge.surge.sh/ex_joystick_hotplugging.html

SiegeLord:
Seems to work fine with two controller as well. Thanks for the change!

beoran:
Indeed, thank you very much for your extra tests and for 
contributing!

PR #1327, opened 2022-04-10
Set joystick button name for SDL game controllers

connorjclark:
SDL provides a function to map a button to a name, but only for SDL game controllers (not SDL joysticks). Allegro doesn't have a distinction between controllers/joysticks, which is fine. But when possible allegro can query SDL for the name of a button if supported.


SiegeLord:
Thanks!

PR #1330, opened 2022-04-20
Fix SDL mouse axis setter

connorjclark:
The correct indices for the z and w mouse values are [2 and 3](https://github.com/liballeg/allegro5/blob/e8457df9492158f2c5f41676e70f0b98c7f5deeb/src/mousenu.c#L173), but the sdl mouse implementation uses 1 and 2. This results in buggy ("sticky") mouse wheel behavior when using SDL (a call to `al_set_mouse_z(0)` would never really clear the z value, so mouse events always kept their previous value given no new wheel input even though the user cleared it)


SiegeLord:
Thanks!

PR #1333, opened 2022-05-10
Fix random garbage at edges of glyph rendering

check-switch-26:
Bugfix - regression bug caused by commit 61b7310

In some cases, glyphs render with thin garbage lines at edges.  This is caused by uninitialized adjacent regions in the glyph cache bitmap.

Fixed by doubling glyph border to two pixels, while using a one-pixel border for rendering.

### DIFF
--- a/addons/ttf/ttf.c
+++ b/addons/ttf/ttf.c
@@ -334,8 +334,8 @@ static unsigned char *alloc_glyph_region(ALLEGRO_TTF_FONT_DATA *data,
 
    /* Copy a displaced pointer for the glyph. */
    return (unsigned char *)data->page_lr->data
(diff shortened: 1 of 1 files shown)


beoran:
Would it not be better to let alloc_glyph_region clear the bitmap in stead?

check-switch-26:
> Would it not be better to let alloc_glyph_region clear the bitmap in stead?

`alloc_glyph_region()` as originally written locks (and clears) only the region of the bitmap that will be used to save the new glyph.  I have tried to keep my changes as minimal as possible so as to keep the intent clear and to not disrupt any original design decisions.

There likely are some opportunities for optimization in the `ttf.c` module, but I'm not sure how aggressively that should be pursued.

SiegeLord:
Thanks!

Issue #555, opened 2016-01-09
al_mangled_main linking issues with bundles


SiegeLord:
See https://www.allegro.cc/forums/reply/615967

peterhull90:
My thoughts (taken from the thread linked above...)

Configuring CMake on OS X gives this warning:
~~~~
CMake Warning (dev):
  Policy CMP0042 is not set: MACOSX_RPATH is enabled by default.  Run "cmake
  --help-policy CMP0042" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  MACOSX_RPATH is not specified for the following targets:

   allegro
   allegro_acodec
   allegro_audio
   allegro_color
   allegro_dialog
   allegro_font
   allegro_image
   allegro_main
   allegro_memfile
   allegro_physfs
   allegro_primitives
   allegro_ttf
   allegro_video
~~~~

Using @rpath with these libraries should make it possible to use them in bundles just by copying them to the Frameworks subdirectory,

except `allegro_main` - if this can be made into a static library it will be able to link to `_al_mangled_main` in the user's code even after symbol stripping.

peterhull90:
Further to this - the rpath stuff is already covered in #532. CMake never works the way I expect it to.

trentg:
One thing I'm seeing as I try this is homebrew is naming libraries like liballegro.5.1.dylib as symlinks to liballegro5.1.12.dylib and so on... once these go into a Framework, the libraries that depend on liballegro.5.1.dylib (example, don't remember if this is one of the actual dependencies) are going to fail because in the .app/Contents/Frameworks there is only liballegro.5.1.12.dylib...

I'm sure it's possible to get this to work using install_name_tool. However, after 30 minutes of changing names I still have a lot left to change. I suggest looking at otool -l and otool -L output of libs and executables. With the example given on allegro.cc, there's a ton of missing links.

On OS X, you have to build libraries a certain way if you want to use them in bundles. Then they won't work right standalone. CMake has things like BUILD_WITH_INSTALL_RPATH and  INSTALL_RPATH that help with this. I set the former to on and the latter to "@executable_path/" (you could add ../Frameworks) when I build dynamic libraries. Then, when they're linked to everything is already set to go. I don't know how you'd do it with homebrew though.

SiegeLord:
Would things be better if we generated frameworks instead of dylibs?

fatcerberus:
[...]

PR #1334, opened 2022-05-11
Don't use -flat_namespace in LINK_FLAGS on macOS

danielnachun:
In Homebrew we have started auditing for the usage of flat namespaces in shared libraries, and have found that they are being used in the libraries built by Allegro 5.  The usage of flat namespaces is considered deprecated in modern versions of macOS (https://developer.apple.com/forums//thread/689991), and it can lead to symbol conflicts if a dynamically linked binary ends up linking to Allegro and another library with identically named symbols.  

These flags were added in 2008 in https://github.com/liballeg/allegro5/commit/646aa94fc52640e79864cb471eff8535d02fd719 and have probably remained as a historical artifact.  On modern macOS systems the proper flag should be `-undefined dynamic_lookup` instead, and Allegro builds successfully with these changes in Homebrew CI: https://github.com/Homebrew/homebrew-core/pull/101176.

### DIFF
--- a/cmake/Common.cmake
+++ b/cmake/Common.cmake
@@ -158,7 +158,7 @@ function(add_our_library target framework_name sources extra_flags link_with)
     if(MACOSX)
         set_target_properties(${target}
             PROPERTIES
-            LINK_FLAGS "-flat_namespace -undefined suppress"
+            LINK_FLAGS "-undefined dynamic_lookup"
             )
     endif(MACOSX)
 


beoran:
Thanks for the contribution. This seems to be ok, but does it also work for the monolith build?

danielnachun:
I tested the build with `-DWANT_MONOLITH=ON` added and it successfully builds the monolith library.  I also confirmed that the small test program we use links to the monolith library and executes successfully.  So I think the monolith build is good to go.

pedro-w:
I'd be happy to see this merged - assuming #555 is no longer a problem

danielnachun:
I believe this PR may also fix https://github.com/liballeg/allegro5/issues/555, as I encountered that exact error when I tried to just delete the extra `LINK_FLAGS` entirely instead of replacing `-flat_namespace -undefined suppress` with `-undefined dynamic_lookup`.  If you look at the error posted there further down:
```
dlopen(libs/lib1-opengl/libOpenGL.so, 8): Symbol not found: __al_mangled_main
Referenced from: /nfs/zfs-student-6/users/tmack/.brew/opt/allegro/lib/liballegro_main.5.2.dylib
Expected in: flat namespace
in /nfs/zfs-student-6/users/tmack/.brew/opt/allegro/lib/liballegro_main.5.2.dylib
```
it appears to be a result of incorrectly using the flat namespace rather than the two-level namespace and telling the linker to ignore missing symbols rather than searching for them dynamically.

beoran:
If the monolith is also ok then we can probably merge this.

SiegeLord:
Looks good. Thanks for the change!

Issue #1335, opened 2022-05-15
Errors building with SDL backend on windows/mac


connorjclark:
The SDL backend works for building with emscripten, but I can't get it to build on Windows or Mac.

```
cmake .. -DALLEGRO_SDL=on -DSDL2_INCLUDE_DIR="/c/Program Files/SDL2/include/"
cmake --build . -t ex_draw --config RelWithDebInfo

output:

C:\Users\cjamc\code\allegro5\src\sdl\sdl_thread.c(49,21): warning C4047: 'function': 'const size_t' differs in levels of indirection from '_AL_THREAD *' [C:\Users\cjamc\code\allegro5\build\allegro.vcxproj]
C:\Users\cjamc\code\allegro5\src\sdl\sdl_time.c(39,24): error C2065: 'ALLEGRO_TIMEOUT_SDL': undeclared identifier [C:\Users\cjamc\code\allegro5\build\allegro.vcxproj]
C:\Users\cjamc\code\allegro5\src\sdl\sdl_time.c(39,37): error C2065: 'timeout_sdl': undeclared identifier [C:\Users\cjamc\code\allegro5\build\allegro.vcxproj]
C:\Users\cjamc\code\allegro5\src\sdl\sdl_time.c(39,54): warning C4047: '=': 'int' differs in levels of indirection from 'void *' [C:\Users\cjamc\code\allegro5\build\allegro.vcxproj]
C:\Users\cjamc\code\allegro5\src\sdl\sdl_thread.c(49,21): warning C4024: 'SDL_CreateThreadWithStackSize': different types for formal and actual parameter 3 [C:\Users\cjamc\code\allegro5\build\allegro.vcxproj]
C:\Users\cjamc\code\allegro5\src\sdl\sdl_time.c(39,54): error C2106: '=': left operand must be l-value [C:\Users\cjamc\code\allegro5\build\allegro.vcxproj]
C:\Users\cjamc\code\allegro5\src\sdl\sdl_thread.c(49,21): warning C4113: 'pfnSDL_CurrentEndThread' differs in parameter lists from 'pfnSDL_CurrentBeginThread' [C:\Users\cjamc\code\allegro5\build\allegro.vcxproj]
C:\Users\cjamc\code\allegro5\src\sdl\sdl_time.c(40,15): error C2065: 'timeout_sdl': undeclared identifier [C:\Users\cjamc\code\allegro5\build\allegro.vcxproj]
[...]

PR #1336, opened 2022-05-15
Get SDL backend working on OSX

connorjclark:
ref #1335

This patch unsets `ALLEGRO_MACOSX` if the SDL backend is requested. Some conditions controlling GL header inclusion had to be changed to a direct platform `__APPLE__` check, and a few `MACOSX` cmake checks had to change to `ALLEGRO_MACOSX`.

I'm able to build and run the `ex_draw` example with this change, with and without the SDL backend. Although with SDL you must disable the native dialog addon manually (`-DWANT_NATIVE_DIALOG=off`). Should that be done in the build by default iff `ALLEGRO_SDL`?


connorjclark:
Should I also try to get travis building SDL on osx in this PR, or a new one?

SiegeLord:
Thanks. We need CI for SDL on every system at some point.

PR #1343, opened 2022-07-03
look in win10 sdk for DirectX libraries

connorjclark:
This fixes the [requirement](https://github.com/liballeg/allegro5/blob/d25c4cff7621db614e5236a1646f5d76309ca73f/CMakeLists.txt#L835) of needing the [deprecated DirectX SDK](https://www.microsoft.com/en-us/download/details.aspx?id=6812) installed to build Allegro 5 on Windows. This patch allows finding a subset of the DirectX libraries/headers from the Windows 10 SDK.

The Windows 10 SDK includes partial support for legacy DirectX 9 libraries, enough that building allegro 5 with default parameters would work if only Allegro's CMakeLists.txt knew where to locate the Windows 10 SDK installation. This will still result in a non-fatal warning `"D3DX9 not found. You will not be able to compress or decompress textures on the GPU"`, but at least things will build and run. (the "extended" direct x libraries are excluded from the win10 sdk)

_________

Sidenote: This _does not_ do anything about the current state of legacy DirectX 9 being used. I'm not very familiar Windows development esp. re: DirectX to know how involved this would be, but: is it feasible to upgrade to non-deprecated graphics libraries for Windows?

Some potentially helpful links:

- https://stackoverflow.com/a/56658216/2788187
- https://walbourn.github.io/living-without-d3dx/

### DIFF
--- a/cmake/FindDirectX.cmake
+++ b/cmake/FindDirectX.cmake
@@ -26,6 +26,27 @@ check_winsdk_root_dir("[HKEY_LOCAL_MACHINE\\\\SOFTWARE\\\\Microsoft\\\\Microsoft
 check_winsdk_root_dir("[HKEY_LOCAL_MACHINE\\\\SOFTWARE\\\\Microsoft\\\\Windows Kits\\\\Installed Roots;KitsRoot]")
 check_winsdk_root_dir("[HKEY_LOCAL_MACHINE\\\\SOFTWARE\\\\Microsoft\\\\Windows Kits\\\\Installed Roots;KitsRoot81]")
 
+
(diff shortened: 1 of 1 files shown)


SiegeLord:
Seems reasonable. Thanks!

> Sidenote: This does not do anything about the current state of legacy DirectX 9 being used. I'm not very familiar Windows development esp. re: DirectX to know how involved this would be, but: is it feasible to upgrade to non-deprecated graphics libraries for Windows?

I don't know.

beoran:
Thanks for this, although in the long run we will have to upgrade do newer versions of DirectX.

PR #1344, opened 2022-07-04
Fix Sub-Menu Behavior On MacOS

NewCreature:
On MacOS, clicking a sub-menu causes the menu to close and triggers an `ALLEGRO_EVENT_MENU_CLICK`. We expect the sub-menu to open immediately when clicking the item and do nothing if the sub-menu is already open. By disabling the menu item's action when the item being added is a sub-menu we get the correct behavior when clicking on a sub-menu.


SiegeLord:
Thanks!

PR #1345, opened 2022-07-05
Use NSColor controlTextColor for textlog on osx

connorjclark:
Minor bug fix - when using the dark theme on Mac, the textlog text is black on a gray background so is unreadable.


SiegeLord:
Thanks!

PR #1351, opened 2022-07-19
SDL: touch: Don't use fingerId as an array index

dos1:
SDL fingerIds are not guaranteed to look in any specific way and can vary greatly between platforms. Without this, we could go out-of-bounds, overwriting chunks of memory and causing random crashes that were quite easy to hit on some of SDL backends, including X11 one.


src/sdl/sdl_touch.c
```diff
@@ -138,7 +160,12 @@ void _al_sdl_touch_input_event(SDL_Event *e)
       return;
    }
 
-   int touchId = e->tfinger.fingerId;
+   int touchId = find_touch_state_index_with_id(e->tfinger.fingerId);
```
SiegeLord, line 163: Maybe this variable should be called something else to reduce confusion? `touch_idx`?

dos1: Makes sense, updated.

PR #1352, opened 2022-07-20
docs: Clarify the difference between minimal and regular shader

dos1:
Both kinds of shaders fully support all alpha blending modes, which isn't clear from the current wording. Currently, the only difference between those shaders is that minimal GLSL shader doesn't implement alpha testing, so make it explicit in the docs.


SiegeLord:
Thanks!

PR #1359, opened 2022-08-01
Implement Support for Filter Patterns On MacOS

NewCreature:
Allegro's native file chooser under MacOS currently lacks support for using patterns to filter which files are shown in the dialog. The MacOS file select dialog supports filtering by file extensions, which is probably the most common use of pattern matching in file select dialogs.

This PR adds support for filtering by file extension to the native dialog add-on under MacOS.

### DIFF
--- a/addons/native_dialog/osx_dialog.m
+++ b/addons/native_dialog/osx_dialog.m
@@ -17,6 +17,39 @@ void _al_shutdown_native_dialog_addon(void)
 }
 #pragma mark File Dialog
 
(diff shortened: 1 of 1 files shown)


SiegeLord:
This looks good, thanks! Ideally we'd use [allowedContentTypes](https://developer.apple.com/documentation/appkit/nssavepanel/3566857-allowedcontenttypes) which supports MIME types too (and isn't deprecated), but this is a good start regardless.

PR #1360, opened 2022-08-02
Two additions (scaled font draw + change wrap) + one bug fix

VirtuallTek:
1. Add _scaled variants for font drawing functions (super easy, doesn't change previous code, just add).
2. Add support for changing wrap anytime. This adds a private flag for bitmaps to notify the driver about changes (no additional storage used). _al_bitmap_wrap_changed returns and resets the flag, applying just one time after the change. For D3D9 there's nothing to do as everytime the render state is updated, wrap does too. For OpenGL I'm checking that in glsl_set_shader_sampler (only works for programmable pipeline).
3. Fix the bug introduced in the latest release. al_set_audio_stream_playmode is always returning false.


include/allegro5/internal/aintern_bitmap.h
```diff
@@ -14,6 +14,11 @@ extern "C" {
 #endif
 
 typedef struct ALLEGRO_BITMAP_INTERFACE ALLEGRO_BITMAP_INTERFACE;
+   
+enum ALLEGRO_BITMAP_INTERNAL_FLAGS
```
SiegeLord, line 18: I'd rather not use flags for this, we could just use a boolean variable in the bitmap.

That said, I'd rather move this into the driver. I.e. add a new vtable entry for `wrap_changed` that is called in `al_set_bitmap_wrap` and store the boolean in `ALLEGRO_BITMAP_EXTRA_OPENGL`, since we only need to do the lazy update there.

src/opengl/ogl_shader.c
```diff
@@ -282,6 +285,9 @@ static bool glsl_set_shader_sampler(ALLEGRO_SHADER *shader,
 
    texture = bitmap ? al_get_opengl_texture(bitmap) : 0;
    glBindTexture(GL_TEXTURE_2D, texture);
+   
+   if (_al_bitmap_wrap_changed(bitmap))
```
SiegeLord, line 289: There's one more place for this in ogl_draw.c:357

SiegeLord (changes requested):
Thanks, sorry for the delay in review.

I have a bunch of implementations comments on the wrapping change, so, if you prefer, not to do that, I can drop those changes from this PR and do them myself.

addons/font/allegro5/allegro_font.h
```diff
@@ -85,6 +85,8 @@ ALLEGRO_FONT_FUNC(ALLEGRO_FONT *, al_create_builtin_font, (void));
 
 ALLEGRO_FONT_FUNC(void, al_draw_ustr, (const ALLEGRO_FONT *font, ALLEGRO_COLOR color, float x, float y, int flags, ALLEGRO_USTR const *ustr));
 ALLEGRO_FONT_FUNC(void, al_draw_text, (const ALLEGRO_FONT *font, ALLEGRO_COLOR color, float x, float y, int flags, char const *text));
+ALLEGRO_FONT_FUNC(void, al_draw_ustr_scaled, (const ALLEGRO_FONT *font, ALLEGRO_COLOR color, float x, float y, float scaleX, float scaleY, int flags, ALLEGRO_USTR const *ustr));
```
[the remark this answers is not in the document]
VirtuallTek: Done.

src/opengl/ogl_shader.c
VirtuallTek: Done.

include/allegro5/internal/aintern_bitmap.h
VirtuallTek: Yeah. I made that way for simplicity and easy implementation, so no need for extra storage and bigger code changes like adding a new entry in the vtable.

src/opengl/ogl_draw.c
```diff
@@ -355,6 +358,9 @@ static void ogl_flush_vertex_cache(ALLEGRO_DISPLAY *disp)
 #endif
       }
       glBindTexture(GL_TEXTURE_2D, disp->cache_texture);
+      
+      if (_al_bitmap_wrap_changed(bitmap))
+         _ogl_update_wrap(bitmap);
```
check-switch-26, line 363: This addition is incorrect.  It doesn't compile.  `bitmap` is not defined.

addons/font/text.c
```diff
@@ -109,6 +109,42 @@ void al_draw_text(const ALLEGRO_FONT *font,
...
+	ASSERT(ustr);
+
+	if (flags & ALLEGRO_ALIGN_CENTRE) {
+		/* Use integer division to avoid introducing a fractional
+		 * component to an integer x value.
+		 */
+		x -= font->vtable->text_length(font, ustr) / 2;
+	}
+	else if (flags & ALLEGRO_ALIGN_RIGHT) {
+		x -= font->vtable->text_length(font, ustr);
```
check-switch-26, line 128: This computation is incorrect.  The length should have the scale applied.

addons/font/font.c
```diff
@@ -189,6 +189,36 @@ static int color_render_char(const ALLEGRO_FONT* f,
...
+	int h = f->vtable->font_height(f);
+	ALLEGRO_BITMAP *g;
+
+	g = _al_font_color_find_glyph(f, ch);
+	if (g) {
+		int width = al_get_bitmap_width(g);
+		int height = al_get_bitmap_height(g);
+		w = width * sw;
+		al_draw_tinted_scaled_bitmap(g, color, 0, 0, width, height,
+			x, y + ((float)h - height) / 2.0f, w, height * sh, 0);
```
check-switch-26, line 212: I believe that the vertical placement is incorrect here.  The `height` should have the scaling applied.  I suggest:
`y + ((float)h - height * sh) / 2.0f`

check-switch-26:
I reviewed the changes for the scaled text rendering and added comments.

Additionally, I note that the implementation is incomplete.  Routines for scaling have not been added to `bmfont.c` and `addons/ttf/ttf.c`

Also, documentation has not been added to `docs\src\refman\font.txt`

check-switch-26:
I recognize the usefulness of `al_draw_ustr_scaled()`, especially for bitmap fonts.

An easier way to implement it would be to use a transform for the scaling and calling `al_draw_ustr()`.  This would alleviate the need to touch every font type implementation.

PR #1363, opened 2022-08-11
Audio: functions for checking if voice/mixer has anything attached

TRDario:
- Adds two new functions (al_mixer_has_attachments, al_voice_has_attachments) that allow checking if the mixer/voice has anything attached to it.
- Corrects al_get_mixer_attached: it previously behaved like current al_mixer_has_attachments, contrary to what the docs say it's supposed to do.
- Updated the docs to match.

### DIFF
--- a/addons/audio/allegro5/allegro_audio.h
+++ b/addons/audio/allegro5/allegro_audio.h
@@ -324,6 +324,7 @@ ALLEGRO_KCM_AUDIO_FUNC(ALLEGRO_MIXER_QUALITY, al_get_mixer_quality, (const ALLEG
 ALLEGRO_KCM_AUDIO_FUNC(float, al_get_mixer_gain, (const ALLEGRO_MIXER *mixer));
 ALLEGRO_KCM_AUDIO_FUNC(bool, al_get_mixer_playing, (const ALLEGRO_MIXER *mixer));
 ALLEGRO_KCM_AUDIO_FUNC(bool, al_get_mixer_attached, (const ALLEGRO_MIXER *mixer));
+ALLEGRO_KCM_AUDIO_FUNC(bool, al_get_mixer_has_attached, (const ALLEGRO_MIXER *mixer));
 ALLEGRO_KCM_AUDIO_FUNC(bool, al_set_mixer_frequency, (ALLEGRO_MIXER *mixer, unsigned int val));
 ALLEGRO_KCM_AUDIO_FUNC(bool, al_set_mixer_quality, (ALLEGRO_MIXER *mixer, ALLEGRO_MIXER_QUALITY val));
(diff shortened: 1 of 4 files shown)


pedro-w:
I like the idea but I wonder if we can think a better names - I didn't find al_get_mixer_attached vs al_get_mixer_has_attached to be intuitive (maybe just me but I know I'd have to look them up every time!). Also, suppose al_get_mixer_attached returns false, then we call al_attach_XXX_to_mixer, then  al_get_mixer_attached will return true - that seems logical to me that we've attached something and then attached is true. English vocab is inadequate here, isn't it?

Also, I've forgotten so can someone remind me, what's the purpose of ALLEGRO_UNSTABLE and should these new functions be guarded by that flag?

TRDario:
@pedro-w  Yeah I'm not sure about the names either, just the first thing I thought of, would definitely be open to a better suggestion.

ALLEGRO_UNSTABLE refers to API that could easily change or be removed in future versions. I think that if we put these functions there wouldn't be much to change (or much of a reason to remove).

TRDario:
Maybe al_get_XXX_has_attachments()? Would help differentiate it from al_get_XXX_attached() while not bring too wordy.

SiegeLord:
`al_XXX_has_attachments` would be my preference. Existing APIs of this form are `al_clipboard_has_text`, `al_ustr_has_prefix`, `_al_pixel_format_has_alpha`.

pedro-w:
So,
* `al_get_mixer_attached`
* `al_mixer_has_attachments`
* `al_voice_has_attachments`

?
That's clearer, as far as I am concerned.

SiegeLord:
`al_get_mixer_attached` should probably have been `al_is_mixer_attached`... that's why we have `ALLEGRO_UNSTABLE`... to avoid this in the future...

TRDario:
I agree with SiegeLord's suggestion. Also, does that mean I should mark them as unstable?

SiegeLord:
For this PR, you don't have to. We'll adjust before making a release if these names feel risk... I don't think they do though.

pedro-w:
We could add `al_is_mixer_attached` as an alias for `al_get_mixer_attached` and deprecate the latter, or just live with it.

I was looking at other audio APIs (pipewire, XAudio2, Apple AudioUnits, ...) to see if there was any precedents but I gave up, sorry. I think if anything they tend to explicitly mention 'input' or 'output' so it might be `al_mixer_has_inputs` etc.

TRDario:
Maybe make the old format a macro for the new one like what `al_convert_bitmaps` did?

SiegeLord:
Thanks for the contribution! Like I said, before 5.2.9 is released, we an rework the names or make this unstable if there's any concern.

PR #1364, opened 2022-08-13
Make al_destroy_event_queue(NULL) do nothing

TRDario:
Most if not all other al_destroy_XXX functions do nothing if NULL is passed to them, while destroying a null event queue currently causes an assert to fail or simply throws an error on runtime. Unless there's a specific reason for this I'm not aware of, I think having it behave like the other destroy functions would be preferable.


SiegeLord:
Good idea. Could you adjust the docs to say `This function does nothing if the queue argument is NULL.`?

TRDario:
Should I mark it somehow to signify it's since 5.2.9?

SiegeLord:
Thanks!

PR #1368, opened 2022-08-15
Docs entries for new functions

TRDario:
It seems I forgot to include the docs entries for al_can_set_keyboard_leds and al_can_set_mouse_cursor_position.


SiegeLord:
Thanks! I was going to add them myself at some later point, but I appreciate you circling back and adding them yourself!

Issue #1341, opened 2022-06-07
Possible errors in voc.c


nick7inc:
allegro5/addons/acodec/voc.c

In ALLEGRO_SAMPLE *_al_load_voc_f(ALLEGRO_FILE *file)
1. buffer = al_realloc(buffer, sizeof(buffer) + bytestoread); 
Looks bad. I have no idea how new size are calculated here.  
2. Memory leak in code, the memory for AL_VOC_DATA allocates twice:  1st - in _al_load_voc_f() body, then address of pointer overwrites after voc_open() call.

beoran:
Thanks for reporting this. A PR with a fix would be most welcome.

nick7inc:
I am not very strong in programming (in git-specific things I almost zero), so I can only advise.
To fix bug with memory leak just remove allocation from _al_load_voc_f() body:
```
ALLEGRO_SAMPLE *_al_load_voc_f(ALLEGRO_FILE *file)
{
 // I prefer in-place declaration, so commenting thins:
  // AL_VOC_DATA *vocdata;
   ALLEGRO_SAMPLE *sample = NULL;
   size_t pos = 0; /* where to write in the buffer */
   size_t read = 0; /*bytes read during last operation */
   char* buffer;

   size_t bytestoread = 0;
   bool endofvoc = false;

// Removing (commenting) unnecessary allocation causes memory leak:
   // vocdata = al_malloc(sizeof(AL_VOC_DATA));
  // memset(vocdata, 0, sizeof(*vocdata));
   /*
    * Open file and populate VOC DATA, then create a buffer for the number of
    * samples of the frst block.
    * Iterate on the following blocks till EOF or terminator block
    */
   AL_VOC_DATA *vocdata = voc_open(file);
   if (!vocdata) return NULL; 
```

nick7inc:
With this line:
```
buffer = al_realloc(buffer, sizeof(buffer) + bytestoread);
```
It is incorrect - **sizeof(buffer)** - as **buffer** is not fixed-size (stack) variable (looks like copy-paste from elsewhere), so this startment returns size of pointer, not a buffer itself. It is requires to use additional variable **size_t buffer_size;** to store current size of buffer before allocation. I am doing back-port of this function for Allegro4 library and by library, so I use this kink of code **before every next block reading occurs** :
```
if (bytestoread + pos > buffer_size)
                {
                  buffer_size = bytestoread + pos;
                  if (buffer)
[...]

PR #1370, opened 2022-08-16
Misc. fixes

TRDario:
- (#1341) Fixes a memory leak when loading .voc samples.
- (#1341) Fixes incorrect buffer scaling during continuation block loading (previously kept adding to the size of the pointer).
- Bumps the early return for NULL vertex/index buffers up above a few functions that can throw errors.
- Adds "Since 5.2.9" to the docs I sent earlier.


SiegeLord:
Thanks!

PR #1372, opened 2022-08-31
TTF bugfixes - al_get_glyph_width(), al_get_glyph_dimenstions(), al_get_glyph()

check-switch-26:
- calling `al_get_glyph_width()` on a non-existent codepoint would crash with a stack overflow if a fallback font is defined
- value returned by `al_get_glyph_width()` was off by two pixels (this is a regression bug caused by a previous commit)
- `al_get_glyph_dimensions()` would return negative width or height for zero-width or zero-height glyphs
- data returned by `al_get_glyph()` was inconsistent with al_get_glyph_dimensions() for zero-width glyphs


SiegeLord:
Thanks!

EdgarReynaldo:
This is almost worth building binaries over. 🥇

PR #1378, opened 2022-10-16
Fix improper display boundary for cursors on OSX

### LINKED ISSUE #1282: ALLEGRO_EVENT_MOUSE_ENTER/LEAVE_DISPLAY triggers incorrectly on MacOS after resize.
After resizing a display by dragging the edge or corner, the events `ALLEGRO_EVENT_MOUSE_ENTER_DISPLAY` and `ALLEGRO_EVENT_MOUSE_LEAVE_DISPLAY` trigger incorrectly under MacOS.

The issue can be reproduced easily by running `ex_mouse_events`, making the window larger using the window edges to resize it, then moving the mouse within the window. The events are triggered when the mouse cursor reaches the area where the mouse would have entered or left the display before it was resized.

connorjclark:
`reshape` is called whenever the window's NSView changes size, and its job is to set the cursor tracking area. However, it was checking [NSOpenGLContext currentContext] existed before updating the tracking area. For some reason it was nil by the time `al_resize_display` is called.

The logic for updating the tracking area doesn't seem to need this context, so it should be safe to delete this guard.

ex_mouse_events is updated to resize the display. Without this patch, such a change to this example presents the bug: the ALLEGRO_EVENT_MOUSE_ENTER_DISPLAY / ALLEGRO_EVENT_MOUSE_LEAVE_DISPLAY events are fired only for the lower-left quadrant of the display.

Fixes #1282

### DIFF
--- a/examples/ex_mouse_events.c
+++ b/examples/ex_mouse_events.c
@@ -75,6 +75,11 @@ int main(int argc, char **argv)
       abort_example("Error creating display\n");
    }
 
+   // Resize the display - this is to excercise the resizing code wrt.
+   // the cursor display boundary, which requires some special care on some
+   // platforms (such as OSX).
+   al_resize_display(display, 640*1.5, 480*1.5);
+
    al_hide_mouse_cursor(display);
 
    cursor = al_load_bitmap("data/cursor.tga");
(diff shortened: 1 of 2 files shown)


pedro-w:
I did a bit of digging - that check on `currentContext` originally guarded some GL calls, so that made sense; over time these seem to have wandered off somewhere else (I couldn't find the commit where that happened!) so I agree it's probably not needed now. 
However I wonder why `currentContext` is nil here, is it a thread thing?

SiegeLord:
Seems reasonable to me, thanks!

PR #1379, opened 2022-10-18
Fix incorrect int-as-pointer bug in sdl_acknowledge_resize

connorjclark:


connorjclark (commented):
Upgrading to a stricter compiler caught this error.

Note: I did not test this.

pedro-w:
Seems remarkable that wasn't picked up before - I thought allegro did already use quite strict compilation flags!

SiegeLord:
Thanks! There's also an instance in the android backend... I'll fix that one.

PR #1384, opened 2022-11-15
Allow passing const arrays to spline functions

TRDario:
The two spline functions currently require a mutable pointer to be passed despite not actually modifying it in any way, meaning a pointless cast is required when using them with const arrays, at least in c++. None of the other primitive functions need a mutable pointer, so I assume this is just an oversight.


SiegeLord:
For these types of changes, we need to be concerned about ABI backwards compatibility. I think in this case, changing the constness of a parameter does not affect either the common calling conventions or the name mangling, so this seems safe.

Thanks!

PR #1397, opened 2022-12-14
set freeimage_initialized to true

allefant:
I know nothing about the freeimage addon - but this seems to fix crashes I get in debug mode.


SiegeLord:
Wow!

PR #1404, opened 2023-02-01
Try fixing some oddities with Shift + Alt on Windows.

### LINKED ISSUE #1348: Pressing SHIFT when holding ALT on windows emits a nil key event
I found this bug because there's a command in my application that does something when SHIFT and ALT are held down. It works when first pressing SHIFT then holding ALT too, but not the other way around.

From debugging, I found that the second key event received when doing ALT then SHIFT is zero'd out and only has the ALT bit enabled in `modifiers`:

```
{type=11 source=0x00007ffe3bcf6d40 {allegro-5.2.dll!ALLEGRO_KEYBOARD the_keyboard} {...} timestamp=341.56358236244910 ...}
    type: 11
    source: 0x00007ffe3bcf6d40 {allegro-5.2.dll!ALLEGRO_KEYBOARD the_keyboard} {...}
    timestamp: 341.56358236244910
    display: 0x000001eed147f840 {...}
    keycode: 0
    unichar: 0
    modifiers: 4
    repeat: false
```

My expectation is that either `modifiers` would have the SHIFT bit enabled, or that (also?) `keycode` is set to the value for the shift key.

Perhaps related: #713

SiegeLord:
There were two issues.

1. Holding Shift + tapping Alt would send spurious Control(!?) key up messages. I can reliably identify them by checking whether the repeat(aka previous) flag is set. MSDN says it always should be 1, but it's 0 for these weird events.

2. Holding Alt + tapping Shift would set weird KEY_CHAR events. This evidently can be fixed by sending the scan code to our event decoder from WM_SYSKEYDOWN messages. I'm not sure why we didn't before.

Fixes #1348

### DIFF
--- a/src/win/wwindow.c
+++ b/src/win/wwindow.c
@@ -711,9 +711,10 @@ static LRESULT CALLBACK window_callback(HWND hWnd, UINT message,
       }
       case WM_SYSKEYDOWN: {
          int vcode = wParam;
+         int scode = (lParam >> 16) & 0xff;
          bool extended = (lParam >> 24) & 0x1;
          bool repeated  = (lParam >> 30) & 0x1;
-         _al_win_kbd_handle_key_press(0, vcode, extended, repeated, win_display);
+         _al_win_kbd_handle_key_press(scode, vcode, extended, repeated, win_display);
          break;
       }
       case WM_KEYDOWN: {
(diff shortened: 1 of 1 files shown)


src/win/wwindow.c
```diff
@@ -731,7 +732,13 @@ static LRESULT CALLBACK window_callback(HWND hWnd, UINT message,
...
          int scode = (lParam >> 16) & 0xff;
          bool extended = (lParam >> 24) & 0x1;
-         _al_win_kbd_handle_key_release(scode, vcode, extended, win_display);
+         bool previous = (lParam >> 30) & 0x1;
+
+         /* The docs say that previous should always be 1, but it's not in practice.
+          * The events with previous = 0 seem malformed? I can get them reliably by
+          * pressing + holding Shift and then tapping Alt. */
+         if (previous)
+            _al_win_kbd_handle_key_release(scode, vcode, extended, win_display);
```
connorjclark, line 741: I couldn't find anything suggesting that `previous == 0` is possible, very strange. Just ignoring these bad events seems alright.

PR #1408, opened 2023-02-13
Repair audio playmode return value

codylico:
This commit modifies two parts of the `al_set_audio_stream_playmode` function:
- The return boolean is actually returned rather than discarded.
- The loop-without-feeder prevention is re-enabled for playmode `ALLEGRO_PLAYMODE_LOOP`. Now only audio streams created by `al_load_audio_stream` should allow the LOOP playmode.


SiegeLord:
Whoops! Thanks for finding and fixing this!

PR #1412, opened 2023-02-24
fix resize on macos

### LINKED ISSUE #1410: al_draw_line coordinates are off when resizing a window on MacOS
Tried the Allegro examples on my old work Mac and one thing I noticed is that ex_resize looks off. For example:

https://user-images.githubusercontent.com/390841/220150126-83e35261-6386-4155-824a-ad98d5d5c59a.mov

I have not investigated yet but ex_resize2 works fine so it's not resizing in general that's broken, but maybe something with al_draw_line? I checked the source of ex_resize and I don't see anything obvious but seemingly the al_draw_line commands assume a larger display than what it is.

Model Name:	MacBook Pro
Model Identifier:	MacBookPro16,1
Processor Name:	8-Core Intel Core i9
System Version:	macOS 12.6 (21G115)

allefant:
This fixes #1410 (the line coordinates, but unfortunately not the red flicker).

The issues I believe is that acknowledge_resize_display_win_main_thread gets run on the main thread and so setup_gl inside of it does not work properly as the main thread can't access the TLS variables from the thread with the user's main() function.

I moved setup_gl(d); out of that function and just call it from both acknowledge_resize_display_win and resize_display_win instead, on whichever thread those are called from. It seems to fix both resizing a window with the handle as well as using al_resize_display to resize it.

The reason resizing is not completely broken without this PR is that calls to al_draw_bitmap can also repair the opengl viewport so in most cases the viewport would only be wrong for a frame or two after a resize but then eventually things would work again.

I also moved another call of _al_ogl_setup_gl to after [context update] which seems to make more sense and removed a superfluous _al_ogl_resize_backbuffer call (it gets called inside _al_ogl_setup_gl anyway). And I added the display dimensions to some of the log messages.

### DIFF
--- a/src/macosx/osxgl.m
+++ b/src/macosx/osxgl.m
@@ -504,7 +504,7 @@ void _al_osx_mouse_was_installed(BOOL install) {
          event.display.width = NSWidth(content);
          event.display.height = NSHeight(content);
          _al_event_source_emit_event(es, &event);
-         ALLEGRO_INFO("Window finished resizing");
+         ALLEGRO_INFO("Window finished resizing %d x %d\n", event.display.width, event.display.height);
       }
       _al_event_source_unlock(es);
    }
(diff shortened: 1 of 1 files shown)


SiegeLord:
Nice! Thanks!

MarkOates:
After pulling this change, the fullscreen mode seems to have broken. I made a (brief) issue: https://github.com/liballeg/allegro5/issues/1415.

I haven't had a chance to investigate other than confirm (through a `git bisect`) that it happened after this change.

PR #1416, opened 2023-03-04
missed a spot that needs setup_gl in the last PR

### LINKED ISSUE #1415: Fullscreen not scaling correctly after #1412 [MacOS]
After pulling and building `mater`, `ALLEGRO_FULLSCREEN` is not scaling correctly since https://github.com/liballeg/allegro5/pull/1412.  Probably related to a retina display?

Expected:

expected

Actual:

actual

Windowed mode appears to work fine:

windowed

[edit: replaced large filesize images]

allefant:
This should fix #1415. In the last PR I had moved setup_gl out of resize_display_win_main_thread() so it can run from the calling thread and missed that set_display_flag also calls it for fullscreen windows.

### DIFF
--- a/src/macosx/osxgl.m
+++ b/src/macosx/osxgl.m
@@ -2401,6 +2401,7 @@ static bool set_display_flag(ALLEGRO_DISPLAY *display, int flag, bool onoff)
 
    if (!win)
       return false;
+   bool __block need_setup_gl = false;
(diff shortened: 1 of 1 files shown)


SiegeLord:
Nice!

MarkOates:
Works! thank you! 🙌

PR #1417, opened 2023-03-07
Sdl examples macos

allefant:
In ex_glext we use ALLEGRO_MACOSX to check whether we need to use Apple's OpenGL include path - but this fails when building the SDL version under MacOS as ALLEGRO_SDL will be defined and not ALLEGRO_MACOSX. Using \_\_APPLE__ for MacOS specific code is suggested here: https://developer.apple.com/library/archive/documentation/Porting/Conceptual/PortingUnix/compiling/compiling.html

### DIFF
--- a/examples/ex_glext.c
+++ b/examples/ex_glext.c
@@ -12,7 +12,7 @@
 
 #include "common.c"
 
-#ifdef __APPLE_CC__
+#ifdef __APPLE__
 #include <OpenGL/glu.h>
 #else
 #include <GL/glu.h>
--- a/examples/ex_glext.c
+++ b/examples/ex_glext.c
@@ -12,7 +12,7 @@
 
 #include "common.c"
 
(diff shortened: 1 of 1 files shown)


MarkOates:
I'm not familiar with how the Allegro defines are setup and used, but I would think we should prefer to use the `ALLEGRO_MACOSX` whenever possible.  Is there a reason why `ALLEGRO_MACOSX` is turned off in the context of `ALLEGRO_SDL`? E.g. It's [set to `0` in the `CMakeLists` here](https://github.com/liballeg/allegro5/blob/e8ac98003744361b2c7e848a31093daff2e74c73/CMakeLists.txt#L369)?

allefant:
Yeah, I think we should change things so ALLEGRO_MACOSX can be defined even when building with SDL. But not sure the location of the OpenGL headers should be tied to it.

PR #1422, opened 2023-03-13
Android: map predictive back gestures to ALLEGRO_KEY_BACK in a backwards compatible way

### LINKED ISSUE #1419: Back button support on Android
Previously, [we could intercept the back button](https://github.com/liballeg/allegro5/blob/4aa54e6c994af21bc63d8b593673ab3df62390f8/examples/ex_android.c#L169) on Android using `ALLEGRO_KEY_BACK` (`KeyEvent.KEYCODE_BACK`). Considering the [introduction of back gestures](https://developer.android.com/guide/navigation/predictive-back-gesture), that solution is obsolete and will break. How can we best handle the back button or gesture in Allegro? Do we need to modify the API? Maybe we can trigger a back event?

Captura de tela_2023-03-10_22-08-49

alemart:
Google is introducing [predictive back gestures](https://developer.android.com/guide/navigation/predictive-back-gesture) in Android 13. This is a breaking change that affects Allegro. Previously, we could intercept back events using `KeyEvent.KEYCODE_BACK`. This is no longer the case.

Any Allegro apps that target API level 33 and that opt in to the predictive back gestures no longer work. I tested this myself and saw that key up/key down events with `ALLEGRO_KEY_BACK` are no longer triggered when a back gesture is performed. Since [predictive back gestures will become mandatory](https://developer.android.com/guide/navigation/predictive-back-gesture#update-custom), we must update the code.

This update handles the back button / gesture on API level 33+ in a backwards compatible way. Back events are intercepted using the new `OnBackInvokedCallback` API. These events are mapped to `ALLEGRO_KEY_BACK` as before, so that [existing Allegro apps will continue to work as usual](https://github.com/liballeg/allegro5/blob/4aa54e6c994af21bc63d8b593673ab3df62390f8/examples/ex_android.c#L169), despite the breaking change in Android.

This fixes #1419 and updates the target API level to 33.

### DIFF
--- a/android/gradle_project/allegro/build.gradle
+++ b/android/gradle_project/allegro/build.gradle
@@ -1,10 +1,10 @@
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion 26
+    compileSdkVersion 33
     defaultConfig {
         minSdkVersion 15
-        targetSdkVersion 26
+        targetSdkVersion 33
         versionCode 1
         versionName "1.0"
     }
--- a/android/gradle_project/allegro/src/main/java/org/liballeg/android/AllegroActivity.java
+++ b/android/gradle_project/allegro/src/main/java/org/liballeg/android/AllegroActivity.java
@@ -25,6 +25,9 @@ import android.view.InputDevice;
 import java.util.Vector;
 import android.os.Build;
 import android.view.View;
+import android.view.KeyEvent;
+import android.window.OnBackInvokedCallback;
+import android.window.OnBackInvokedDispatcher;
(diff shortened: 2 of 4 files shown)


alemart:
I just received an e-mail from Google Play stating that starting August 31st, 2023, new apps and app updates must target API level 33. More info [here](https://developer.android.com/google/play/requirements/target-sdk).

Captura de tela_2023-03-21_21-32-39

SiegeLord:
Seems like it works. Thanks a lot for investigating and fixing this!

Issue #1420, opened 2023-03-12
`al_install_joystick` can sometimes freeze [MacOS]


MarkOates:
## Problem

If one or multiple joysticks are already connected when `al_install_joystick()` is called, Allegro will _sometimes_ get stuck in an infinite loop.

## Location

I haven't been able to do much research other than [localize the problem to this loop in `al_install_joystick()`](https://github.com/liballeg/allegro5/blob/master/src/macosx/hidjoy.m#L588-L599), and try it with one and two joysticks.  I also haven't been able to isolate the conditions to reproduce the problem reliably, but I'd say it happens 1 in 4 times or so.

It seems like the loop is doing polling, waiting/expecting for callbacks to finish and the number of `joysticks` (allegro local variable) to match the number of `devices` (os-local count of devices matching the "joystick" criteria).

pedro-w:
What are the values of `count` and `size` when it gets stuck (are either of them 'correct'?)
Also I wonder what the purpose of that loop is - the joysticks will be added eventually (and only a little while later, surely) and that's only the same as if you'd plugged one in after the start of the game.

MarkOates:
When infinite looping, `size` (the number of allegro-known devices) is one less than `count` (number of os-connected devices).  The number of os-connected devices seems to be correct.

> Also I wonder what the purpose of that loop is - the joysticks will be added eventually (and only a little while later, surely) and that's only the same as if you'd plugged one in after the start of the game.

They could also be removed, too!

I'm not sure the design is correct.  My expectation would be to:
1) Enumerate and create the number of connected devices
2) Register callbacks to the OS for newly connected/disconnected devices

The current design seems to mix the two, leaving step 1 after step 2, and then polling for connection events to catch any that were missed.  Thoughts are welcome... I'm still poking around on it.

MarkOates:
Some progress has been made. I'll share a [WIP branch here](https://github.com/liballeg/allegro5/compare/master...MarkOates:allegro5:fix-joystick-infinite-loop?diff=unified) that manually enumerates connected joysticks on initialization (rather than waiting for callbacks), but it is still incomplete. I believe the manual enumeration needs to not compete with the existing callbacks that can also occur.

MarkOates:
Ok, I've got a PR up that seems to fix the issue [here](https://github.com/liballeg/allegro5/pull/1423).

MarkOates:
Issue appears to be be fixed with https://github.com/liballeg/allegro5/pull/1423 🎉

PR #1423, opened 2023-03-16
[macOS] Fix Joystick Infinite Loop

MarkOates:
## Problem

Issue https://github.com/liballeg/allegro5/issues/1420

Sometimes when `al_install_joystick()` is called while one or multiple joysticks are already connected, the system can get stuck in an infinite loop.  This happens on the example programs and in my code as well about 1 in 6 times.

I isolated the problem to [this loop](https://github.com/liballeg/allegro5/blob/master/src/macosx/hidjoy.m#L588-L599), which performs polling, waiting for the number of Allegro-recognized devices to match the number of os-recognized devices, while waiting for callbacks.

## Solution

Rather than polling in a loop to check if the devices may have been created, this will create the devices manually and remove the polling loop. At any time, the same callbacks may occur and try to create the devices, but those devices will not be created if they already exist (and we will not potentially be stuck in an infinite loop while doing so). The `device_add` process is protected behind a mutex.

### DIFF
--- a/src/macosx/hidjoy.m
+++ b/src/macosx/hidjoy.m
@@ -298,16 +298,8 @@ static void osx_joy_generate_configure_event(void)
    _al_generate_joystick_event(&event);
 }
 
-static void device_add_callback(
-   void *context,
-   IOReturn result,
-   void *sender,
-   IOHIDDeviceRef ref
-) {
-   (void)context;
-   (void)result;
-   (void)sender;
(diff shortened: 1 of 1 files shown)


SiegeLord:
Seems alright, and my tests show that it works. Thanks for the change!

MarkOates:
Sweeet 🎉

PR #1425, opened 2023-03-17
prevent crash when using al_get_mouse_state in the SDL port

allefant:
The mouse->display can be NULL in several cases like when the mouse is outside the window and we shouldn't crash in that case. I'm not sure the behavior of the mouse coordinates for al_get_mouse_state is well defined when the mouse is outside of any Allegro window so I'm not too worried about coordinates losing scaling for now.

### DIFF
--- a/src/sdl/sdl_mouse.c
+++ b/src/sdl/sdl_mouse.c
@@ -195,7 +195,9 @@ static bool sdl_set_mouse_axis(int which, int value)
 static void sdl_get_mouse_state(ALLEGRO_MOUSE_STATE *ret_state)
 {
    int x, y, i;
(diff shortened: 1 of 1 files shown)


SiegeLord:
It appears that the state is roughly the sum of mouse dx/dys, so it just remains at 0, 0 before mouse enters the display and the last value it was after exiting it. I think the right thing to do here is store the ratio inside the mouse structure and update it during events.

allefant:
Like this? The ratio is only calculated when the display changes and stored (and initially 1.0). The crash is still fixed with this. I think ideally ALLEGRO_MOUSE_STATE should also contain the display the mouse is currently in but that would be a separate issue.

SiegeLord:
Thanks!

PR #1427, opened 2023-03-29
Use correct pointer type. This fixes flood fill crash on 64-bit builds.

NewCreature:
@connorjclark sent a patch in to my [Allegro Legacy](https://github.com/NewCreature/Allegro-Legacy) project which I found should also be applied to Allegro to fix a crash in the `floodfill()` routine when compiled on 64-bit systems.


SiegeLord:
Nice! Thanks.

PR #1428, opened 2023-04-07
implement create_mouse_cursor for SDL

connorjclark:


SiegeLord:
Nice, seems to work!

I wonder if SDL has system cursors, since otherwise you can never reset the cursor back.

PR #1429, opened 2023-04-09
implement sdl_get_monitor_dpi

allefant:
al_get_monitor_dpi returns just 0 with the SDL driver, this PR implements it.


SiegeLord:
Merged via 23ad1bf. Thanks!

PR #1432, opened 2023-04-17
Fixes #1431.

### LINKED ISSUE #1431: al_set_display_menu() causes extra height to be added to display on Windows and breaks al_resize_display().
While working on the Windows version of one of my current projects, I noticed when I called `al_resize_display()` the display would continually gain a bit of height with each call. After a bit of testing, I found when using `al_set_display_menu()` the display gains an extra 20 pixels in height. This extra height was reported by `al_get_display_height()` which I was using as a basis for the `height` argument in my call to `al_resize_display()`.

Looking at Allegro's code I see that there was some code added in the past to make the window automatically resize to keep your set dimensions (6bc85ccf6070a24c9681c529e9e39a5b044290cc). Later, some other code was added to fix issues with the window size that seems to conflict with this earlier change (75fdcaa2613ddf9530a33ee49141bf71b213d257).

~~Putting `al_set_config_value(al_get_system_config(), "compatibility", "automatic_menu_display_resize", "false");` in my code before doing anything with menus, fixes the issue.~~

I made a little test program that demonstrates the issue using `al_set_display_menu()` (see below). The issue ultimately stems from the `display->extra_resize_height` variable being set in `al_set_display_menu()` which is then used by `al_resize_display()` to modify the resize height to include the menu height. The code related to `extra_resize_height` was only used on Windows. If 75fdcaa2613ddf9530a33ee49141bf71b213d257 does indeed remove the need for the original workaround, that code should probably be removed. As it is, the out-of-the-box experience with `al_set_display_menu()` and `al_resize_display()` on Windows is broken.

```
#include <allegro5/allegro.h>
#include <allegro5/allegro_font.h>
#include <allegro5/allegro_native_dialog.h>
#include <stdio.h>

int main(int argc, char * argv[])
{
	ALLEGRO_DISPLAY * display = NULL;
	ALLEGRO_MENU * top_menu = NULL;
	ALLEGRO_MENU * menu = NULL;
	ALLEGRO_FONT * font = NULL;
	ALLEGRO_EVENT_QUEUE * queue = NULL;
	ALLEGRO_EVENT event;
	int height[3] = {0};
	int i;

	if(!al_init())
	{
		goto fail;
	}
	if(!al_install_keyboard())
	{
		goto fail;
	}
	if(!al_init_native_dialog_addon())
	{
		goto fail;
	}
	if(!al_init_font_addon())
	{
		goto fail;
	}
	display = al_create_display(800, 600);
	if(!display)
	{
		goto fail;
	}
	height[0] = al_get_display_height(display);
	font = al_create_builtin_font();
	if(!font)
	{
		goto fail;
	}
	queue = al_create_event_queue();
	if(!queue)
	{
		goto fail;
	}
	al_register_event_source(queue, al_get_display_event_source(display));
	al_register_event_source(queue, al_get_keyboard_event_source());
	menu = al_create_menu();
	if(!menu)
	{
		goto fail;
	}
	al_append_menu_item(menu, "Item 1", 0, 0, NULL, NULL);
	al_append_menu_item(menu, "Item 2", 1, 0, NULL, NULL);
	top_menu = al_create_menu();
	if(!top_menu)
	{
		goto fail;
	}
	al_append_menu_item(top_menu, "Main", 2, 0, NULL, menu);

	while(1)
	{
		al_get_next_event(queue, &event);
		if(event.type == ALLEGRO_EVENT_DISPLAY_CLOSE)
		{
			break;
		}
		else if(event.type == ALLEGRO_EVENT_KEY_DOWN)
		{
			height[1] = al_get_display_height(display);
			if(!top_menu)
			{
				top_menu = al_remove_display_menu(display);
			}
			else
			{
				al_set_display_menu(display, top_menu);
				top_menu = NULL;
			}
			height[2] = al_get_display_height(display);
		}
		al_rest(1.0 / 60.0);
		al_clear_to_color(al_map_rgb(0, 0, 0));
		for(i = 0; i < 3; i++)
		{
			al_draw_textf(font, al_map_rgb(255, 255, 255), 0, i * al_get_font_line_height(font), 0, "%d", height[i]);
		}
		al_flip_display();
	}
	al_destroy_font(font);
	al_destroy_event_queue(queue);
	al_destroy_display(display);

	return 0;

	fail:
	{
		if(top_menu)
		{
			al_destroy_menu(top_menu);
		}
		if(menu)
		{
			al_destroy_menu(menu);
		}
		if(queue)
		{
			al_destroy_event_queue(queue);
		}
		if(font)
		{
			al_destroy_font(font);
		}
		if(display)
		{
			al_destroy_display(display);
		}
		return -1;
	}
}
```

NewCreature:
This commit fixes #1431 by eliminating the `extra_resize_height` variable and its use in the code. As mentioned in #1431, 6bc85cc implemented a change to the resizing behavior which conflicted with part of 75fdcaa, leading to extra height being added to the display when attaching a menu.

### DIFF
--- a/addons/native_dialog/menu.c
+++ b/addons/native_dialog/menu.c
@@ -792,7 +792,6 @@ bool al_set_display_menu(ALLEGRO_DISPLAY *display, ALLEGRO_MENU *menu)
       _al_vector_delete_at(&display_menus, i);
 
       if (automatic_menu_display_resize && menu_height > 0) {
-         display->extra_resize_height = 0;
          al_resize_display(display, al_get_display_width(display), al_get_display_height(display));
       }
    }
@@ -826,7 +825,6 @@ bool al_set_display_menu(ALLEGRO_DISPLAY *display, ALLEGRO_MENU *menu)
          /* Temporarily disable the constraints so we don't send a RESIZE_EVENT. */
          bool old_constraints = display->use_constraints;
          display->use_constraints = false;
-         display->extra_resize_height = menu_height;
          al_resize_display(display, al_get_display_width(display), al_get_display_height(display));
          display->use_constraints = old_constraints;
       }
--- a/include/allegro5/internal/aintern_display.h
+++ b/include/allegro5/internal/aintern_display.h
@@ -158,11 +158,6 @@ struct ALLEGRO_DISPLAY
 
    /* Issue #725 */
    bool use_constraints;
-
-   /* On Windows, menus take up the display's height so whenever doing a
-    * manual resize we need to add this number to the height for things
-    * to work correctly. See issue #860. */
-   int extra_resize_height;
 };
 
 int  _al_score_display_settings(ALLEGRO_EXTRA_DISPLAY_SETTINGS *eds, ALLEGRO_EXTRA_DISPLAY_SETTINGS *ref);
--- a/src/display.c
+++ b/src/display.c
@@ -71,7 +71,6 @@ ALLEGRO_DISPLAY *al_create_display(int w, int h)
    display->max_w = 0;
    display->max_h = 0;
    display->use_constraints = false;
-   display->extra_resize_height = 0;
 
    display->vertex_cache = 0;
    display->num_cache_vertices = 0;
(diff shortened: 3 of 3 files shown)


NewCreature:
After running this code on another machine for testing, it seems this fix won't be adequate. On that machine, the call to `al_set_display_menu()` does resize the window, but the drawable area is not pushed down below the menu. So rendering to `(0, 0)` renders underneath the menu and there is a blank space at the bottom of the window.

Calls to `al_resize_display()` or resizing with the window handles makes the issue go away. I can't reproduce this problem on the original machine.

NewCreature:
Alright, been looking into this for a while now. The menu overlapping the drawable area bug is separate from the window size bug, so I think this patch is good for fixing #1431 still.

The bug I encountered with the menu overlapping the drawable area happens only with OpenGL on Windows and with or without this patch. I'll make a separate issue for this bug for now.

SiegeLord:
Nice, thanks for the investigation and fix!

Issue #1435, opened 2023-04-24
Native file selector soft locks on MacOSX 10.6.


NewCreature:
When opening a native file selector dialog on MacOSX 10.6, the program soft locks. The output on the terminal is:
```
... -[NSCFArray objectAtIndexedSubscript:]: unrecognized selector set to instance at 0x...
... -[NSCFArray objectAtIndexedSubscript:]: unrecognized selector set to instance at 0x...
```

pedro-w:
We don't use that function `objectAtIndexedSubscript:` directly but as I understand it, it's used to implement the direct indexing of `NSArray`, e.g. `an_array[index]` becomes a call to this. See for example
https://github.com/liballeg/allegro5/blob/23ad1bf8a700255aa0ed174c4f356e1946b473af/addons/native_dialog/osx_dialog.m#L22-L28
If you replace `work_array[i]` with `[work_array objectAtIndex: i]` does it fix it? (there may be other instances of direct indexing, too)

NewCreature:
Yes, that did the trick.

NewCreature:
Yeah, I had a feeling when I encountered the soft lock it had something to do with my file type filter patch.

PR #1436, opened 2023-04-25
Fixes #1434 and #1435.

### LINKED ISSUE #1434: Support for MacOSX < 10.12 Broken
One of my current projects targets MacOSX 10.6 as the minimum system requirement. The program crashed upon trying to load it up on a 10.6 machine. Looking into it, the crash resulted from `clock_gettime()` not being implemented until MacOSX 10.12.

In my research I found that Apple implemented `clock_gettime()` in MacOSX 10.12 but doesn't prevent code that targets older versions of MacOSX from linking it in. Since it is a weak link, it will go unnoticed until the program tries to call that function.

My initial thought was to switch to the 10.11 SDK, but Allegro won't compile with that SDK. I was able to implement an alternative to `clock_gettime()` (see below) that fixes the crashing issue with my program. This alternative would need to be called when building for MacOSX and the minimum required version is set to < 10.12.

```
int _internal_clock_gettime(clockid_t clock_id, struct timespec* t)
{
    mach_timebase_info_data_t timebase;
    mach_timebase_info(&timebase);
    uint64_t time;
    time = mach_absolute_time();
    double nseconds = ((double)time * (double)timebase.numer)/((double)timebase.denom);
    double seconds = ((double)time * (double)timebase.numer)/((double)timebase.denom * 1e9);
    t->tv_sec = seconds;
    t->tv_nsec = nseconds;
    return 0;
}
```

NewCreature:
When targeting MacOSX versions < 10.12, running the binary on any version of MacOSX < 10.12 resulted in a crash. The first commit adds a code path for using `mach_absolute_time()` in place of `clock_gettime()` when building for MacOSX with a target version < 10.12.

Also, when opening the native file dialog on MacOSX 10.6 the program soft locks. This bug was introduced in b5bfadaf2ee15b0675054abd817056b172a5e236. The second commit fixes this bug by using `objectAtIndex` instead of direct indexing.

### DIFF
--- a/addons/native_dialog/osx_dialog.m
+++ b/addons/native_dialog/osx_dialog.m
@@ -22,7 +22,7 @@ static NSArray * remove_mime_types(NSArray * array)
    NSMutableArray * work_array = [array mutableCopy];
 
    for(NSInteger i = work_array.count - 1; i >= 0; i--){
-      if([work_array[i] rangeOfString:@"/"].location != NSNotFound){
+      if([[work_array objectAtIndex: i] rangeOfString:@"/"].location != NSNotFound){
          [work_array removeObjectAtIndex: i];
       }
    }
--- a/src/unix/utime.c
+++ b/src/unix/utime.c
@@ -32,13 +32,33 @@ ALLEGRO_STATIC_ASSERT(utime,
 struct timespec _al_unix_initial_time;
 
 
+/* implement clock_gettime() for MacOS < 10.12 */
+#if defined(ALLEGRO_MACOSX) && MAC_OS_X_VERSION_MIN_REQUIRED < 101200
+#include <mach/mach_time.h>
+int _internal_clock_gettime(clockid_t clock_id, struct timespec* t)
+{
+    mach_timebase_info_data_t timebase;
+    mach_timebase_info(&timebase);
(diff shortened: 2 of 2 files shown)


NewCreature:
Sorry, this PR causing issues with timers. I'm closing this. I'm not sure where to go from here on #1434.

pedro-w:
What were the issues  with timers?

NewCreature:
They periodically jump forward, causing incessant stuttering.

pedro-w:
Don't know why that would be. Could you use gettimeofday() ?

It seems to be what Apple use in their libc implementation, see
https://opensource.apple.com/source/Libc/Libc-1439.40.11/gen/clock_gettime.c.auto.html

NewCreature:
Yeah, `gettimeofday()` would work, but I didn't think to use it because it was replaced in a previous commit with a mention that it's deprecated.

NewCreature:
Might be a good idea to use `gettimeofday()` if the target is MacOS < 10.12. At least for now this will fix the issue. If `gettimeofday()` is ever removed completely, maybe an availability check can be added to the build script if Allegro still supports targeting versions of MacOS < 10.12 at that point in time.

PR #1437, opened 2023-04-30
Fix for #1435.

NewCreature:
When opening the native file dialog on MacOSX 10.6, the program soft locks. This bug was introduced in b5bfadaf2ee15b0675054abd817056b172a5e236. This commit fixes this bug by using `objectAtIndex` instead of direct indexing.


SiegeLord:
Thanks!

PR #1438, opened 2023-05-01
Fix for #1434.

NewCreature:
Use `gettimeofday()` instead of `clock_gettime()` when building for MacOSX with a target version < 10.12.


SiegeLord:
Thanks!

Issue #1443, opened 2023-05-28
OpenGL 3.x on macos


j-w-c-b:
I was trying to build [xu4](https://github.com/xu4-engine/u4) on my mac this weekend, and ran into some issues getting it to run. It uses some GLSL shaders to smoothly scale some of the 80s graphics, but they use modern GL features which the display returned by allegro5 doesn't support on macos.

From what I can tell the issues are

* passing a hint in the pixel format to get opengl 3
* updating the allegro default shaders to not use deprecated features
* injecting a `#version xxx` tag when compiling the shaders; macos is strict about that for the 3.2+ profiles---they must match the version used by the display.

I'll create a PR with my changes for this, but I started the weekend knowing nothing about shaders, so it might not be quite right, but is close.

alemart:
Won't there be interoperability issues across platforms if a `#version` line is hard-coded into `glsl_attach_shader_source()` in src/opengl/ogl_shader.c as in #1444 ? Wouldn't it be preferable if the application sorted this out instead of Allegro?

https://github.com/liballeg/allegro5/blob/20df29dd32008ed86479644069fe95b01a83a7b3/src/opengl/ogl_shader.c#L127-L142

SiegeLord:
This is now fixed. Note that there is a remaining issue where the primitives don't draw. One way to fix it is via #1003, but something intermediate could be an option too.

PR #1444, opened 2023-05-28
Add support for OpenGL 3 displays on macos

j-w-c-b:
When creating a mac display wth the ALLEGRO_OPENGL_3_0 flag, set the pixel attribute NSOpenGLPFAOpenGLProfile to NSOpenGLProfileVersion3_2Core to enable modern OpenGL. This conflicts with NSOpenGLPFAWindow, so that is dropped.

Fix checking for ALLEGRO_RENDER_METHOD in the required options to enable NSOpenGLPFAAccelerated. I'm not sure if that makes a difference, but the old code was using the option as a value and not a bit offset.

Duplicate the shaders in GLSL 3.3+ syntax (adding a _gl3 suffix to the static variables). Check the shader language version, and select the _gl3 versions when using a version != 1.20 on macos.

When compiling the shaders on macos, check for the supported shader version. If != 1.20, insert a #version line based on the value returned by the OpenGL context.

[https://github.com/liballeg/allegro5/issues/1443]

### DIFF
--- a/src/macosx/osxgl.m
+++ b/src/macosx/osxgl.m
@@ -767,8 +767,6 @@ static void osx_set_opengl_pixelformat_attributes(ALLEGRO_DISPLAY_OSX_WIN *dpy)
       // Take over the screen.
        *a = NSOpenGLPFAScreenMask; a++;
        *a = CGDisplayIDToOpenGLDisplayMask(dpy->display_id); a++;
-   } else {
-      *a = NSOpenGLPFAWindow; a++;
    }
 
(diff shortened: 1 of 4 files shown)


SiegeLord:
This got merged (with changes) via #1644. Sorry for the insane delay, this PR didn't initially work for me and I completely misdiagnosed the problem, so I kinda dropped it. Recently I revisited it, found the real issue (Allegro's bitmap drawing routines were not using VAOs on MacOS due a mistaken assumption).

PR #1446, opened 2023-05-30
fix bad merge

allefant:
Somehow merging in https://github.com/liballeg/allegro5/pull/1429 went terribly wrong, not sure exactly how that did happen but I blame github :P


SiegeLord:
...probably my fault, but I've no idea how it happened either.

PR #1448, opened 2023-06-14
Implement `al_get_joystick_name()` on MacOS.

NewCreature:
This commit implements retrieving the name of joystick devices on MacOS. This was implemented in a previous commit (0828211c7d1b1e4647efe228c5b6e08e49f16bac) and was apparently accidentally reverted in a subsequent commit (e18a22b73d9d50ec94e5ed710eabf692c1c8fade).

I plan to go back over the original commit to see if any other changes from that commit need to be re-implemented.

### DIFF
--- a/src/macosx/hidjoy.m
+++ b/src/macosx/hidjoy.m
@@ -63,6 +63,7 @@ typedef struct {
    CONFIG_STATE cfg_state;
    ALLEGRO_JOYSTICK_STATE state;
    IOHIDDeviceRef ident;
+   char * name;
 } ALLEGRO_JOYSTICK_OSX;
 
 static IOHIDManagerRef hidManagerRef;
@@ -757,10 +758,25 @@ static bool reconfigure_joysticks(void)
    return ret;
 }
 
-// FIXME!
 static const char *get_joystick_name(ALLEGRO_JOYSTICK *joy_)
 {
-   (void)joy_;
+   ALLEGRO_JOYSTICK_OSX *joy = (ALLEGRO_JOYSTICK_OSX *)joy_;
+   CFStringRef str;
+
+   str = IOHIDDeviceGetProperty(joy->ident, CFSTR(kIOHIDProductKey));
+   if (str) {
+      CFIndex length = CFStringGetLength(str);
+      CFIndex maxSize = CFStringGetMaximumSizeForEncoding(length, kCFStringEncodingUTF8) + 1;
+      if (joy->name) {
+         free(joy->name);
+      }
+      joy->name = (char *)malloc(maxSize);
+      if (joy->name) {
+         if (CFStringGetCString(str, joy->name, maxSize, kCFStringEncodingUTF8)) {
+            return joy->name;
+         }
+      }
+   }
    return "Joystick";
 }
 


pedro-w:
This was a long time ago but I think
https://github.com/liballeg/allegro5/commit/0828211c7d1b1e4647efe228c5b6e08e49f16bac
was not reverted accidentally, it was because of a problem reported by
Trent (see comments at the bottom)

That PR also had a change to the code to get the element names from MacOS
(`get_element_name`) - looking at it now there may be a more robust way of
doing it (analogous to the way `get_joystick_name` works in this PR)

NewCreature:
Yeah, I had forgotten about that original PR and that a modified version of it was still hanging out here. That was 7 years ago. The project I was needing to fix the code for got put on hold. Now I'm trying to add joystick support to one of my current projects and thought I would just fix `al_get_joystick_name()` real quick since I noticed it wasn't implemented.

Joystick code in general is a mess since all the manufactures can't seem to follow the HID spec correctly. I didn't realize how bad it was since I wasn't trying to do autodetection back then. I fired up `ex_joystick_events` on the Steam Deck to see what it would report. Sure enough it was giving me the same messed up configuration I was getting under MacOS.

Apparently, the standard way to handle joysticks is to just assume people are using XBox 360 or compatible controllers and handling the axis/button configuration based on that assumption.

Should I go back to 0828211c7d1b1e4647efe228c5b6e08e49f16bac to see if I can make that work? I still have the device that wasn't working properly back then and have added quite a few others that I can test with.

SiegeLord:
Thanks.

> Apparently, the standard way to handle joysticks is to just assume people are using XBox 360 or compatible controllers and handling the axis/button configuration based on that assumption.

> Should I go back to https://github.com/liballeg/allegro5/commit/0828211c7d1b1e4647efe228c5b6e08e49f16bac to see if I can make that work? I still have the device that wasn't working properly back then and have added quite a few others that I can test with.

I'd love to have a system that detects XBox-like controllers and works across OSs. As it is now, I don't think this happens, e.g. my Logitech F310 splits its left thumbstick into two sticks.

NewCreature:
Unfortunately, the only way to get something that resembles detection is to keep a database of joysticks and have a remapping system.

I ended up implementing a remapping system into my own framework that sits on top of Allegro. The one thing that Allegro lacks is a way to get the joystick GUID/UUID. If it had that, I would be able to support more joysticks with my remapping system. It looks like some work on this is already being done:  #1426.

Currently, I'm just assuming controllers are XBox and allowing the user to remap if it is incorrect.

MarkOates:
Thanks for the merge @SiegeLord.

@NewCreature What types of name values can we expect from the `al_get_joystick_name()` here?

> Currently, I'm just assuming controllers are XBox and allowing the user to remap if it is incorrect.

Thanks for this assumption, it happens to work for me.

Also, thank you for mentioning https://github.com/liballeg/allegro5/pull/1426, if you wish to work on it and move it forward that would be great 🙏. I'll help if I have bandwidth.

NewCreature:
`al_get_joystick_name()` currently gets the name as reported by the HID device itself. This is what I consider the expected behavior. For instance, my 8BitDo controller correctly reports itself as an 8BitDo controller even when set to XInput mode (at least under MacOS). The controls themselves are mapped like an XBox controller would be.

As for #1426, I would be happy to help with it if needed. What work is left to be done there?

SiegeLord:
> Unfortunately, the only way to get something that resembles detection is to keep a database of joysticks and have a remapping system.

That's not what I meant. What I mean is, I am certain that XBox-like controllers all look the same to each OS, and Allegro should provide the same mapping across OS's. This is not the case right now.

PR #1451, opened 2023-06-28
Dat tool broken on Windows, can no longer create dat files. Temporary…

hanpab:
… file creation fails. tmpnam returns an absolute path on modern MSVC compilers, which no longer results in a valid file path when concatenated to the temporary directory path. Exclusively use Windows API (GetTempPath, GetTempFileName) to create the temporary file.


SiegeLord:
Seems reasonable. Thanks!

PR #1452, opened 2023-07-01
[OSX] Do not wait for main thread when appending to native text log

connorjclark:
I tracked down a huge FPS drop in my application to this line. The application was doing tons of logging to a native text log, and each one (on OSX) waits for the main thread to ack the message. The FPS difference was something like 1000 -> 30.

I don't see a reason to wait for this message to be ack'd, and things still seem to function just the same.

I did not confirm if other platforms have the same problem.

### DIFF
--- a/addons/native_dialog/osx_dialog.m
+++ b/addons/native_dialog/osx_dialog.m
@@ -352,7 +352,7 @@ void _al_append_native_text_log(ALLEGRO_NATIVE_DIALOG *textlog)
         NSString *text = [NSString stringWithUTF8String: al_cstr(textlog->tl_pending_text)];
         [view performSelectorOnMainThread:@selector(appendText:)
                                withObject:text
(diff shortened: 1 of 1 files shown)


pedro-w:
If it works, it works I suppose, but what is the lifetime of the autoreleased string `text`? Could it be released before the `appendText:` gets run, or does `performSelectorOnMainThread` retain a reference to the object passed in (I can't remember)
Otherwise this is a great performance increase, well spotted!

pedro-w:
It's weird the docs don't mention it - even that the answer to that StackOverflow thread you cited just asserts that it does without any evidence. I'm sure it's right though because it would be awkward to use `performSelector...` correctly if it did not.

I wonder if there are any other cases where we're waiting unnecessarily in the allegro code?

SiegeLord:
Let's try it without additional retains. Thanks!

PR #1454, opened 2023-07-10
add DND support to X11

allefant:
I needed drag&drop support for my application so added it. Since I only use X11 it is only implemented for X11, but figured I might as well create a PR in case someone else wants it. The API consists only of a new ALLEGRO_DRAG_AND_DROP display flag which, if set for a display will cause that display to issue ALLEGRO_EVENT_DROP events when files are dragged over it from another application.


SiegeLord:
Nice. Merged via https://github.com/liballeg/allegro5/commit/6cf1825922e13ce83b67d90c9a050a3ad8aa26c7.

PR #1455, opened 2023-07-17
add `[compatability] osx_tell_dock_outside_bundle`

connorjclark:
This old behavior is undesirable for purely-console applications, as it makes OSX flicker the screen / current app loses focus.

Default is to keep doing this, but now there is a new option to suppress it.

_____

Discussed on Discord: https://discord.com/channels/993415281244393504/1013128461319159932/1130312994564358244

```
connorclark — Today at 6:40 PM
I have a console application that uses allegro 5. On init, for OSX we get a flash of a window gaining focus. This comes from the init sequence calling osx_tell_dock, signaling the process as a graphical app:  TransformProcessType(&psn, kProcessTransformToForegroundApplication).

Is there anything part of the allegro 5 api that can be used to signal that we are not a graphical app, such that this al_init call could know not to do this? 
Perhaps we could move this to the create_display code for OSX.
SiegeLord — Today at 6:59 PM
That seems like a good idea
If that's impossible for some reason, we could add a config option to disable this behavior
Well, probably need a config option anyway... so perhaps best to just do that to start
connorclark — Today at 7:07 PM
I just tried moving it there, and it seems to work well.

This config would be on the flags given to initialize?
SiegeLord — Today at 7:08 PM
No, allegro5.cfg
connorclark — Today at 7:08 PM
To be clear, you're wanting to preserve the (osx-only) behavior that a console application is viewable in the dock?
SiegeLord — Today at 7:08 PM
Yeah, I think so since that's not uncommon to do
Like console as opossed to in a bundle
Not console as in no windows
I'm... not sure what you'd want that, maybe that's a always a bug?
Or maybe that's useful sometimes?
```

### DIFF
--- a/src/macosx/osxgl.m
+++ b/src/macosx/osxgl.m
@@ -1301,6 +1301,8 @@ static ALLEGRO_DISPLAY* create_display_fs(int w, int h)
       osx_run_fullscreen_display(dpy);
    });
    [pool drain];
+
+   _al_osx_tell_dock();
    return &dpy->parent;
 }
 #if 0
@@ -1696,6 +1698,7 @@ static ALLEGRO_DISPLAY* create_display_win(int w, int h) {
 
    [pool drain];
 
+   _al_osx_tell_dock();
    return display;
 }
 
--- a/allegro5.cfg
+++ b/allegro5.cfg
@@ -238,3 +238,8 @@ max_page_size = 0
 # automatically, but may break old code that handled this eventuality.
 # Set this to false for such code.
 automatic_menu_display_resize = true
+
+# On OSX outside an app bundle, on system init allegro manually promotes the process
+# to a graphical application. This may be undesirable for console applications. Set
+# this to false to disable this behavior.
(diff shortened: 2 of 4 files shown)


src/macosx/osxgl.m
```diff
@@ -1090,6 +1090,26 @@ static CVReturn display_link_callback(CVDisplayLinkRef display_link,
    return kCVReturnSuccess;
 }
 
+/* osx_tell_dock:
+ *  Tell the dock about us; promote us from a console app to a graphical app
+ *  with dock icon and menu
+ */
+static void osx_tell_dock(void)
```
SiegeLord, line 1097: Would you mind changing the implementation of the system.m's copy of this function to what you have here, renaming it to `_al_osx_tell_dock` and then calling it from osxgl.m? Trying to avoid copy-pasting here.

The forward declaration for that would go here: `include/allegro5/platform/aintosx.h`

SiegeLord:
Thanks!

Issue #1300, opened 2021-12-25
OSX hat events are weird


SiegeLord:
The `hat_mapping` might be off by 1, and it's also very weird how it always sends two events per press.

ReiquelApplegate:
Can this issue be closed against https://github.com/liballeg/allegro5/pull/1457 now?

SiegeLord:
The "two events per press" is still active, AFAIK.

PR #1457, opened 2023-08-02
osx: fix off-by-one error for hat switches

connorjclark:
FYI, I only have a single device with a hat switch (on my mac, my X360 controller dpad is a hat switch).

I noticed that for cardinal-direction inputs, I was getting a non-zero value for _both_ axises. Diagonal inputs were off too. I realized that it was an off-by-one error in the hat switch mapping.

Here's a spec sheet for how [these values work](https://www.usb.org/sites/default/files/documents/hid1_11.pdf). An excerpt:

> Indicates whether the control has a state in which it
is not sending meaningful data. One possible use of
the null state is for controls that require the user to
physically interact with the control in order for it to
report useful data. For example, some joysticks
have a multidirectional switch (a hat switch).
When a hat switch is not being pressed it is in a
null state. When in a null state, the control will
report a value outside of the specified Logical
Minimum and Logical Maximum (the most
negative value, such as -128 for an 8-bit value). 

In addition to fixing the off-by-one issue, because of the above paragraph I changed how the null state is detected. Now it looks at the min/max range.

### DIFF
--- a/src/macosx/hidjoy.m
+++ b/src/macosx/hidjoy.m
@@ -494,12 +494,12 @@ static void value_callback(
    int int_value = IOHIDValueGetIntegerValue(value);
 
    if (joy->dpad == elem){
-      if (int_value > 0 && int_value <= MAX_HAT_DIRECTIONS) {
-         osx_joy_generate_axis_event(joy, joy->dpad_stick, joy->dpad_axis_vert,  (float)hat_mapping[int_value-1].axisV);
-         osx_joy_generate_axis_event(joy, joy->dpad_stick, joy->dpad_axis_horiz, (float)hat_mapping[int_value-1].axisH);
-      } else if (joy->min[joy->dpad_stick][1] > int_value || joy->max[joy->dpad_stick][1] < int_value) {
+      if (joy->min[joy->dpad_stick][1] > int_value || joy->max[joy->dpad_stick][1] < int_value) {
          osx_joy_generate_axis_event(joy, joy->dpad_stick, joy->dpad_axis_vert,  0);
          osx_joy_generate_axis_event(joy, joy->dpad_stick, joy->dpad_axis_horiz, 0);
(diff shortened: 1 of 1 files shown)


SiegeLord:
Nice find, this is probably what caused #1300.

SiegeLord:
Hmm... well this is awkward. I was testing this with my Logitech F310 and the current code actually works fine with it, and the changes here make it be off by 1. That's... puzzling.

SiegeLord:
Could you log what your `joy->min`/`joy->max` values are? They are 0 and 7 for me.

connorjclark:
I get 1 and 8. I think I see the issue, will update in a minute.

connorjclark:
How's the new patch work for you?

src/macosx/hidjoy.m
```diff
@@ -493,11 +492,19 @@ static void value_callback(
...
 
    if (joy->dpad == elem){
-      if (int_value >= 0 && int_value < MAX_HAT_DIRECTIONS) {
-         osx_joy_generate_axis_event(joy, joy->dpad_stick, joy->dpad_axis_vert,  (float)hat_mapping[int_value].axisV);
-         osx_joy_generate_axis_event(joy, joy->dpad_stick, joy->dpad_axis_horiz, (float)hat_mapping[int_value].axisH);
+      if (int_value >= min && int_value <= max) {
+         int index = int_value - min;
+         if (index < MAX_HAT_DIRECTIONS) {
+            osx_joy_generate_axis_event(joy, joy->dpad_stick, joy->dpad_axis_vert,  (float)hat_mapping[index].axisV);
+            osx_joy_generate_axis_event(joy, joy->dpad_stick, joy->dpad_axis_horiz, (float)hat_mapping[index].axisH);
```
connorjclark, line 503: if the device has just 4 distinct states instead of 8, prob can support by just multiplying `index` by 2. But I don't have such a device so I can't test that theory.

SiegeLord:
Great, this works. Thanks!

Issue #1462, opened 2023-08-25
[osx] Unexpected behavior for right stick of modern Xbox controller


connorjclark:
[Xbox Wireless Controller](https://www.amazon.com/gp/product/B01GW3H3U8/ref=ppx_yo_dt_b_search_asin_title?ie=UTF8&psc=1)

This controller has three sticks: 0 and 1 are the left and right sticks, and 2 is the DPAD (hat switch).

According to the allegro joystick interface on OSX, these switches have these number of axis:

Stick 0 has 3 axis (kHIDUsage_GD_X, kHIDUsage_GD_Y, kHIDUsage_GD_Y)
Stick 1 has 1 axis (kHIDUsage_GD_Rz)
Stick 2 has 2 axis (the hat switch)

I noticed that when processing stick 1 in `static void add_elements`, two HID usage interfaces are ignored with decimal values 196 and 197. These values are not defined in the [HID header](https://opensource.apple.com/source/IOHIDFamily/IOHIDFamily-34.1.1/IOHIDSystem/IOKit/hidsystem/IOHIDUsageTables.h.auto.html).

Listing the odd things:

- Why does Stick 0 have 3 axis instead of 2? I know it's a totally different stack, but https://hardwaretester.com/gamepad only lists two axis for both Stick 0 and 1.
- Why don't Sticks 0 and 1, which are identical physical controls, have an identical number of axis?
- What are usages 196 and 197? Did this Xbox controller make up their own usage ids?
- When moving the right stick (Stick 1) to the right, I get a mix of events for Stick 1, and for Stick 0 Axis 2, which is super broken. When moving that right stick Up I don't get any events from Stick 0.

connorjclark:
I found a similar issue on Windows for DirectInput `wjoydxnu.cpp` w/ the Stadia controller. Same exact stick layout, same exact number of axis as the example described above.

![image](https://github.com/liballeg/allegro5/assets/4071474/c26169f9-20ed-43d3-a7d7-a1465abb1bc5)

![image](https://github.com/liballeg/allegro5/assets/4071474/50a03220-95e9-4b1f-9779-d133136ad09e)

PR #1463, opened 2023-08-26
Better support for secondary stick in DirectInput driver

connorjclark:
As referenced here: https://github.com/liballeg/allegro5/issues/1462#issuecomment-1694014874

This patch fixes detection of the secondary thumbstick. I tested on two devices: a PS4 Dualshock, and a Stadia controller.

The code assumed that Rx, Ry, Rz will always be what maps to the secondary stick. However, that apparently was too hard for controller manufacturers to do. For example, my above mentioned controllers use Z and Rz for the second stick. Short of including a database of controller mappings, this patch introduces a heuristic to guess how to map to a secondary stick.

The idea for the solution came from https://www.gamedev.net/forums/topic/613913-directinput-identifying-second-thumbstick/

### DIFF
--- a/src/win/wjoydxnu.cpp
+++ b/src/win/wjoydxnu.cpp
@@ -79,6 +79,7 @@
 #include <mmsystem.h>
 #include <process.h>
 #include <dinput.h>
+#include <utility>
 
 /* We need XInput detection if we actually compile the XInput driver in. 
 */ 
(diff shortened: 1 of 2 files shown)


src/win/wjoydxnu.cpp
```diff
@@ -550,6 +557,38 @@ static BOOL CALLBACK object_enum_callback(LPCDIDEVICEOBJECTINSTANCE lpddoi, LPVO
...
+   // The first two axis (not X or Y) are stored and used as the secondary stick.
+   // The order of the enumeration is assumed to give us the X followed by the Y axis. However,
+   // an override is provided in the case that the second axis seen here is for Z, in which case that
+   // will be used as the X axis. This is to fit behaviors seen on specific controllers:
+   // - PS4 DualShock
+   // - Stadia
+   // ...it may be that using `Z` really is common for some input devices.
+   //
+   // Solution came from https://www.gamedev.net/forums/topic/613913-directinput-identifying-second-thumbstick/
+   if (j)
```
SiegeLord, line 569: Now that we have a backwards compatibility mechanism, we can merge a version of this. Could you add check like this https://github.com/liballeg/allegro5/blob/8575e373ab79c7bc25f8425990fc6780e75e69c8/src/macosx/hidjoy.m#L192C4-L192C76 here, so this behavior can be configured?

Also, could `j` be called something more informative, like... `offset_override` maybe?

PR #1464, opened 2023-08-26
Add name lookup for xinput controllers for limited set of devices

connorjclark:
This patch optionally makes use of an undocumented xinput 1.4 API to get the vendor and product ids, and then uses a small lookup function of some popular controllers to get the actual name.

The vid/pid numbers are USB identifiers. This [site](https://devicehunt.com/view/type/usb/vendor/045E/device/02FF) might be useful to find additional devices. I got lucky and found a list of devices somewhere on GitHub.


SiegeLord:
Looks good, thanks. It detects my 8bitdo Pro 2 Wired as "XBox Elite", which kind of underscores the limitations of this kind of API, but I guess it's a slight improvement either way!

connorjclark:
Yeah one of my 8bitdo reads as Pro Controller :p

PR #1467, opened 2023-09-11
Rework how rate-limiting of resize events works on Windows.

### LINKED ISSUE #1381: Child window with ALLEGRO_MENU causes crash upon open/close display.
Using the ex_menu.c example in a program.
The program will crash, if you open/close the child window multiple times.

This is a problem, since I'm developing a game where you can open editor windows for e.g. maps, items, etc.

Crash upon close is in 'allegro5/platform/aintwthr.h' line 59
```
AL_INLINE(void, _al_mutex_lock, (struct _AL_MUTEX *m),
{
   if (m->cs)
      EnterCriticalSection(m->cs);
})
```

Crash upon open is in 'allegro/src/win/wwindow.c' line 196
`GetWindowInfo(my_window, &wi);`
and sometimes other places, like allegro/src/memory.c line 44
```
/* Function: al_malloc_with_context
 */
void *al_malloc_with_context(size_t n,
   int line, const char *file, const char *func)
{
   if (mem)
      return mem->mi_malloc(n, line, file, func);
   else
      return malloc(n); //This right here line 44
}
```

SiegeLord:
Before, we'd spawn bespoke threads every 50ms. One issue with this design was that destroying a display before that 50ms elapsed caused crashes, which happened in the menu code.

The new code has a persistent thread that periodically sends those resize events. More importantly, we make sure that thread is aware of displays being destroyed.

Fixes #1381

### DIFF
--- a/src/win/wwindow.c
+++ b/src/win/wwindow.c
@@ -311,7 +311,7 @@ static void win_generate_resize_event(ALLEGRO_DISPLAY_WIN *win_display)
    WINDOWINFO wi;
    int x, y, w, h;
 
-   if (win_display->ignore_resize) {
+   if (win_display->ignore_resize || win_display->d3d_ignore_resize) {
       return;
    }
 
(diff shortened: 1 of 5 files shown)


src/win/wwindow.c
```diff
@@ -959,31 +974,43 @@ static LRESULT CALLBACK window_callback(HWND hWnd, UINT message,
...
           * ends up blocking the loop in d3d_display_thread_proc (which is
           * where we are called from, if using D3D).  Rather than batching up
           * intermediate resize events which the user cannot acknowledge in the
           * meantime anyway, make it so only a single resize event is generated
           * at WM_EXITSIZEMOVE.
           */
+         al_lock_mutex(resize_event_thread_mutex);
          if (d->flags & ALLEGRO_DIRECT3D_INTERNAL) {
-            resize_postponed = true;
+            win_display->ignore_resize = true;
```
SiegeLord, line 999: Use a different variable, so that al_resize_display can be ostensibly called while live resize is happening and we don't get out of sync.

PR #1472, opened 2023-09-26
Fix cflags include path in allegro.pc

kkaempf:
Include files are installed into /usr/include/allegro5 making a `#include <allegro.h>` fail.

Therefore `pkgconfig --cflags allegro-5` must report `/usr/include/allegro5` instead of `/usr/include`

### DIFF
--- a/misc/allegro.pc.in
+++ b/misc/allegro.pc.in
@@ -11,4 +11,4 @@ Description: Allegro game programming library
 Version: ${version}
 Libs: -L${libdir} -lallegro${suffix}
 Libs.private: @link_with@
-Cflags: -I${includedir}
+Cflags: -I${includedir}/allegro${versuffix}


SiegeLord:
I think it's expected to include Allegro headers as: `#include <allegro5/allegro.h>`. Where did you see a plain `#include <allegro.h>`?

kkaempf:
@SiegeLord thanks for taking a look !


> I think it's expected to include Allegro headers as: `#include <allegro5/allegro.h>`.
😮 
So that's a change coming with version 5 of allegro ?!

> Where did you see a plain `#include <allegro.h>`?
I saw it in https://github.com/alvaroalea/aspectrum - but it doesn't seem to support allegro-5 anyways.

Feel free to close this PR.

SiegeLord:
>  So that's a change coming with version 5 of allegro ?!

Yeah, I think that happened pretty early on, don't know the reason for it though.

PR #1478, opened 2023-10-03
Android: add the possibility to request a specific OpenGL ES version

### LINKED ISSUE #1476: Android: unable to require an OpenGL ES 3.x context
I would like to use GLSL ES 3 shaders on Android. However, I cannot do that reliably because I am unable to require an OpenGL ES 3.x context.

As can be seen below, Allegro will request at most an OpenGL ES 2.0 context. Requiring a specific version with [al_set_new_display_option](https://liballeg.org/a5docs/trunk/display.html#al_set_new_display_option) has no effect whatsoever.

https://github.com/liballeg/allegro5/blob/0be7a0286655893d4fac691243f3062f2107a2f4/android/gradle_project/allegro/src/main/java/org/liballeg/android/AllegroEGL.java#L201-L215

According to the [EGL spec](https://registry.khronos.org/EGL/specs/eglspec.1.5.pdf) (sections 3.7.1 and 3.7.1.1), the created context will be of the requested version or of any later version which is backwards compatible with that version. When Allegro requests an OpenGL ES 2.0 context, an implementation **may** return an OpenGL ES 3.x context - if it's supported. Typically it will. Maybe it won't. In any case, I am unable to require an OpenGL ES 3.x context at this time.

In addition: according to the [Android documentation](https://developer.android.com/develop/ui/views/graphics/opengl/about-opengl), when creating a context whose version that is higher than the minimum required by the app, one may first "create a minimum-supported OpenGL ES context and check the version value. [...] With this approach, if you discover that the device supports a higher-level API version, *you must destroy the minimum OpenGL ES context and create a new context with the higher available API version.*"

alemart:
On Android, Allegro requested at most an OpenGL ES 2.0 context. This patch adds the possibility to request a specific OpenGL ES version - including 3.x - with `al_set_new_display_option`, [as stated in the documentation](https://liballeg.org/a5docs/trunk/display.html#al_set_new_display_option). Hints `ALLEGRO_REQUIRE` and `ALLEGRO_SUGGEST` are taken into account. This fixes #1476.

### DIFF
--- a/android/gradle_project/allegro/src/main/java/org/liballeg/android/AllegroEGL.java
+++ b/android/gradle_project/allegro/src/main/java/org/liballeg/android/AllegroEGL.java
@@ -10,7 +10,8 @@ class AllegroEGL
 {
    private static final String TAG = "AllegroEGL";
 
-   private static final int EGL_CONTEXT_CLIENT_VERSION = 0x3098;
+   private static final int EGL_CONTEXT_MAJOR_VERSION = 0x3098;
+   private static final int EGL_CONTEXT_MINOR_VERSION = 0x30fb;
    private static final int EGL_OPENGL_ES_BIT = 1;
    private static final int EGL_OPENGL_ES2_BIT = 4;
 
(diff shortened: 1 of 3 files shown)


SiegeLord:
Thanks!

PR #1480, opened 2023-10-03
Android: trigger keyboard events of Android game keys with or without an active joystick

### LINKED ISSUE #1479: Android-specific keys never trigger keyboard events
Allegro defines a set of "Android game keys" in keycodes.h. Of the 18 defined keys, 10 never trigger keyboard events on Android after a successful call to `al_install_joystick()`:

* The following 10 never trigger a keyboard event: `ALLEGRO_KEY_BUTTON_A`, `ALLEGRO_KEY_BUTTON_B`, `ALLEGRO_KEY_BUTTON_X`, `ALLEGRO_KEY_BUTTON_Y`, `ALLEGRO_KEY_BUTTON_L1`, `ALLEGRO_KEY_BUTTON_R1`, `ALLEGRO_KEY_DPAD_UP`, `ALLEGRO_KEY_DPAD_DOWN`, `ALLEGRO_KEY_DPAD_LEFT`, `ALLEGRO_KEY_DPAD_RIGHT`.
* The following 8 do trigger keyboard events: `ALLEGRO_KEY_START`, `ALLEGRO_KEY_SELECT`, `ALLEGRO_KEY_BUTTON_L2`, `ALLEGRO_KEY_BUTTON_R2`, `ALLEGRO_KEY_THUMBL`, `ALLEGRO_KEY_THUMBR`, `ALLEGRO_KEY_DPAD_CENTER`, `ALLEGRO_KEY_SEARCH`.
* `ALLEGRO_KEY_MENU`, although not defined side-by-side with the Android game keys, also never triggers a keyboard event.

The keys that never trigger keyboard events are instead mapped to joystick input. Conversely, the keys that are not mapped to joystick input trigger keyboard events.

The present state of things leads to inconsistent behavior: in order to detect if the START button has been pressed, I have to listen to keyboard input. On the other hand, if I want to detect A, B, X, Y, I have to listen to joystick input. Other buttons behave in the same way: L1/R1 are detected as joystick input only, whereas L2/R2 are detected as keyboard input only.

Since Allegro defines 18 "Android game keys", could we have them all trigger keyboard events?
* Keycodes have publicly defined names, whereas joystick button numbers do not. In my code, I remap the joystick button numbers defined in `AllegroActivity` to the XInput button numbers defined in src/win/wjoyxi.c for consistency.
* If multiple gamepads are connected to a device, they all trigger keyboard events. This is not an issue, as Android game keys are already defined and gamepad differentiation can be done using the joystick API.
* All 18 game keys are correctly recognized as long as you do not call `al_install_joystick()`. Once you do and it succeeds, Allegro blocks 10 keys (in addition to the menu key), leading to inconsistent behavior.

alemart:
Allegro defines 18 "Android game keys" in keycodes.h. Keyboard events of 10 of them will not be triggered after a successful call to `al_install_joystick()`, leading to inconsistent behavior. This patch makes events get triggered for these keys as well, regardless if a joystick has been installed or not. It makes behavior consistent and fixes #1479.

### DIFF
--- a/android/gradle_project/allegro/src/main/java/org/liballeg/android/KeyListener.java
+++ b/android/gradle_project/allegro/src/main/java/org/liballeg/android/KeyListener.java
@@ -139,21 +139,17 @@ class KeyListener implements View.OnKeyListener
       if (index >= 0) {
          code = getCode(keyCode, event, index);
       }
-      if (code == -1) {
-         return onKeyboardKey(v, keyCode, event);
-      }
-      else if (code == -2) {
-         return true;
-      }
-      if (event.getAction() == KeyEvent.ACTION_DOWN) {
-         if (event.getRepeatCount() == 0) {
-            nativeOnJoystickButton(index, code, true);
+      if (code >= 0) {
+         if (event.getAction() == KeyEvent.ACTION_DOWN) {
+            if (event.getRepeatCount() == 0) {
+               nativeOnJoystickButton(index, code, true);
+            }
+         }
+         else {
+            nativeOnJoystickButton(index, code, false);
          }
       }
-      else {
-         nativeOnJoystickButton(index, code, false);
-      }
-      return true;
+      return onKeyboardKey(v, keyCode, event);
    }
 
    private void volumeChange(int inc)


SiegeLord:
This isn't needed anymore now that  #1483 is merged, right?

alemart:
> This isn't needed anymore now that #1483 is merged, right?

I think that there is still a design issue that hasn't been addressed. If you agree, I would like to make a change.

Even though the joystick API now supports all buttons, game controllers are still mapped to the keyboard API, and only partially so. Such mapping has no use since #1483 and has an incomplete button set. This behavior isn't documented anywhere and is inconsistent with other platforms. Why keep it? Gamepads are not keyboards.

We could now have users rely on the joystick API to read input from game controllers, as they do in other platforms. We could use the keyboard API to catch input of [remote controllers of Smart TVs](https://developer.android.com/training/tv/start/controllers) instead.

In this proposal, devices and APIs will no longer overlap. This is a more useful design in my view. May I proceed with this idea?

SiegeLord:
Yeah, let's do that. Thanks for thinking through this!

PR #1481, opened 2023-10-03
Android: add boundary checking to Key.alKey()

alemart:
This patch adds boundary checking to `Key.alKey()`. The input `keyCode` [is not guaranteed to be in the [0,255] range](https://developer.android.com/reference/android/view/KeyEvent) (e.g., Android TV).


SiegeLord:
Thanks!

PR #1483, opened 2023-10-04
Android: add new joystick buttons

### LINKED ISSUE #1482: Android: missing joystick buttons
Currently, the Android version of Allegro detects gamepads as having 11 buttons: A, B, X, Y, L1, R1, DPAD_LEFT, DPAD_RIGHT, DPAD_UP, DPAD_DOWN and MENU. It would be nice if we could use the joystick API to read the state of other buttons such as: SELECT, START, THUMBL, THUMBR, L2, R2, DPAD_CENTER...

https://developer.android.com/develop/ui/views/touch-and-input/game-controllers/controller-input

Related to #1479

alemart:
This patch adds the following joystick buttons: DPAD_CENTER, L2, R2, THUMBL, THUMBR, START, SELECT, MODE.

Closes #1482.


alemart:
Commit c3a855995a4da1f5834759dc92f66ec660a0430e adds all gamepad buttons, as requested in #1479, and improves the design of the Android joystick as follows:   
- Adds BUTTON_Z and BUTTON_C
- Adds KEYCODE_BACK as an alias to KEYCODE_BUTTON_SELECT, as in SDL
- Changes KEYCODE_MENU and makes it an alias to KEYCODE_BUTTON_START, as in SDL
- Adds generic gamepad buttons 1 to 16 and maps them to button numbers 20 .. 35, as in SDL

References:
- https://github.com/libsdl-org/SDL/blob/5d95cbde37c749183ae5de97008bee99cb7f40a5/src/joystick/android/SDL_sysjoystick.c#L69-L172
- https://developer.android.com/reference/android/view/KeyEvent#KEYCODE_BUTTON_1
- https://wiki.archlinux.org/title/Gamepad
- https://ftctechnh.github.io/ftc_app/doc/javadoc/com/qualcomm/robotcore/hardware/Gamepad.html

alemart:
I have also added the button names.

alemart:
@SiegeLord Were you able to test the joystick? I find that it works with my Bluetooth controller and also with a USB controller I purchased recently. All buttons and [axes](https://github.com/liballeg/allegro5/pull/1507) of the standard gamepad on Android are mapped, as requested. See pic:

Screenshot_20231127-151147

SiegeLord:
Okay, I managed to test it. It works well, with one exception. The back key (select), does not get triggered, because it gets sent to the onKey as KEYCODE_BACK rather than a joystick button... we can fix this later.

SiegeLord:
Thanks a lot!

alemart:
> Okay, I managed to test it. It works well, with one exception. The back key (select), does not get triggered, because it gets sent to the onKey as KEYCODE_BACK rather than a joystick button... we can fix this later.

@SiegeLord My controller emits `KEYCODE_BUTTON_SELECT`. It works properly in this case. [Some controllers emit `KEYCODE_BACK` in its place](https://ftctechnh.github.io/ftc_app/doc/javadoc/com/qualcomm/robotcore/hardware/Gamepad.html). I don't own one of these in order to do the testing, but the code is expected to handle them as well. The key is tied to a device ID. Maybe there is a missing detail?

SiegeLord:
Yeah, it's the Logitech F310. The issue is that device id is -1 for that event, so there's probably nothing we can do.

PR #1484, opened 2023-10-05
Android: make detection of game controllers more robust

alemart:
Currently, Allegro determines that an input device should be recognized as a joystick if its source type matches `SOURCE_JOYSTICK`. The Android documentation states that this source type indicates that the input device has analog control sticks. There is another value, `SOURCE_GAMEPAD`, that indicates that the input device has gamepad buttons. Both should be taken into account when testing if an input device is a game controller. This patch makes the adjustment.

For more information and an example, refer to:

https://developer.android.com/develop/ui/views/touch-and-input/game-controllers/controller-input?hl=en#input


SiegeLord:
Thanks!

PR #1485, opened 2023-10-06
Android: implement `al_get_joystick_name()`

alemart:
This patch implements `al_get_joystick_name()` on Android.

If the [product ID / vendor ID](https://github.com/liballeg/allegro5/pull/1426) API pans out, the Android implementation can be done in a similar way.


SiegeLord:
Thanks!

PR #1493, opened 2023-10-19
Android: fix corruption of bitmaps created with the `ALLEGRO_NO_PRESERVE_TEXTURE` flag

### LINKED ISSUE #1492: Android: bitmaps created with the `ALLEGRO_NO_PRESERVE_TEXTURE` flag get corrupted
On Android, bitmaps created with the `ALLEGRO_NO_PRESERVE_TEXTURE` flag get corrupted and stay unusable when the application loses focus. Video bitmaps created without the flag are recovered by calling `al_acknowledge_drawing_resume()`. Bitmaps created with the flag could be "recovered" similarly (i.e., made usable again, even though they lose their pixel data).

---

Here is an example of bitmap corruption. I render the scene to bitmaps created with the `ALLEGRO_NO_PRESERVE_TEXTURE` flag. I can walk around the issue by destroying and recreating such bitmaps. However, it would be better if Allegro handled the issue transparently, as it already does for other video bitmaps.

https://github.com/liballeg/allegro5/assets/221477/071312d9-1856-4e3e-bd06-cc04a51329d5

The controls are rendered directly to the screen, so they aren't affected by the glitch.

alemart:
This patch fixes #1492: corruption of bitmaps created with the `ALLEGRO_NO_PRESERVE_TEXTURE` flag when the application loses focus on Android.

Video bitmaps created without the `ALLEGRO_NO_PRESERVE_TEXTURE` flag are already recovered by Allegro. I take a similar approach:

- I recreate the internal textures of the bitmaps - even though they lose their pixel data - when `al_acknowledge_drawing_resume()` is called.

- I change `_al_ogl_upload_bitmap_memory()` so that it accepts `ptr` == NULL. This function was already documented as accepting NULL, but its code demanded otherwise.

### DIFF
--- a/src/android/android_display.c
+++ b/src/android/android_display.c
@@ -841,8 +841,7 @@ static void android_acknowledge_drawing_halt(ALLEGRO_DISPLAY *dpy)
       int bitmap_flags = al_get_bitmap_flags(bmp);
 
       if (!bmp->parent &&
-         !(bitmap_flags & ALLEGRO_MEMORY_BITMAP) &&
-         !(bitmap_flags & ALLEGRO_NO_PRESERVE_TEXTURE))
+         !(bitmap_flags & ALLEGRO_MEMORY_BITMAP))
       {
          ALLEGRO_BITMAP_EXTRA_OPENGL *extra = bmp->extra;
          al_remove_opengl_fbo(bmp);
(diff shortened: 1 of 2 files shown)


SiegeLord:
Looks good, thanks!

PR #1494, opened 2023-10-20
Android: delete local JNI string references and use the jstring type

alemart:
This patch calls `DeleteLocalRef` to delete local JNI string references and uses the `jstring` type in `_jni_getString()`.


SiegeLord:
Thanks!

Issue #1496, opened 2023-10-22
X11 al_set_window_position unexpected behavior


elias-pschernig:
In X11 al_get_window_position returns the client area position (where the 0/0 the Allegro window is on the screen, not where the outer window border is), but in al_set_window_position we set a different position, so al_get_window_position(&x, &y) followed by al_set_window_position(x, y) will move the window around. I tried several window managers from non-compositing X11 to compositing to Wayland and the behavior at least seems consistently wrong.

I assume the expected behavior would be that passing al_get_window_position coordinates to al_set_window_position would not move the window?

SiegeLord:
Yeah, I believe that's expected. There was a similar issue on Windows that was fixed (had to do with initial positions: https://github.com/liballeg/allegro5/commit/fead757535d9d416cd956880c5574ad215582d0a). In addition to that lack of movement, ideally (0, 0) (or whatever al_get_monitor_info returns for the top-left corner) should correspond to that top left corner. This isn't the case on Windows, for example, and its hard to fix: https://github.com/liballeg/allegro5/issues/1459

PR #1497, opened 2023-10-24
add al_get_window_borders() and implement for X11

elias-pschernig:
Fixes https://github.com/liballeg/allegro5/issues/1496

I added a function to query window borders (like SDL_GetWindowBordersSize) and made al_set_window_position and al_get_window_position both use *outer* window coordinates for X11.

Also added a swap button and move-to-same-place key to ex_windows to test the changes.

I initially thought just using outer coordinates is fine as https://liballeg.org/a5docs/trunk/display.html#al_set_window_position and https://liballeg.org/a5docs/trunk/display.html#al_get_window_position do not specify if they mean inner or outer coordinates. However https://liballeg.org/a5docs/trunk/display.html#al_set_new_window_position specifically mentions the "top left pixel of the client area". So considering that, I'd be fine with changing the PR to use inner coordinates everywhere, should be easy to change now. Also for what it's worth, SDL uses inner coordinates for everything (but provides SDL_GetWindowBordersSize to easily convert between inner and outer).

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -405,10 +405,10 @@ void al_get_window_position(ALLEGRO_DISPLAY *display, int *x, int *y)
 
 /* Function: al_get_window_borders
  */
-bool al_get_window_borders(ALLEGRO_DISPLAY *display, int *left, int *right, int *top, int *bottom)
+bool al_get_window_borders(ALLEGRO_DISPLAY *display, int *left, int *top, int *right, int *bottom)
 {
    if (display && display->vt && display->vt->get_window_borders) {
-      return display->vt->get_window_borders(display, left, right, top, bottom);
+      return display->vt->get_window_borders(display, left, top, right, bottom);
    }
    else {
       return false;
--- a/src/display.c
+++ b/src/display.c
@@ -406,7 +406,7 @@ void al_get_window_position(ALLEGRO_DISPLAY *display, int *x, int *y)
  */
 bool al_get_window_borders(ALLEGRO_DISPLAY *display, int *left, int *right, int *top, int *bottom)
 {
-   if (display && display->vt && display->vt->get_window_position) {
+   if (display && display->vt && display->vt->get_window_borders) {
       return display->vt->get_window_borders(display, left, right, top, bottom);
    }
    else {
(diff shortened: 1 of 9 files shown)


SiegeLord:
Ah drat, that's a great find about the `al_set_new_window_position` documentation. Let me check what Windows is doing and see what's feasible there, but probably we'll need to change it to inner coordinates.

src/display.c
```diff
@@ -402,6 +402,18 @@ void al_get_window_position(ALLEGRO_DISPLAY *display, int *x, int *y)
    }
 }
 
+/* Function: al_get_window_position
+ */
+bool al_get_window_borders(ALLEGRO_DISPLAY *display, int *left, int *right, int *top, int *bottom)
+{
+   if (display && display->vt && display->vt->get_window_position) {
```
SiegeLord, line 409: This should be `get_window_borders`

SiegeLord:
Ok, for windows I'll have to revert https://github.com/liballeg/allegro5/commit/fead757535d9d416cd956880c5574ad215582d0a and then add that logic to al_get/set functions.

So yeah, let's do inner coordinates for this PR as well.

SiegeLord:
Hmm... borders aren't being got under my KDE. Investigating.

elias-pschernig:
> Hmm... borders aren't being got under my KDE. Investigating.

Let me install KDE here as well. Maybe that's what the old XTranslateCoordinates was for and we need both depending on WM *sigh* I'll also try the SDL build and if that works can at least just copy more code :)

elias-pschernig:
I got it to work on KDE. For some reason the call to _al_xwin_get_borders from the window creation function doesn't work. I simply moved it into the XConfigure event handler and there it works. I also re-query the borders on every such event now, not sure that is necessary. Else we could do it only if the initial placement flag is still set. Or we could make the window creation wait for the first XConfigure event.

Also was worried about calling XMoveWindow from within the event handler but it seems to work.

elias-pschernig:
Hm, also, initially I had made the new al_get_window_borders function use the same coordinate order as X11, but for Allegro left-top-right-bottom is the usual order so changed to that.

SiegeLord:
Great, that seems to work, thanks! I'll see if I can implement it for other platforms (Linux also has the GTK_TOPLEVEL vtable that needs this).

PR #1498, opened 2023-10-26
Use 16-bit operand for fnstsw instruction

greg-kennedy:
EDIT: Whoops, this was a mistake, closing it in error :)

### DIFF
--- a/src/i386/icpus.s
+++ b/src/i386/icpus.s
@@ -67,7 +67,7 @@ FUNC(_i_is_fpu)
 
    fninit
    movl $0x5A5A, %eax
-   fnstsw %ax
+   fnstsw %eax
    cmpl $0, %eax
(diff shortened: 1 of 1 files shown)


greg-kennedy:
Hmm, actually, maybe don't merge this one without a bit of testing.  Playing around with this on godbolt.org I can get errors with `%eax` but with `%ax` it works fine.  Strange...

pedro-w:
*Is* this a mistake? - seems you're correct about AX vs EAX
https://godbolt.org/z/7x8WxdEjP

greg-kennedy:
%ax (what you have now) is correct, %eax is wrong.

This was fixed back in 2009 in commit ed688d82020ee5c8c31dd8ac921c5be1a38a427c and I mistakenly sent a PR to un-do that, sorry :)

PR #1501, opened 2023-10-30
[osx] Fix infinite loop in set_window_position

connorjclark:
Randomly stumbled upon this.


pedro-w:
Well spotted! While we're on that function, in this line:
https://github.com/liballeg/allegro5/blob/9bdf4face2e2085547f5af766f8001ffec11dcae/src/macosx/osxgl.m#L2284
Should some of those `>` or `<` be `>=` or `<=`?

SiegeLord:
Thanks! Just C things...

PR #1507, opened 2023-11-14
Android joystick: add trigger buttons

alemart:
This adds the trigger buttons to the Android joystick. They are mapped as two analog sticks with a single axis each, just like in the Windows XInput driver.

This complements #1483.


SiegeLord:
This is great, thanks! Merged as https://github.com/liballeg/allegro5/commit/22cd69ad897dcdd6d9224e9d4a2e58ce6a0f2016.

Issue #1510, opened 2023-11-17
Timer thread hangs when system clock changes


connorjclark:
If you advance your clock a bunch (to simulate user computer sleeping -> resuming), `timer_thread_proc` will get a huge time delta, and try to process every tick. This has bad results for the application's timer functions.

https://github.com/liballeg/allegro5/blob/master/src/timernu.c#L96-L105

For my application this hangs everything, as this retains the timer mutex / fires an important callback very irregularly, which blocks my app from continuing.

I confirmed this on Mac. I got user reports for this issue on Windows when their computer came back from hibernation and they attempted to resume the app.

To fix, `timer_thread_proc` should either:

1) use a monotonic clock (not the real clock, [as unix backend does](https://github.com/liballeg/allegro5/blob/549f189df3408bba27524763289545f20771c3b0/src/unix/utime.c#L73)). 
2) do the above, but by directly changing al_get_timer() (instead of using a separate timer impl)
3) or, cap the time delta to some reasonable maximum

(2) seems best to me, but it is a breaking change. Though, it should be considered a bug fix, as the documentation states:

> Return the number of seconds since the Allegro library was initialised. The return value is undefined if Allegro is uninitialised. The resolution depends on the used driver, but typically can be in the order of microseconds.

https://www.allegro.cc/manual/5/al_get_time

 so it should have been monotonic in the first place.

SiegeLord:
We used to use monotnic clock, but it caused issues with pthread: https://github.com/liballeg/allegro5/commit/de07380e0c0b372e9996a6996733e5ed4d439eee

pedro-w:
Can this be closed now that #1511 is merged?

PR #1511, opened 2023-11-17
Limit time interval for when system clock changes

connorjclark:
a few commits here:

1) [osx > 10.2] use `CLOCK_MONOTONIC_RAW` instead of `CLOCK_REALTIME`, so al_get_time does not go backwards or jump forwards in time. Still respects time spent suspended. May be able to do for linux too, but I can't verify right now.

2) bound the time interval in timer control thread to `(0, 10s)`, to prevent clock shifts from seriously breaking an application. May still result in sudden jumps (for example, it causes a huge ruckus for my app's midi player, a shockingly loud amount of notes are played at once). I tried bounding by 1s also and it seemed fine too but I'm afraid to go low here.

3) [osx > 10.2] in `timernu.c` use `CLOCK_UPTIME_RAW` instead of `al_get_time()` so that suspended time is not used

ref #1510 (reduces problem for all platforms, correct behavior for timers for osx, but all other platforms still have not-ideal behavior wrt clock shifts)

### DIFF
--- a/src/timernu.c
+++ b/src/timernu.c
@@ -59,6 +59,45 @@ static _AL_THREAD * volatile timer_thread = NULL;
 static ALLEGRO_COND *timer_cond = NULL;
 static bool destroy_thread = false;
 
+// Allegro's al_get_time measures "the time since Allegro started", and so
+// does not ignore time spent in a suspended state. Further, some implementations
+// currently use a calendar clock, which changes based on the system clock.
+// However, the timer control thread needs to ignore suspended time, so that it always
+// performs the correct number of timer ticks.
+// This has only been verified for Allegro, so for now fallback to old
+// behavior on other platforms. `_al_timer_thread_handle_tick` will bound
+// the interval to a reasonable value to prevent other platforms from behaving
+// poorly when the clock changes.
+// See https://github.com/liballeg/allegro5/issues/1510
+// Note: perhaps this could move into a new public API: al_get_uptime
+
+#if defined(ALLEGRO_MACOSX) && MAC_OS_X_VERSION_MIN_REQUIRED >= 101200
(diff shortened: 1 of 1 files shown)


src/unix/utime.c
```diff
@@ -70,7 +77,7 @@ double _al_unix_get_time(void)
    struct timespec now;
    double time;
 
-   _al_clock_gettime(CLOCK_REALTIME, &now);
```
allefant, line 73: This would change the behavior of al_get_time. The documentation says: "Return the number of seconds since the Allegro library was initialised."

There is no room for interpreting that as skipping hibernation etc. Not saying anything about the other changes in this PR, but al_get_time is real time. It would potentially break a lot of code if that is changed.

allefant: True that sounds like the correct one to me.

SiegeLord:
Looks great, thanks! I like the clamping trick.

src/timernu.c
```diff
@@ -59,6 +59,45 @@ static _AL_THREAD * volatile timer_thread = NULL;
 static ALLEGRO_COND *timer_cond = NULL;
 static bool destroy_thread = false;
 
+// Allegro's al_get_time measures "the time since Allegro started", and so
+// does not ignore time spent in a suspended state. Further, some implementations
+// currently use a calendar clock, which changes based on the system clock.
+// However, the timer control thread needs to ignore suspended time, so that it always
+// performs the correct number of timer ticks.
+// This has only been verified for Allegro, so for now fallback to old
```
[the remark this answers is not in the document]
SiegeLord: Fixed in https://github.com/liballeg/allegro5/commit/9ced31b85310c307c19b0edab6b151e05aea4d3e

PR #1513, opened 2023-11-25
Fix link in README.txt

Jimmy-Hu:
Fix link in README.txt


SiegeLord:
Whoops, I've fixed the redirect on the DNS side. Thanks for reporting this!

PR #1521, opened 2023-12-20
Android: add a DeleteLocalRef call to android_get_monitor_info()

alemart:
This patch adds a `DeleteLocalRef` call to `android_get_monitor_info()`.


SiegeLord:
Thanks!

PR #1522, opened 2023-12-20
Android: map ALLEGRO_TEMP_PATH to the application cache

alemart:
In Allegro 5.2.9 for Android, `ALLEGRO_TEMP_PATH` is a synonym of `ALLEGRO_USER_DATA_PATH` and of other paths (see the code snippet below). Currently, `ALLEGRO_TEMP_PATH` is mapped to the app-specific directory for persistent files located in internal storage - it relies on `getFilesDir()`. On Android, the app's cache directory is meant to be used for temporary storage instead ([see docs](https://developer.android.com/training/data-storage/app-specific#internal-create-cache)).

This patch modifies `ALLEGRO_TEMP_PATH` so that it is mapped to the internal cache. No additional permissions are required.

https://github.com/liballeg/allegro5/blob/9ced31b85310c307c19b0edab6b151e05aea4d3e/src/android/android_system.c#L565-L572

### DIFF
--- a/android/gradle_project/allegro/src/main/java/org/liballeg/android/AllegroActivity.java
+++ b/android/gradle_project/allegro/src/main/java/org/liballeg/android/AllegroActivity.java
@@ -117,6 +117,11 @@ public class AllegroActivity extends Activity
       return getFilesDir().getAbsolutePath();
    }
 
+   String getTempDir()
+   {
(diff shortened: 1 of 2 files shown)


SiegeLord:
Makes sense, thanks!

Issue #1456, opened 2023-07-30
Fullscreen problems with newer Linux versions


codeflorist:
a user of my Dune Dynasty branch has reported a problem with fullscreen and newer linux versions here:
https://github.com/gameflorist/dunedynasty/issues/2

> I'm on Linux (Zorin 16) and I've been having some issues I'd like to report. I had these issues even before your branch.
> 
> Starting the game fullscreenwindow I can't maximize the game with alt-enter. It just doesn't stretch the window and I get a blank black screen
> 
> Starting the game fullscreen my taskbar at the bottom of the screen always shows over the game

is this something you can confirm?

SiegeLord:
The development version has some fixes for fullscreen windows. I'll attempt to compile the game with the newer version and see if that helps.

SiegeLord:
Hmm, so far I have been unable to reproduce the issue with either old or new version of Allegro on Ubuntu 22.04, Gnome/KDE Wayland/X11. Let me try the Linux version that user had and see if it's distro specific somehow.

codeflorist:
@SiegeLord Thanks a lot for your efforts!

SteveFosdick:
See pull request: https://github.com/liballeg/allegro5/pull/1524

PR #1524, opened 2024-01-03
Fix failure to enter full-screen mode on Linux with native dialog.

SteveFosdick:
Since the native dialog addon has started overriding the call to al_set_display_flag, (commit cc5c63cadfe1a49bf3d5d34153f284e4d71c34ad), setting the flag ALLEGRO_FULLSCREEN_WINDOW does nothing and the call returns false.  This appears to be mistake as a function xgtk_set_fullscreen_window is available in the file gtk_xgtk.c

This commit implements processing of the ALLEGRO_FULLSCREEN_WINDOW flag by calling this function.

A possibly relevant issue in the Allegro5 GitHub is https://github.com/liballeg/allegro5/issues/1456

### DIFF
--- a/addons/native_dialog/gtk_xgtk.c
+++ b/addons/native_dialog/gtk_xgtk.c
@@ -490,6 +490,9 @@ static bool xgtk_set_display_flag(ALLEGRO_DISPLAY *display, int flag, bool onoff
 
          return true;
       }
+      case ALLEGRO_FULLSCREEN_WINDOW:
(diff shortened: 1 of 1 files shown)


SiegeLord:
Awesome! Thanks for finding this. This fix needs a bit of work, since the system mutex appears to be locked twice which shouldn't be done (once in `xgtk_set_fullscreen_window` and once in the calling `xdpy_set_display_flag`. I'll investigate.

PR #1535, opened 2024-02-21
Prefill ALLEGRO_AUDIO_STREAMs with data in docstrings/examples/acodec.

SiegeLord:
The audio system will attempt to grab data from a stream *before* it has sent any events that it needs fragments. This causes a bit of lag, and also log spam as the system complains about having no data.

This change adjusts most places to prefill the streams with data to avoid this issue. It should be safe, and in fact beneficial as we should get a few less sample-worths of lag when using
`al_play/load_audio_stream`.


SiegeLord:
The data needs to come from somewhere, it's not appropriate for it to always be silence. So, there's no case where you don't want to prefill, but there's also no way to have a default prefilling code.

An alternative to this would be to somehow have stream start out as paused, have it send the event, have it unpause when it gets the first fragment. It's not obvious that that's a better solution, since it introduces some additional lag, in principle.

PR #1536, opened 2024-02-24
Free allocated memory during joystick initialization

### LINKED ISSUE #1534: Memory leak in MacOS joystick code
The static analyzer tells me there's a leak here:
https://github.com/liballeg/allegro5/blob/edcc50e85036e95474a31a7c75089b6b7653a4e5/src/macosx/hidjoy.m#L363-L375
`device_arr` is calloc'd and never freed.
Probably not a big deal as it's called only once by `init_joystick`

pedro-w:
Fixes #1534


SiegeLord:
Thanks!

PR #1539, opened 2024-02-26
Example program for audio recorder

aldrik-ramaekers:
An example program I wrote to test the audio recorder interface

https://github.com/liballeg/allegro5/assets/22401766/e9d55b64-3380-4622-a3f7-5cb5c7da7488


SiegeLord:
Thanks. How does this differ from `ex_record`?

aldrik-ramaekers:
.. did not realize we already have ex_record 😁

PR #1540, opened 2024-02-27
Fix building reproducible

gusnan:
The allegro-doc package doesn't build reproducible - these commits fix the "Last updated:" message on each html page by using the SOURCE_DATE_EPOCH build timestamp (more info here: https://reproducible-builds.org/docs/source-date-epoch/ ) and the other commit fix sorting example documents which are referenced alongside functions.

Original bug report in Debian here: https://bugs.debian.org/1064648 with links to more information regarding the problem.

https://reproducible-builds.org

### DIFF
--- a/docs/scripts/insert_timestamp.c
+++ b/docs/scripts/insert_timestamp.c
@@ -7,6 +7,9 @@
 #include <stdio.h>
 #include <time.h>
 #include <string.h>
+#include <errno.h>
+#include <limits.h>
+#include <stdlib.h>
 #include "dawk.h"
 
 int main(int argc, char **argv)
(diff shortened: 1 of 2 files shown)


docs/scripts/scan_examples.c
```diff
@@ -81,9 +90,13 @@ int main(int argc, char* argv[])
    lookup = calloc(argc, sizeof(lookup_t));
    for (j = 0; j < argc; ++j) {
       lookup[j].index = j;
+      int len = strlen(argv[j]);
```
SiegeLord, line 93: `lookup[j].filename = strdup(argv[j]);` ?

docs/scripts/insert_timestamp.c
```diff
@@ -15,7 +18,35 @@ int main(int argc, char **argv)
...
    char const *version = "unknown";
 
-   time(&now);
+   char *source_date_epoch;
+   unsigned long long epoch;
+   char *endptr;
+
+   source_date_epoch = getenv("SOURCE_DATE_EPOCH");
+   if (source_date_epoch) {
+      errno = 0;
```
SiegeLord, line 27: Can we just use `sscanf`?

```c
uint64_t epoch;
char trailing;
int num_parsed = sscanf(source_date_epoch, "%"SCNu64"%c", &epoch, &trailing);
if (num_parsed != 1) // some error handling
```

SiegeLord:
Thanks, this is great work. I suggested some possible simplifications (the `strdup` is probably a clear improvement, but `sscanf` could be argued against).

docs/scripts/scan_examples.c
pedro-w: Can't you just store the `argv[i]` pointer, no need to make a copy of the string it points to?

gusnan: @SiegeLord That seems very reasonable yes - and I have tested your method too now @pedro-w - that seems even better. I'll push to my branch here using this instead. Thanks!

docs/scripts/insert_timestamp.c
gusnan: This code was simply copied from reproducible-builds.org with just tiny tweaks - unfortunately the site is down right now.

gusnan: Now the site seems to work again - This was simply copied from https://reproducible-builds.org/docs/source-date-epoch/  - See the C example a bit down on that page.

docs/scripts/scan_examples.c
```diff
@@ -54,10 +54,19 @@ typedef struct {
    int index;
    /* How many APIs are used in each example */
    int count;
+   /* Store a pointer to the file string for comparison purposes */
+   char *filename;
 } lookup_t;
 static lookup_t *lookup;
+static int number_of_items = 0;
```
SiegeLord, line 61: No need for this anymore.

SiegeLord:
Thanks again!

PR #1542, opened 2024-03-10
add al_set_audio_output_device for audio output device selection

aldrik-ramaekers:
Currently only implemented on dsound and aqueue backend. I also extended ex_audio_simple for easy testing.


SiegeLord:
Thanks!

It looks like for AQueue all the voices remain valid, but what about DirectSound? I see you re-create it for the default voice, but what about the others? Maybe we should maintain a vector of voices to update them. I'll take a look at what the other backends can handle (although I'm sure re-creating the voices works no matter what).

aldrik-ramaekers:
It would also possible to let the user specify the voice to update e.g al_set_audio_output_device_for_voice(device, voice).
This would make sense because according to the docs a voice represents an audio device on the system.

SiegeLord:
That would certainly be ideal, if it's possible, but we'd need to investigate that it's an option for all backends.

aldrik-ramaekers:
Just glancing at the code dsound, aqueue and SDL2 backend will only require minor changes.

for PulseAudio & ALSA the buffer fill routine seems to not lock mutex while writing so might have to do something like ``stop voice -> block main thread untill routine stopped -> set new handle & params -> start voice``

openal, opensl, oss do not have device iteration yet so not looking at it for now.

recreating the voice is probably the easiest way to go, but at that point you might as well just let the user call a function like ``al_create_voice_with_output_device`` and scrap ``al_set_audio_output_device_for_voice``.

user code would then look something like:
```
voice = al_create_voice_with_output_device(device)
al_attach_mixer_to_voice(al_get_default_mixer(), voice)
al_set_default_voice(voice)
```

I'm fine with both methods

PR #1543, opened 2024-03-20
Fix multiarch install location for cmake package config

gusnan:
Debians Multi-arch-hinter gives warnings for the cmake package config files installation directories - this should fix this. (For a comparison, checkout the folders `/usr/lib/cmake/` and` /usr/lib/x86_64-linux-gnu/cmake/` - there's a massive amount more using the x86_64-linux-gnu folder).


SiegeLord:
Do you know which version of cmake this became available? The internet also says that this comes from `include(GNUInstallDirs)`, which we don't explicitly include anywhere.

gusnan:
Yeah, it looks to me it simply comes from GNUInstallDirs, so I guess the specific variable has been in cmake since that (which looks like around 3.0, somewhere around 2018). But you are right, I guess an include of GNUInstallDirs wouldn't hurt - I'll push a commit fixing this here.

PR #1551, opened 2024-04-11
Add OpenMPT support

SiegeLord:


pedro-w:
How are you testing this? I tried using `ex_acodec` with some .mod files I have (and know to be ok) but the error dialog said "Could not load sample from 'XXX.mod'!" (nor did it work with the DUMB renderer on released version of A5 so I guess it's `ex_acodec` ?)

SiegeLord:
I compile it with both enabled, but hack it to use openmpt to load `.xm` files rather than DUMB. I generally use `ex_stream_file` with `--loop` or not. `ex_stream_seek` is also pretty good.

At the moment, I'm struggling to get looping to work, I don't really understand how openmpt handles looping (the only relevant APIs are `openmpt_module_get_ctls` with `play.at_end` and `openmpt_module_set_repeat_count`).

PR #1554, opened 2024-04-23
don't need to typedef struct in C++

ReiquelApplegate:
You don't need to typedef struct in C++.

Caution: I don't have a way to run the allegro tests yet.


SiegeLord:
Thanks!

PR #1564, opened 2024-06-16
Fix opaque black background pixels in explosion sprites.

tstoeter:
In the explosion sprites, make black pixels fully transparent, which otherwise block parts of the scene, e.g. stars and ship. Also make all other pixels semi-transparent, which appeared nicer.


SiegeLord:
Thanks, it does look nicer!

PR #1566, opened 2024-06-18
Fixes issue #1547 for skater and shooter demos: Segfault after failed audio subsystem installation on Linux.

tstoeter:


SiegeLord:
Is there a desire to have the game playing in silence? We could just exit the game if audio failed to install.

tstoeter:
Of course it is only half the fun without music and sound. But I could now play the games on a computer that had no sound card installed.

SiegeLord:
Merged as 0797c22d12a3933ea631ba7c6febbb77f05247e0.

Belated thanks!

PR #1568, opened 2024-07-22
Fix iPhone build broken by fbb3f0c.

NewCreature:
Building for iPhone was failing on `src/opengl/extensions.c`. I traced the failure back to a change in commit fbb3f0c.

In order to keep that workaround functioning, I changed the `#if defined ALLEGRO_MACOSX || defined __APPLE__` to `#if defined __APPLE__ && !defined ALLEGRO_IPHONE`.

This fix ensures the proper OpenGL header is included if we are building for MacOS (including the SDL port which disables the `ALLEGRO_MACOSX` platform definition) and it is not included if we are building for iPhone.

### DIFF
--- a/src/opengl/extensions.c
+++ b/src/opengl/extensions.c
@@ -36,7 +36,7 @@
    #include "allegro5/internal/aintern_xsystem.h"
 #endif
 
-#if defined ALLEGRO_MACOSX || defined __APPLE__
+#if defined __APPLE__ && !defined ALLEGRO_IPHONE
 #include <OpenGL/glu.h>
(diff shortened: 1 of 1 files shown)


SiegeLord:
Are you referencing the right commit? Did you mean 74edb77851bd7e1820bed115e3ec1397aa2142ae?

Looks good either way (I don't have an iPhone handy to test this... I wonder if we can get a iPhone CI for this).

NewCreature:
Yeah, not sure how I got the wrong commit. I had 74edb77 pulled up while I was working on the patch.

PR #1570, opened 2024-08-05
FIx Building on MacOS with SDK versions < 12.0

NewCreature:
As described in the commit message, Apple renamed `kAudioObjectPropertyElementMaster` to `kAudioObjectPropertyElementMain` in the MacOS v12.0 SDK. This change made building Allegro with older SDKs fail.

This pull request includes a commit that allows Allegro to be built with SDKs < v12.0 by using the old variable name if the SDK being used is < v12.0.


SiegeLord:
Sigh...

Thanks.

PR #1579, opened 2024-09-09
[osx] use CLOCK_REALTIME for init_timeout

connorjclark:
#1511 changed osx clocks to use CLOCK_MONOTONIC_RAW, but that is an issue for the timeout structure because pthread expects an absolute system time. This resulted in timed event calls returning instantly.

See https://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_cond_timedwait.html#:~:text=Timed%20Wait%20Semantics

### DIFF
--- a/src/unix/utime.c
+++ b/src/unix/utime.c
@@ -108,7 +108,7 @@ void _al_unix_init_timeout(ALLEGRO_TIMEOUT *timeout, double seconds)
 
    ASSERT(ut);
 
-   _al_clock_gettime(CLOCK_ID, &now);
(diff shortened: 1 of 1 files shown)


SiegeLord:
I couldn't reproduce the instant timeouts (I used `ex_timedwait`, suspended the system, resumed and then saw if the timeouts were affected... they were not), but your reasoning seems reasonable.

SiegeLord:
I tried the debugger, but still no luck. I don't mind merging this if it helps on your system though.

PR #1583, opened 2024-09-27
Add "al_get_render_state"

MarkOates:
During some debugging, I noticed I didn't have access to the current render state, which would be helpful. This would be particularly useful in instances where different parts of a game system manage render state differently, and checking for possible leaky render state. The values for it seem to be stored in `ALLEGRO_DISPLAY`.

## Some points to be aware of in this PR:
- In the refman/, not sure what "Since 5.2.11" to put it under.
- Not sure about a return value or behavior when there is no current `display`. This implementation just has it return `0` with no warning or anything.

### DIFF
--- a/docs/src/refman/graphics.txt
+++ b/docs/src/refman/graphics.txt
@@ -2046,9 +2046,9 @@ See also: [al_set_render_state]
 Returns one of several render attributes; see [ALLEGRO_RENDER_STATE]
 for details.
 
-This function will return 0 when there is no current display.
+This function will return -1 when there is no current display.
 
-Since: 5.2.11
+Since: 5.2.10
 
 See also: [al_set_render_state], [ALLEGRO_RENDER_STATE], [ALLEGRO_RENDER_FUNCTION],
 [ALLEGRO_WRITE_MASK_FLAGS]
--- a/docs/src/refman/graphics.txt
+++ b/docs/src/refman/graphics.txt
@@ -2040,6 +2040,20 @@ Since: 5.1.2
 
 See also: [al_set_render_state]
 
+
+### API: al_get_render_state
+
+Returns one of several render attributes; see [ALLEGRO_RENDER_STATE]
+for details.
+
+This function will return 0 when there is no current display.
+
+Since: 5.2.11
+
+See also: [al_set_render_state], [ALLEGRO_RENDER_STATE], [ALLEGRO_RENDER_FUNCTION],
+[ALLEGRO_WRITE_MASK_FLAGS]
+
+
 ### API: al_set_render_state
 
 Set one of several render attributes; see [ALLEGRO_RENDER_STATE]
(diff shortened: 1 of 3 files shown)


src/display.c
```diff
@@ -602,6 +602,41 @@ void al_acknowledge_drawing_resume(ALLEGRO_DISPLAY *display)
...
 }
 
+/* Function: al_get_render_state
+ */
+int al_get_render_state(ALLEGRO_RENDER_STATE state)
+{
+   ALLEGRO_DISPLAY *display = al_get_current_display();
+
+   if (!display)
+      return 0;
```
SiegeLord, line 612: Let's do -1 instead of 0.

src/display.c
```diff
@@ -602,6 +602,41 @@ void al_acknowledge_drawing_resume(ALLEGRO_DISPLAY *display)
...
+         return display->render_state.depth_function;
+         break;
+      case ALLEGRO_ALPHA_FUNCTION:
+         return display->render_state.alpha_function;
+         break;
+      case ALLEGRO_ALPHA_TEST_VALUE:
+         return display->render_state.alpha_test_value;
+         break;
+      default:
+         ALLEGRO_WARN("unknown state to retrieve: %d\n", state);
```
SiegeLord, line 634: We usually capitalize these.

Also, this should be ALLEGRO_ERROR.

SiegeLord (changes requested):
Seems like a nice addition, thanks!

src/display.c
MarkOates: ✅ done

src/display.c
MarkOates: ✅ done

There was another uncapitalized message in `al_set_render_state` I fixed as well in b1a67f8.

docs/src/refman/graphics.txt
```diff
@@ -2040,6 +2040,20 @@ Since: 5.1.2
 
 See also: [al_set_render_state]
 
+
+### API: al_get_render_state
+
+Returns one of several render attributes; see [ALLEGRO_RENDER_STATE]
+for details.
+
+This function will return 0 when there is no current display.
```
SiegeLord, line 2049: Please update this too.

docs/src/refman/graphics.txt
```diff
@@ -2040,6 +2040,20 @@ Since: 5.1.2
...
+
+Since: 5.2.11
```
SiegeLord, line 2051: Since should be 5.2.10. 5.2.9 is the currently released version, while 5.2.10 is the version on master, which will be released next.

docs/src/refman/graphics.txt
MarkOates: ✅ done

docs/src/refman/graphics.txt
MarkOates: ✅ done

SiegeLord:
Thanks!

MarkOates:
Sweet! I missed the merge notification on this one.

I've already had several times where this would be useful. Thank you!

Issue #1600, opened 2024-12-03
color emoji support


allefant:
FreeType 2.10+ supports loading of color emojis, but a special flag is needed: https://freetype.org/freetype2/docs/reference/ft2-glyph_retrieval.html#ft_load_color

Looks like all we need to support them in Allegro is that flag and we should make copy_glyph_* aware of possible different source bitmap format.

PR #1601, opened 2024-12-05
enable color emoji support in freetype

allefant:
For issue https://github.com/liballeg/allegro5/issues/1600, this PR adds color emoji support.

The only real change required is to add the FT_LOAD_COLOR flag to FT_Load_Glyph.

However this clashes with the no_bitmap flag, so selectively have to remove that when detecting a bitmap font. Also, for bitmap fonts, FreeType requires a call to select the size (for now we always select size 0, I have yet to find a bitmap font in TTF format with multiple sizes). Finally, our pixel copy functions needed to be made aware of FT sometimes using a 32-bit instead of 8-bit source bitmap format.

I also added an example font with color emoji (selection of some NotoColorEmoji glyphs), with appropriate license notice. And an example to use that font.


SiegeLord:
Lovely

PR #1602, opened 2024-12-16
Fix MAC_OS_X_VERSION_MIN_REQUIRED checks to use correct six-digit values

Rhys-T:
The MAC_OS_VERSION_* constants jumped from 1090 to 101000, so checks for >= 1100 were passing in older versions than they should. Since Nixpkgs builds against 10.12 by default (on x86_64), this was causing Allegro 5.2.10.0 to fail to build because it couldn't find `UniformTypeIdentifiers/UniformTypeIdentifiers.h`.


SiegeLord:
Whoops. Thanks!

SiegeLord:
This will go in 5.2.10.1

PR #1604, opened 2024-12-23
Fix memory leak in xdpy_set_system_mouse_cursor()

Helodity:
Currently, Allegro does not delete the system cursor when a new one is created in X11.

This fix frees the system cursor whenever a new cursor is set by keeping track of whether the current cursor is a system cursor or a user set one.

See [here](https://canary.discord.com/channels/993415281244393504/1060574341584125982/1319443549993046078) for more information about this.


SiegeLord:
Looks good, thank you!

Issue #1533, opened 2024-02-15
Fix the joyticks


SiegeLord:
Joyticks have been subtly and not so subtly broken for years. If we want to have a joystick feature at all, it should work well.

SDL has established a standard for joystick interpretation, and other libraries have sought to emulate it. It makes sense to do the same. If the SDL API appears to be deficient, at least our API should be a strict improvement over it.

Tasks:

- Investigate the SDL/GLFW APIs for joysticks, see if it makes sense to map to Allegro's.
- Adjust the code for all 3 major platforms to start. Use the `joystick_version` compatibility switch to maintain compatibility with the older code.

ReiquelApplegate:
https://www.libsdl.org/release/SDL-1.2.15/docs/html/joystick.html

PR #1616, opened 2025-01-16
Hotplug fixes

SiegeLord:
Do a pass on joystick hotplugging.
- MacOS wasn't reusing joystick slots
- Linux was getting duplicate joysticks due to reading from
  /dev/input/by-path
- Windows XInput wasn't re-initializing joystick names
- SDL was completely broken

Also, I redid ex_joystick_hotplugging to be more useful/exciting.

See #1533


pedro-w:
Not a big deal but some of the changed files don't seem to be related to this. (unless they are in some subtle way!)

> Also, I redid ex_joystick_hotplugging to be more useful/exciting.

👍👍👍👍👍

SiegeLord:
> Not a big deal but some of the changed files don't seem to be related to this. (unless they are in some subtle way!)
Maybe you're referring to the files changed by the first commit, which is in separate PR #1612. Github still has no way to do chained PRs, it seems.

pedro-w:
> Maybe you're referring to the files changed by the first commit, which is in separate PR #1612. 

Yes, that's right. There are changes to files in `acodec` for example. I just wanted to check that it wasn't some unrelated update that had got dragged into this PR.

SiegeLord:
Those changes are from the other PR: since the new joystick API is unstable, a bunch of files which incidentally include joysticks need to define the internal unstable macro... it's very annoying, I wish for a better solution.

PR #1617, opened 2025-01-18
Fixed reinitialization of the window resize event thread

kwellwood:
This merge fixes a problem that occurs when initializing and running allegro multiple times in the same application session on Windows, i.e. during unit testing.

Vars end_resize_event_thread and resize_event_thread_ended were uninitialized in _al_win_init_window(). This prevented the resize event thread from running a second time because (end_resize_event_thread == true) as soon as it began.

### DIFF
--- a/src/win/wwindow.c
+++ b/src/win/wwindow.c
@@ -1046,6 +1046,8 @@ int _al_win_init_window(void)
 
    resize_event_thread_mutex = al_create_mutex();
    _al_vector_init(&resizing_displays, sizeof(ALLEGRO_DISPLAY_WIN *));
+   end_resize_event_thread = false;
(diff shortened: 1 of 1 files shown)


pedro-w:
This will probably work. 

As an aside, I wonder why we didn't use `al_join_thread` (in `_al_win_shutdown_window`) and `al_get_thread_should_stop` (in `resize_event_thread_proc`) instead?

SiegeLord:
Thanks!

> As an aside, I wonder why we didn't use al_join_thread (in _al_win_shutdown_window) and al_get_thread_should_stop (in resize_event_thread_proc) instead?

It's not using Allegro threads, but native Windows threads. I don't know why.

PR #1630, opened 2025-04-01
Fix Issue#1629: Add locks to avoid stale clipboard text.

SteveFosdick:
al_get_clipboard_text was returning stale text, i.e. from the previous call because xdpy_get_clipboard_text was not waiting for the SelectionNotify event.  This was in turn due to not have locked the mutex passed to pthread_cond_timedwait.

This commit fixes this by locking the mutex.


SiegeLord:
This makes sense. Even though I couldn't reproduce this, it's clear that the current code was violating the preconditions of `pthread_cond_timedwait`. Thanks!

PR #1642, opened 2025-06-05
Update mouse.txt

j-w-c-b:
Add ALLEGRO_MOUSE_STATE.display to documentation since it is public in the header and documented there (and has been since 2008)


SiegeLord:
Thanks! Usually it's the opposite, if it's not documented, it's not public regardless, but in this case I don't see why it shouldn't be.

PR #1647, opened 2025-06-28
Remove GETTER/SETTER from generated docs

j-w-c-b:
This patch ends prototype scanning during doc generation when a line starts with GETTER or SETTER. This shows up for example here: https://liballeg.org/a5docs/trunk/state.html#al_get_errno


SiegeLord:
Thanks!

PR #1650, opened 2025-07-01
Allow setting doubles in shaders.

### LINKED ISSUE #1649: Add a way to pass doubles into shaders

SiegeLord:
Fixes #1649


SiegeLord:
For Direct3D, we probably should be using https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/asdouble since it doesn't support passing doubles directly. I.e. we'd pass a pair of integers, and document this behavior. Not sure how this works for arrays.

PR #1656, opened 2025-07-25
Fix #1618: ALLEGRO_FULLSCREEN_WINDOW not working with al_set_display_…

### LINKED ISSUE #1618: ALLEGRO_FULLSCREEN_WINDOW doesn't work with al_set_display_flags() under ubuntu
I'm writing an Emulator and can't change to full screen mode using ALLEGRO_FULLSCREEN_WINDOW passing the argument to al_set_display_flags().
When i'm going to change to fullscreen mode i do a: 

`al_set_display_flag(disp, ALLEGRO_FULLSCREEN_WINDOW, true);`

I'm using last allegro release which is allegro-5.2.10.1.tar.gz. 

In windows it just work perfect. There seems to be a problem in my ubuntu linux with NVIDIA drivers (i'm using the last release)

Anyway with ALLEGRO_FRAMELESS and ALLEGRO_MAXIMIZED there are no problems in my linux.

Is a bug problem or i am making something wrong?

killerdevildog:
…flag() on Linux

The issue was in xdpy_set_fullscreen_window_default() function which was always calling _al_xwin_set_fullscreen_window() with value 2 (toggle mode) instead of passing the correct value based on the onoff parameter.

Changed the call to pass:
- 1 when enabling fullscreen window mode (onoff=true)
- 0 when disabling fullscreen window mode (onoff=false)

This matches the behavior of the GTK implementation and the expected X11 _NET_WM_STATE protocol where:
- 0 = remove property
- 1 = add property
- 2 = toggle property

The toggle mode (2) was unreliable because it depended on the current state being synchronized, but the comment even mentions that this flag may not be in sync with reality.

### DIFF
--- a/src/x/xdisplay.c
+++ b/src/x/xdisplay.c
@@ -1388,7 +1388,10 @@ static void xdpy_set_fullscreen_window_default(ALLEGRO_DISPLAY *display, bool on
       _al_mutex_lock(&system->lock);
 
       _al_xwin_reset_size_hints(display);
-      _al_xwin_set_fullscreen_window(display, 2);
+      /* Pass the correct value instead of always toggling:
+       * 0 = disable fullscreen, 1 = enable fullscreen
+       */
+      _al_xwin_set_fullscreen_window(display, onoff ? 1 : 0);
       /* XXX Technically, the user may fiddle with the _NET_WM_STATE_FULLSCREEN
        * property outside of Allegro so this flag may not be in sync with
        * reality.


SiegeLord:
Thanks. Does this fix the bug on your system? It doesn't work for me in KDE 5.27.12, on X11.  Specifically, while the fullscreen is set correctly, when toggling it to non-fullscreen, it merely adds the window decoration and moves the window but does not change its size. With the current code, for me, setting ALLEGRO_FULLSCREEN_WINDOW to false resizes the window as well.

killerdevildog:
turns out it was a partial fix, @SiegeLord  please git checkout my branch and see if its fixed now for you, please build allegro5/build/examples and run ./ex_fs_window I think this is a solid fix.

SiegeLord:
Thanks. And you're saying this changed helps your case? As far as I can tell, the net effect of your two commits should be equivalent to the existing code, I'm actually confused how it can help.

SiegeLord:
Also, how are you reproducing the bug with `ex_fs_window`. Are you just toggling it and it doesn't match the state as displayed in text on the window?

killerdevildog:
@SiegeLord Thanks for reviewing. To address your confusion: the two commits together are not exactly equivalent to the existing code, and here's why.

First, regarding the issue you saw with the window not resizing when toggling back to windowed mode on KDE—that was likely due to the first commit, where I changed the call to _al_xwin_set_fullscreen_window to use explicit 1 (add) or 0 (remove) instead of 2 (toggle). As noted in the code comments and X11/EWMH specs, explicit add/remove can be unreliable across different window managers (some WMs may ignore them or handle them inconsistently, leading to behaviors like not resizing the window properly when removing the fullscreen state). That's why the second commit reverts to using toggle (2) for the X11 communication, which is more reliable and matches the original code's approach, but with an important improvement to the internal flag handling.

Why the original code wasn't working (bidirectional toggling failure): The root issue was the use of XOR (^=) for updating the internal display->flags. This assumes perfect synchronization between Allegro's flag and the actual X11/WM state at all times. However, as the XXX comment notes, the flag can desync due to external changes to _NET_WM_STATE_FULLSCREEN (e.g., user or WM interventions). In practice, this led to mismatches where toggling would only work in one direction—e.g., from fullscreen to windowed, but not back—because the XOR would flip the flag based on a potentially incorrect current state, causing subsequent calls to al_set_display_flag to misjudge whether a change was needed.

With the fix in the second commit:

    We still use toggle (2) to communicate with X11, ensuring reliable state changes across WMs (and restoring the resizing behavior you expect when disabling fullscreen).
    But for the internal flag, we now set it explicitly based on the requested onoff value: |= to enable, &= ~ to disable. This forces the internal state to match the intended state after the toggle, reducing the impact of any prior desync. Since xdpy_set_fullscreen_window_default is only called when a state change is detected (i.e., current != requested), the toggle flips the WM state to the desired one, and we align the internal flag accordingly.

This should fix the bidirectional toggling without introducing the reliability issues of explicit 0/1.

To reproduce the original bug with ex_fs_window (built from examples): Start in windowed mode, toggle to fullscreen with 'f'—it may work the first time. Then toggle back to windowed (works), but trying to toggle to fullscreen again fails (window stays windowed, but the text might indicate fullscreen due to desync). With this fix, toggling works reliably in both directions, and the state matches the displayed text.

Please test the updated branch—it should now resize properly on KDE when disabling, like the current master code, while fixing the original issue on Ubuntu/NVIDIA.

PR #1657, opened 2025-07-25
Fix maximized windows not being draggable

killerdevildog:
Set _NET_WM_ALLOWED_ACTIONS property to explicitly allow window movement even when maximized. This ensures that maximized windows can still be dragged and moved by the user in X11 window managers.

The fix adds the following allowed actions to all windows:
- _NET_WM_ACTION_MOVE (allows dragging maximized windows)
- _NET_WM_ACTION_RESIZE
- _NET_WM_ACTION_MINIMIZE
- _NET_WM_ACTION_MAXIMIZE_HORZ
- _NET_WM_ACTION_MAXIMIZE_VERT
- _NET_WM_ACTION_CLOSE

This property is set during window creation in xdpy_create_display_window() to ensure the window manager knows these actions are allowed regardless of the window's current state.

### DIFF
--- a/src/x/xdisplay.c
+++ b/src/x/xdisplay.c
@@ -276,6 +276,28 @@ static bool xdpy_create_display_window(ALLEGRO_SYSTEM_XGLX *system,
                    32, PropModeReplace,
                    (unsigned char *)&_NET_WM_WINDOW_TYPE_NORMAL, 1);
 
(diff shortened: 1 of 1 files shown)


SiegeLord:
Thanks. What WM does this not work in? It seems to work fine with the current code on my KDE system, but I can test others.

PR #1664, opened 2025-08-20
Update osxgl.m

MarekRatajczak2024:
Drag and drop for MacOS

### DIFF
--- a/src/macosx/osxgl.m
+++ b/src/macosx/osxgl.m
@@ -15,7 +15,6 @@
 *      See readme.txt for copyright information.
 */
 
-
 #include "allegro5/allegro.h"
 #include "allegro5/allegro_opengl.h"
 #include "allegro5/internal/aintern.h"
@@ -27,6 +26,7 @@
 #include "allegro5/platform/aintosx.h"
 #include "./osxgl.h"
 #include "allegro5/allegro_osx.h"
+
 #ifndef ALLEGRO_MACOSX
 #error something is wrong with the makefile
 #endif
@@ -65,9 +65,11 @@ enum {
 #define MINIMUM_WIDTH 48
 #define MINIMUM_HEIGHT 48
 
+
 /* by MAREK */
-#define FILE_URL_SIZE 254
-/* end by MAREK */
+static char **cfileURLs;
+int cfileURLs_n=0;
+/* */ //by MAREK
 
 /* Unsigned integer; data type only avaliable for OS X >= 10.5 */
 #if MAC_OS_X_VERSION_MIN_REQUIRED < 1050
@@ -216,9 +218,9 @@ void _al_osx_keyboard_was_installed(BOOL install) {
    /* This is passed onto the event functions so we know where the event came from */
    ALLEGRO_DISPLAY* dpy_ptr;
 }
-/* by MAREK */
+/* by MAREK* /
 - (id)initWithFrame:(NSRect)frameRect;
-/* end by MAREK */
+/* */ //by MAREK
 -(void)setAllegroDisplay: (ALLEGRO_DISPLAY*) ptr;
 -(ALLEGRO_DISPLAY*) allegroDisplay;
 -(void) reshape;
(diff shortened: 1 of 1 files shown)


src/macosx/osxgl.m
```diff
@@ -65,6 +65,10 @@
 #define MINIMUM_WIDTH 48
 #define MINIMUM_HEIGHT 48
 
+/* by MAREK */
+#define FILE_URL_SIZE 254
```
pedro-w, line 69: Is this used anywhere?

src/macosx/osxgl.m
```diff
@@ -292,8 +299,119 @@ void _al_osx_mouse_was_installed(BOOL install) {
...
+- (id)initWithFrame:(NSRect)frameRect
+{
+    self = [super initWithFrame:frameRect];
+    if (self) {
+        // Register for file URL drops (e.g., images)
+        [self registerForDraggedTypes:@[NSPasteboardTypeFileURL]];
+        // Or other types like NSStringPboardType, NSFileContentsPboardType, etc.
+    }
+
+    NSString *cachesDirectory = [NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES) objectAtIndex:0];
```
pedro-w, line 330: Is this variable used?

src/macosx/osxgl.m
```diff
@@ -292,8 +299,119 @@ void _al_osx_mouse_was_installed(BOOL install) {
...
+
+                _al_event_source_lock(es);
+                ALLEGRO_EVENT event;
+
+                event.drop.type = ALLEGRO_EVENT_DROP;
+                event.drop.timestamp = al_get_time();
+                event.drop.x = (int)mousePos.x;
+                event.drop.y = (int)mousePos.y;
+                NSLog(@"Event with file (%d/%d) : %@", n+1, (int)fileURLs.count, filePath);
+                event.drop.text = (char*) cfileURL;
```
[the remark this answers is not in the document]
pedro-w: From [Allegro docs](https://liballeg.org/a5docs/trunk/events.html#allegro_event_drop) I think you need to copy cFileURL into allocated memory, which the user program must free.

pedro-w:
I was glad to see this. I have added some comments, hopefully helpful

src/macosx/osxgl.m
MarekRatajczak2024: Oh, sorry, no, is not used anymore, there was another concept in my mind to send 1 single message and receive the event as a bunch of file names separated by \r\n so exactly like MacOS is doing, but eventually i changed minds and did at exactly like xdnd function for X11 in Allegro 5 is doing, so there is single message for each file name. So that definition is not used anymore. Remove it please.

src/macosx/osxgl.m
```diff
@@ -292,8 +299,119 @@ void _al_osx_mouse_was_installed(BOOL install) {
    });
 }
 
+/* by MAREK */
+@implementation NSString (Char)
+
+/*
+ * Convert a NSString to a char pointer
+*/
+-(char *)toChar{
```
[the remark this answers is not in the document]
MarekRatajczak2024: as in my previous comment, this comes from previous concept, and NSString now is converted directly to char array, no need to use that function. Delete it please.

src/macosx/osxgl.m
MarekRatajczak2024: i'm really sorry, it should be remarked with // like next rows below, what was a part of my tests i needed to divert errout to the file, originally in cachesDirectory, to check how events are received. so those 3 rows:

NSString *cachesDirectory = [NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES) objectAtIndex:0];

    // NSString *logPath = @"allegro5.log";
    // freopen([logPath cStringUsingEncoding:NSASCIIStringEncoding], "a+", stderr);


can be freely removed.

MarekRatajczak2024:
OK, i resent that file after all changes suggested. Now pointers are independent, and will exist until next drag n drop operation or display is closed.

src/macosx/osxgl.m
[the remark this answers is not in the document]
pedro-w: I think you just need to `al_malloc` each individual dropped file name, and it is the _application_'s responsibility to `al_free` it; see the `ex_drag_and_drop.c` for example:
https://github.com/liballeg/allegro5/blob/efb3129bb54ecfbfbc59984645b948d86766f23d/examples/ex_drag_and_drop.c#L183

pedro-w:
It seems to me that you don't need to track the memory allocated to file names at all. In the X version

https://github.com/liballeg/allegro5/blob/efb3129bb54ecfbfbc59984645b948d86766f23d/src/x/xdnd.c#L307-L314

Just use the convenience function `al_cstr_dup` and put it into the event, then the application frees it later.

The stuff with `cfileURLs` and `cfileURLs_n` is not needed. What do you think @SiegeLord ?

Also the X code calls `_al_event_source_needs_to_generate_event` which may need adding here too; if we generate events when there is nothing listening then memory will leak.

PR #1675, opened 2025-09-18
Restore previous GL_UNPACK_ROW_LENGTH

ARoxdale:
The first commit here is the actual change. It is possible that the original client GL_UNPACK_ROW_LENGTH might not be 0, so it is set back to the previous unpack value instead. 
The second commit is an attempt at adding a helper function to try to clarify the intent of the reverts. Please let me know if this is not in the preferred style.

I hope this is useful, but I have not worked with opengl or allegro code before, so feel free to decline this.

### DIFF
--- a/src/opengl/ogl_lock.c
+++ b/src/opengl/ogl_lock.c
@@ -670,6 +670,9 @@ static void ogl_unlock_region_nonbb_nonfbo(ALLEGRO_BITMAP *bitmap,
    unsigned char *start_ptr;
    GLenum e;
 
+   int previous_unpack_row_length = 0;
+   glGetIntegerv(GL_UNPACK_ROW_LENGTH, &previous_unpack_row_length);
+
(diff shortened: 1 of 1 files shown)


src/opengl/ogl_lock.c
```diff
@@ -662,14 +662,16 @@ static void ogl_unlock_region_nonbb_fbo_readwrite(ALLEGRO_BITMAP *bitmap,
    }
 }
 
-
```
SiegeLord, line 665: Please revert this, we do two newlines between function definitions.

SiegeLord (changes requested):
Thanks. Yeah, I think the second commit is a bit too different than the local style of that file. I think the function itself doesn't really aid readability, and, more importantly, the "reset_X" booleans are better than doing the inequality check, since it lets us make fewer OpenGL state calls, which are relatively expensive.

The first is fine, though.

Issue #1676, opened 2025-09-23
Make high level primitives have defined u/v coordinates.


SiegeLord:
It would be convenient for shaders to have them set to something reasonable. A good option seems to be the local pixel coordinates of the circumscribed rectangle for most 2D shapes, and some sort of length x width for lines.

Since, at the moment, they're undefined (at least in the few primitives I checked), this is a backwards compatible change.

PR #1677, opened 2025-09-26
Add u/v coordinates to high level primitives, excluding ribbons and splines

Helodity:
The U/V coordinates utilize the prim's local coordinates, though (0,0) varies depending on the shape.
- For shapes such as rectangles and triangles, (0,0) is the top left of the circumscribed rectangle. For non filled primitives, this does not offset the u/v by the thickness value. This was done to preserve alignment when drawing a filled rectangle and non filled rectangle in the same location.
- For ellipses, circles, and pie slices, (0,0) is the center of the primitive.
- For lines, (0,0) is the top left, though treated like a rotated rectangle.

The following image shows how the u/v coordinates appear (note that the shader scales the coordinates down to 0-1 for visibility):
- Top row: `al_draw_rectangle`, `al_draw_filled_rectangle`, `al_draw_filled_pieslice`
- Middle row: `al_draw_line` (0 and non zero thickness), `al_draw_triangle` (0 and non zero thickness), `al_draw_filled_triangle`, `al_draw_pieslice` (0 and nonzero thickness)
- Bottom row: `al_draw_filled_ellipse`, `al_draw_ellipse`, `al_draw_rounded_rectangle`, `al_draw_filled_rounded_rectangle`

image

This impacts #1676 , though ribbons and splines still need to be implemented before closing.

### DIFF
--- a/addons/primitives/high_primitives.c
+++ b/addons/primitives/high_primitives.c
@@ -331,8 +331,8 @@ void al_draw_filled_triangle(float x1, float y1, float x2, float y2,
 {
    ALLEGRO_VERTEX vtx[3];
 
-   float min_x = min(x1, min(x2, x3));
-   float min_y = min(y1, min(y2, y3));
+   float min_x = fmin(x1, fmin(x2, x3));
+   float min_y = fmin(y1, fmin(y2, y3));
 
    vtx[0].x = x1; vtx[0].y = y1; vtx[0].u = x1 - min_x; vtx[0].v = y1 - min_y;
    vtx[1].x = x2; vtx[1].y = y2; vtx[1].u = x2 - min_x; vtx[1].v = y2 - min_y;
--- a/addons/primitives/high_primitives.c
+++ b/addons/primitives/high_primitives.c
@@ -81,10 +81,10 @@ static float get_scale(void)
 void al_draw_line(float x1, float y1, float x2, float y2,
    ALLEGRO_COLOR color, float thickness)
 {
+   float len = hypotf(x2 - x1, y2 - y1);
    if (thickness > 0) {
       int ii;
       float tx, ty;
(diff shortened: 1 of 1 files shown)


addons/primitives/high_primitives.c
```diff
@@ -126,6 +126,10 @@ void al_draw_line(float x1, float y1, float x2, float y2,
 void al_draw_triangle(float x1, float y1, float x2, float y2,
    float x3, float y3, ALLEGRO_COLOR color, float thickness)
 {
+
+   float min_x = min(x1, min(x2, x3));
```
SiegeLord, line 130: These need to be `fmin`

addons/primitives/high_primitives.c
```diff
@@ -455,9 +449,14 @@ void al_calculate_arc(float* dest, int stride, float cx, float cy,
          for (ii = 0; ii < num_points; ii ++) {
             *dest =       r2 * x + cx;
             *(dest + 1) = r2 * y + cy;
+            *(dest + 3) = r2 * x;
```
SiegeLord, line 452: Unfortunately this isn't backwards compatible, we can't change what this function is doing, since existing users could have been just passing 2*num_points floats and no more.

This will need to be computed after the fact in al_draw_circle etc (by subtracting away cx/cy from x/y)

addons/primitives/high_primitives.c
```diff
@@ -432,6 +425,7 @@ void al_calculate_arc(float* dest, int stride, float cx, float cy,
    float c;
    float s;
    float x, y, t;
+   float min_x, min_y;
```
SiegeLord, line 428: Unused.

SiegeLord (changes requested):
This looks pretty good, thanks for making these changes!

PR #1680, opened 2025-10-24
Export CMake's build tree and relax pcx stride vs width check to account for padding

mlt:
Here are a couple of independent commits. I hope it is okay it is one PR.
1. I came across an issue with a tiny PCX file (width 7 px) having 1 byte padding. Let's relax the guard check to allow for padding.
2. I propose to export targets from a CMake build tree for use by an outside project. This eases debugging.

One may use `CMAKE_PREFIX_PATH` set to Allegro's build tree to build their projects without "installation". For example, it is possible to put the following in `CMakePresets.json` to use development version of Allegro directly from its build tree instead of the one from vcpkg:
```
{
  "version": 3,
  "configurePresets": [
    {
      "name": "some-preset"
      "cacheVariables": {
        "CMAKE_PREFIX_PATH": "C:/path/to/allegro"
        "VCPKG_MANIFEST_INSTALL": "OFF",
        "VCPKG_MANIFEST_MODE": "OFF"
      }
}
```
[While searching for a library, CMake will take a look](https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure) into `cmake` subfolder on Windows besides the root of a buildtree, hence rename of `generated` into `cmake` for convenience.

Note that there is an extra `;` after `BUILD_INTERFACE:` as CMake inserts project root there and I have no idea how to turn it off.

### DIFF
--- a/addons/image/pcx.c
+++ b/addons/image/pcx.c
@@ -55,7 +55,7 @@ ALLEGRO_BITMAP *_al_load_pcx_f(ALLEGRO_FILE *f, int flags)
 
    bytes_per_line = al_fread16le(f);
    /* It can only be this because we only support 8 bit planes */
-   if (bytes_per_line != width) {
+   if (bytes_per_line < width) {
       ALLEGRO_ERROR("Invalid bytes per line %d\n", bytes_per_line);
       return NULL;
    }


CMakeLists.txt
```diff
@@ -1261,6 +1261,14 @@ install(
   DESTINATION ${PKG_CONFIG_INSTALL_DIR}
 )
 
+# Export build tree to ease development and debugging
+target_include_directories(allegro INTERFACE
```
SiegeLord: Could you take a look at `cmake/Common.cmake` `add_our_library` and see if this should go there instead (alongside the other `target_include_directories` call)?

SiegeLord: Ok, I think I get what's going here. We only add it to allegro because addons link to allegro, so we don't really need to set it for all of them as well.

I think then what's missing is to make this conditional on ALLEGRO_MONOLITH not being defined (and then call this again for the monolith).

Does that make sense?

SiegeLord: If nothing else, this needs to be disabled if WANT_MONOLITH is on, since otherwise it gives an error that the `allegro` target doesn't exist.

SiegeLord (approved):
This looks good, thank you!

PR #1682, opened 2025-10-25
Also search in the minimp3 directory.

SiegeLord:
This is how vcpkg packages it.

### DIFF
--- a/cmake/FindMiniMP3.cmake
+++ b/cmake/FindMiniMP3.cmake
@@ -7,6 +7,7 @@
 # Look for the header file.
 find_path(MINIMP3_INCLUDE_DIRS
     NAMES minimp3.h minimp3_ex.h
+    PATHS minimp3
 )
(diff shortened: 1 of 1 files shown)


cmake/FindMiniMP3.cmake
```diff
@@ -7,6 +7,7 @@
 # Look for the header file.
 find_path(MINIMP3_INCLUDE_DIRS
     NAMES minimp3.h minimp3_ex.h
+    PATHS minimp3
```
mlt: PATH_SUFFIXES might be better? That way there is some flexibility when pointing MINIMP3_INCLUDE_DIRS manually either directly to downloaded header or to a proper install location.
That is what FindDirectX.cmake is doing… and [missing directxsdk in there](https://github.com/mlt/allegro5/commit/504009d1cb1915229b578fe11067968f1feb8d5a#diff-5eb6f0f98ac442eb459a3af2f8fc3afcc817a052a111766f022cf9e267431ca6).

PR #1699, opened 2025-12-09
Support cmake configuration for multi-config generators by default

connorjclark:
Without this patch, using `-G 'Ninja Multi-Config'` resulted in only the `RelWithDebInfo` configuration being generated, which was unexpected.


SiegeLord:
This is a long desired feature, thanks. I'm a little concerned that it looks so simple, we'll need to carefully test it with MSVC/XCode etc.

PR #1703, opened 2025-12-09
Fix undefined behavior caused by returning uninitialized D3D_STATE

hanpab:
Debug build with MSVC 2026 triggered an error dialog on first call to al_draw_filled_rectangle: "Run-Time Check Failure #3 - The variable 'state' is being used without being initialized.". Returning an uninitialized variable technically is UB.


SiegeLord:
Thanks! I'm not worried about uninitialized reads, the code was doing a decent-enough job at writing/reading a consistent subset of variables.

Issue #1702, opened 2025-12-09
`al_fseek` broken on current mingw32


fesh0r:
`al_fseek` seems to be broken when allegro is compiled for MINGW32 on an up to date MSYS2 install.

Had compiled a program that used `al_load_sample` and it was failing to load a standard wav file, but only when compiled for mingw32, the same code could load and play the same wav file correctly under mingw64, ucrt64, and clang64.

Tracing in gdb, any calls to `al_fseek` in `_al_identify_wav` and `wav_open` from `addons/acodec/wav.c` would fail with errno 22 "Invalid Argument" which was causing the wav header to not get parsed correctly due to not skipping over some fields.

I could reproduce this with the official MSYS2 mingw32 allegro package, the official liballeg.org mingw32 package, and when compiling allegro locally.

Testing with my locally compiled version, run_tests has several failures, and examples/ex_file gives the following output:

```
# line 34
OK   al_ftell(f)
OK   al_fgetc(f)
OK   al_fgetc(f)
OK   al_ftell(f)

# line 41
OK   al_fungetc(f, 0x55)
OK   al_ftell(f)

# line 46
OK   al_fread(f, bs, sizeof(bs))
OK   bs[0]
OK   bs[1]
OK   bs[2]
OK   bs[3]
OK   bs[4]
OK   bs[5]
OK   bs[6]
OK   bs[7]
OK   al_ftell(f)

# line 59
FAIL al_fseek(f, 13, ALLEGRO_SEEK_SET)
FAIL al_ftell(f)
FAIL al_fgetc(f)

# line 65
FAIL al_fseek(f, 0, ALLEGRO_SEEK_CUR)
FAIL al_ftell(f)
FAIL al_fgetc(f)

# line 71
OK   al_fungetc(f, 0x55)
FAIL al_fseek(f, 0, ALLEGRO_SEEK_CUR)
FAIL al_ftell(f)
FAIL al_fgetc(f)

# line 78
FAIL al_fseek(f, -3, ALLEGRO_SEEK_CUR)
FAIL al_ftell(f)
FAIL al_fgetc(f)

# line 84
FAIL al_ftell(f)
OK   al_fungetc(f, 0x66)
FAIL al_ftell(f)
OK   al_fseek(f, -2, ALLEGRO_SEEK_CUR)
FAIL al_ftell(f)
FAIL al_fgetc(f)

# line 93
OK   al_fseek(f, 0, ALLEGRO_SEEK_END)
[...]

PR #1705, opened 2025-12-11
prefer _fseeki64 and _ftelli64 if available

fesh0r:
For mingw32 builds using fseeko gives an "Invalid argument" error due to conflicting definitions of off_t.
fixes: #1702


SiegeLord:
Thanks for the investigation and the fix!

PR #1708, opened 2025-12-18
[SDL] Fix various issues with keyboard events

connorjclark:
1. ALLEGRO_EVENT_KEY_CHAR no longer has the wrong value for keycode when not using a US standard keyboard
2. ALLEGRO_EVENT_KEY_CHAR no longer has a potentially bad value for the modifiers state
3. ALLEGRO_EVENT_KEY_DOWN and ALLEGRO_EVENT_KEY_UP no longer set unichar, as that field isn't meant for these events

1: This resolves an issue reported by one of my users regarding poor keyboard support for non-US keyboards in the SDL driver (ref https://github.com/ZQuestClassic/ZQuestClassic/issues/1006).

Repro: Build with SDL; pressing the first key on the third row (US "Z", German "Y") should always set `event.keyboard.keycode` to 26. That's true for US standard keyboard, but if you use/emulate a QWERTZ keyboard (which swaps Y and Z on a standard US keyboard), then 25 is incorrectly set instead. That's because the driver makes the incorrect assumption that the unicode value can be used to map to the scancode.

2: I also noticed that the code was reading from two different parts of the SDL event union at the same time: namely for the modifiers key state, which it was incorrectly getting for the CHAR event. I actually considered removing modifiers from being set for the KEY_UP/KEY_DOWN events because they [are not documented](https://liballeg.org/a5docs/trunk/events.html), but I saw the windows driver was setting them and found [this commit](https://github.com/liballeg/allegro5/commit/5496090d2627b2f3027db25a9e18e08db0456459) so I guess the docs are just stale.

3: Lastly, I removed the code that was setting `unichar` on the KEY_UP/KEY_DOWN events. I similarly saw it is not documented as being on those event, but this time I also saw that the windows driver always sets them to 0 so I felt confident in removing them.

1 and 2 are clear bugs that I doubt anyone would have worked around. 3 could arguably be something people are relying on, so if you want that change behind a version check I can do that.

### DIFF
--- a/src/sdl/sdl_keyboard.c
+++ b/src/sdl/sdl_keyboard.c
@@ -25,6 +25,7 @@ typedef struct ALLEGRO_KEYBOARD_SDL
    int inverse[1024];
    int unicode[1024];
    int inverse_unicode[1024];
+   int last_down_keycode;
    bool create_extra_char[1024];
    ALLEGRO_DISPLAY *display;
 } ALLEGRO_KEYBOARD_SDL;
@@ -70,7 +71,6 @@ void _al_sdl_keyboard_event(SDL_Event *e)
    _al_event_source_lock(es);
    event.keyboard.timestamp = al_get_time();
    event.keyboard.display = NULL;
-   event.keyboard.modifiers = get_modifiers(e->key.keysym.mod);
    event.keyboard.repeat = false;
 
    if (e->type == SDL_TEXTINPUT) {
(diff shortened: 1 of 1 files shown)


src/sdl/sdl_keyboard.c
```diff
@@ -82,17 +82,21 @@ void _al_sdl_keyboard_event(SDL_Event *e)
...
          _al_event_source_emit_event(es, &event);
       }
-
    }
    else if (e->type == SDL_KEYDOWN) {
       event.keyboard.type = ALLEGRO_EVENT_KEY_DOWN;
-      event.keyboard.keycode = keyboard->table[e->key.keysym.scancode];
-      event.keyboard.unichar = keyboard->unicode[e->key.keysym.scancode];
+      event.keyboard.modifiers = get_modifiers(e->key.keysym.mod);
+      event.keyboard.keycode = keyboard->last_down_keycode = keyboard->table[e->key.keysym.scancode];
```
connorjclark, line 99: NOTE: don't merge. this needs to continue setting unichar, or _at least_ explicitly set it to 0 otherwise the field is set to a random value.

SiegeLord:
Continued in #1738

Issue #1674, opened 2025-09-14
Add `al_blend_color`


SiegeLord:
`al_blend_color(ALLEGRO_COLOR c1, ALLEGRO_COLOR c2, float factor)`

Note that this should go via the linear space (i.e. internally it'd call `al_color_rgb_to_linear`, blend and then convert back).

Optional: Add a version that does not do the sRGB conversion, although I can't imagine it being useful that often. Allegro colors are sRGB, and should be treated as such.

SiegeLord:
We could call it `al_interpolate_color`, so avoid confusion with the blenders, although I don't think it's a huge deal with proper documentation. Note, that I expect the outputs to be the same for equivalent blend modes (i.e. ADD + appropriate alphas).

SiegeLord:
Implemented in #1735

PR #1735, opened 2026-03-22
Add function al_mix_color (#1674)

Juroupi:
The "Since" version in the documentation needs to be updated


SiegeLord:
This looks good, thanks! The `Since` section should say "Since: 5.2.12"

SiegeLord:
Thanks!

PR #1736, opened 2026-03-25
[SDL] Add unichar assignment for extra key character events

connorjclark:
I noticed that on the SDL backend, holding Backspace would randomly insert the last key I pressed. The reason is that the `ALLEGRO_EVENT event` local var is not fully initialized, and this particular code path doesn't set `event.keyboard.unichar` - so the event's state is a bit random.


connorjclark:
Sorry, made a mistake here. There's no bug here, it was just in my fork 🫠

PR #1740, opened 2026-04-11
Second try to enable monotonic clock on Unixes.

SiegeLord:
This time, there's a configure check to check the existence of the necessary pthread functions.

Monotonic time has better behavior with timezone changes and other NTP nonsense.

TODO: Needs more manual tests, especially on MacOS.


pedro-w:
I checked on MacOS 26.4.1, compiles OK. Is there anything I can help with, not sure what you meant by "manual tests"?

SiegeLord:
Thanks for looking, @pedro-w. What we need to test is that `ex_timed_wait`, `ex_mixer_pp` and `ex_fs_window` still work. They have internally some timing code that depends on this the timing being correct.

Espyo:
Tested on Ubuntu 22.04, seems fine to me.

pedro-w:
All those examples work for me, AFAICT. Was tempted to dance around when  ex_mixer_pp was on, managed to resist.

SiegeLord:
Nice, thank you for testing @pedro-w!

PR #1745, opened 2026-05-15
Accept (0, 0) with ALLEGRO_FULLSCREEN_WINDOW on X11 and macOS

Derezo:
## Summary

`al_create_display` documents that under `ALLEGRO_FULLSCREEN_WINDOW` the passed
width and height are ignored initially and the display is sized to the
desktop. The Windows (wgl + d3d) backends already implement this by
substituting the target monitor's size during creation, so `(0, 0)` is a
valid "use the desktop size" request on Windows. The X11 and macOS
backends did not, so the documented contract was not delivered on those
platforms.

This change brings X11 and macOS in line:

- **X11 (`src/x/xdisplay.c`)**: `xdpy_create_display` rejected any `w <= 0
  || h <= 0` outright before it ever looked at the display flags. Resolve
  the target adapter's desktop dimensions via `_al_xglx_get_monitor_info`
  when `FULLSCREEN_WINDOW` is set and the caller passed a non-positive
  size; fall through to the existing `Invalid window size` error in all
  other cases.
- **macOS (`src/macosx/osxgl.m`)**: `create_display_win` built the
  temporary `NSWindow`/`ALOpenGLView` at the passed `(w, h)` — a 0-sized
  rect when callers relied on the documented contract — before later
  overwriting `display->w/h` with the screen size. Resolve the desktop
  size up front via `al_get_monitor_info` inside the existing
  `FULLSCREEN_WINDOW` branch.

Android (surface-driven sizing via JNI `onChange`) and the SDL backend
(`SDL_WINDOW_FULLSCREEN_DESKTOP`) already accept `(0, 0)` natively; no
changes needed.

Behaviour for non-zero sizes and for other flag combinations is
unchanged. The X11 `Invalid window size` error path still fires for plain
windowed/fullscreen requests with a non-positive size.

## Test plan

- [x] Builds cleanly on Linux/X11 (CMake + Ninja, Release): single
  rebuild of `liballegro.so.5.2.12`, no new warnings on `xdisplay.c.o`.
- [ ] Manual: `al_set_new_display_flags(ALLEGRO_FULLSCREEN_WINDOW);
  al_create_display(0, 0);` returns a display sized to the primary
  monitor on X11 (verified locally) and on macOS (needs reviewer
  verification — I don't have a macOS machine).
- [ ] Manual: `al_create_display(0, 0)` without `FULLSCREEN_WINDOW`
  continues to return `NULL` and log `Invalid window size 0x0` on X11.
- [ ] CI passes on the existing matrix.

## Notes for reviewers

- macOS `create_display_win` already substitutes the screen frame size
  into `display->w/h` after window creation (line 1704 in the
  pre-existing code). This change does the substitution earlier so the
  *initial* `NSRect` passed to `NSWindow initWithContentRect:` is also
  sane, which matters because `MINIMUM_WIDTH`/`MINIMUM_HEIGHT` clamping
  and the FULLSCREEN_WINDOW capability check via `ALOpenGLView` both
  occur on the original rect.
- The X11 `_al_xglx_get_monitor_info` / `_al_xglx_get_default_adapter`
  calls are made outside `system->lock`, which matches existing
  precedent: `xglx_get_monitor_info` and `xglx_get_num_video_adapters`
  in `src/x/xsystem.c` also call these helpers lock-free.
- No public API change, no header changes, no behavior change for
  existing well-formed callers.

### DIFF
--- a/src/x/xdisplay.c
+++ b/src/x/xdisplay.c
@@ -634,11 +634,12 @@ static ALLEGRO_DISPLAY *xdpy_create_display(int w, int h)
       if (_al_xglx_get_monitor_info(system, a, &mi)) {
          w = mi.x2 - mi.x1;
          h = mi.y2 - mi.y1;
+         ALLEGRO_DEBUG("FULLSCREEN_WINDOW: using desktop size %dx%d from adapter %d\n", w, h, a);
       }
    }
 
    if (w <= 0 || h <= 0) {
-      ALLEGRO_ERROR("Invalid window size %dx%d\n", w, h);
+      ALLEGRO_ERROR("Failed to determine a reasonable window size, got: %dx%d\n", w, h);
       return NULL;
    }
 
--- a/src/x/xdisplay.c
+++ b/src/x/xdisplay.c
@@ -610,11 +610,6 @@ static ALLEGRO_DISPLAY *xdpy_create_display(int w, int h)
       return NULL;
    }
 
-   if (w <= 0 || h <= 0) {
-      ALLEGRO_ERROR("Invalid window size %dx%d\n", w, h);
-      return NULL;
-   }
-
    flags = al_get_new_display_flags();
    if (flags & ALLEGRO_GTK_TOPLEVEL_INTERNAL) {
       if (gtk_override_vt == NULL) {
@@ -627,9 +622,28 @@ static ALLEGRO_DISPLAY *xdpy_create_display(int w, int h)
       }
    }
 
+   adapter = al_get_new_display_adapter();
+
+   /* FULLSCREEN_WINDOW ignores the passed size (see al_create_display docs);
+    * substitute the desktop size of the target adapter so this path matches
+    * the wgl/d3d behaviour and (0, 0) is accepted as a "use desktop" request.
+    */
+   if ((w <= 0 || h <= 0) && (flags & ALLEGRO_FULLSCREEN_WINDOW)) {
+      ALLEGRO_MONITOR_INFO mi;
+      int a = (adapter < 0) ? _al_xglx_get_default_adapter(system) : adapter;
+      if (_al_xglx_get_monitor_info(system, a, &mi)) {
+         w = mi.x2 - mi.x1;
+         h = mi.y2 - mi.y1;
+      }
+   }
+
+   if (w <= 0 || h <= 0) {
+      ALLEGRO_ERROR("Invalid window size %dx%d\n", w, h);
+      return NULL;
+   }
+
    _al_mutex_lock(&system->lock);
 
-   adapter = al_get_new_display_adapter();
    display = xdpy_create_display_locked(system, flags, w, h, adapter);
 
    _al_mutex_unlock(&system->lock);
--- a/src/macosx/osxgl.m
+++ b/src/macosx/osxgl.m
@@ -1534,6 +1534,21 @@ static ALLEGRO_DISPLAY* create_display_win(int w, int h) {
     * window can be created.
     */
    if (al_get_new_display_flags() & ALLEGRO_FULLSCREEN_WINDOW) {
+      /* FULLSCREEN_WINDOW ignores the passed size (see al_create_display
+       * docs); substitute the target adapter's desktop size so callers may
+       * pass (0, 0) and so the temporary view below is created at a sane
+       * extent.
+       */
+      if (w <= 0 || h <= 0) {
+         ALLEGRO_MONITOR_INFO mi;
+         int adapter = al_get_new_display_adapter();
+         if (adapter < 0)
+            adapter = 0;
+         if (al_get_monitor_info(adapter, &mi)) {
+            w = mi.x2 - mi.x1;
+            h = mi.y2 - mi.y1;
+         }
+      }
       __block BOOL ok;
       dispatch_sync(dispatch_get_main_queue(), ^{
          NSRect rc = NSMakeRect(0, 0, w,  h);


Derezo:
## Downstream validation

Built and exercised this branch against a real-world consumer
(internal project, X11/Mesa/Linux) to verify the documented
`ALLEGRO_FULLSCREEN_WINDOW` contract is now delivered end-to-end.

**Build**

- CMake 3.28 + Ninja, Release, `liballegro.so.5.2.12` plus the nine
  standard addons (image/primitives/font/ttf/audio/acodec/color/dialog/
  main). No new warnings; `xdisplay.c.o` compiles clean.
- Installed to an isolated prefix (`~/.local/allegro-patched`), wired
  in via `PKG_CONFIG_PATH`.

**Runtime**

Consumer was previously carrying a `al_get_monitor_info(0, &mi)` +
explicit `(w, h)` workaround for the X11 reject. Workaround removed;
fullscreen path now calls `al_create_display(0, 0)` with
`ALLEGRO_FULLSCREEN_WINDOW`. `ldd` confirms the binary loads the
patched `liballegro.so.5.2`; the process reaches the renderer-init
log line that fires after `al_create_display` returns, with no
`Invalid window size 0x0` error and no crash.

**Regression checks (X11)**

- Plain windowed mode (no FULLSCREEN_WINDOW): `al_create_display(W, H)`
  with positive dims unchanged — display creates at requested size.
- Windowed with `(0, 0)` and no FULLSCREEN_WINDOW: still hits the
  existing `Invalid window size` error path; behavior preserved.
- Fullscreen with `(0, 0)` + FULLSCREEN_WINDOW: now produces a
  desktop-sized display (was: NULL).
- Fullscreen with positive `(W, H)` + FULLSCREEN_WINDOW: unchanged —
  the substitution block is gated on `w <= 0 || h <= 0`.

**Not validated by me**

- macOS path: I don't have a Mac to run on. The change mirrors the
  existing post-creation `display->w/h = sc.size.width/height`
  substitution at line 1705 and stays inside the already-existing
  `FULLSCREEN_WINDOW` branch, but a reviewer with a macOS build
  environment confirming `create_display_win` is appreciated.
- CI-visible regression coverage: none added — gating this in tests
  would need an X server / NSScreen in CI, which isn't part of the
  current matrix. Happy to add a manual reproducer to
  `examples/` if useful.

Locking note (X11): `_al_xglx_get_monitor_info` and
`_al_xglx_get_default_adapter` are invoked outside `system->lock` in
this change. That matches existing precedent — the public
`xglx_get_monitor_info` / `xglx_get_num_video_adapters` entry points
in `src/x/xsystem.c` also call these helpers lock-free, and
`init_mmon_interface` is idempotent.

src/x/xdisplay.c
```diff
@@ -627,9 +622,28 @@ static ALLEGRO_DISPLAY *xdpy_create_display(int w, int h)
...
+   /* FULLSCREEN_WINDOW ignores the passed size (see al_create_display docs);
+    * substitute the desktop size of the target adapter so this path matches
+    * the wgl/d3d behaviour and (0, 0) is accepted as a "use desktop" request.
+    */
+   if ((w <= 0 || h <= 0) && (flags & ALLEGRO_FULLSCREEN_WINDOW)) {
+      ALLEGRO_MONITOR_INFO mi;
+      int a = (adapter < 0) ? _al_xglx_get_default_adapter(system) : adapter;
+      if (_al_xglx_get_monitor_info(system, a, &mi)) {
+         w = mi.x2 - mi.x1;
+         h = mi.y2 - mi.y1;
```
SiegeLord, line 636: Let's add an ALLEGRO_DEBUG here that indicates that we picked this particular resolution.

src/x/xdisplay.c
```diff
@@ -627,9 +622,28 @@ static ALLEGRO_DISPLAY *xdpy_create_display(int w, int h)
...
+      }
+   }
+
+   if (w <= 0 || h <= 0) {
+      ALLEGRO_ERROR("Invalid window size %dx%d\n", w, h);
```
SiegeLord: This should say "Failed to determine a reasonable window size, got: %dx%d\n", since we'd only get here if the monitor_info stuff failed (which I suppose it can sometimes)

SiegeLord (changes requested):
Thanks!

This seems like a reasonable idea, although I note you did not make any changes to the documentation. I'm not sure it's a great idea to document this until we nail down the precise semantics of what the passed in coordinates to `al_create_display` do when a fullscreen window is requested. For example, right now the undocumented behavior of this on Windows is that those initialize values, although ignored for the overall display size, are stored for later use in case you ever want to toggle the ALLEGRO_FULLSCREEN_WINDOW flag. I don't think we do this on other platforms.

Anyway, that just affects the documentation side of things. The actual implementation looks fine (except for the comments). I'll test the MacOS behavior later.

Derezo:
Thanks for the review! Addressed both inline comments in 8474434d1.

**`xdisplay.c:636`** — Added `ALLEGRO_DEBUG("FULLSCREEN_WINDOW: using desktop size %dx%d from adapter %d\n", w, h, a)` after the successful `_al_xglx_get_monitor_info` substitution so the picked resolution is visible in the debug log.

**`xdisplay.c:641`** — Reworded the error to `"Failed to determine a reasonable window size, got: %dx%d\n"`. With the new substitution block above it, this branch is now reached only when `_al_xglx_get_monitor_info` returns false (or the caller passed a bad size without `FULLSCREEN_WINDOW`), so the old "Invalid window size" wording no longer described the failure accurately.

**On the documentation question** — I checked `docs/src/refman/display.txt:98–104` (the `ALLEGRO_FULLSCREEN_WINDOW` flag description), which already states:

> *"the pixel dimensions of the created display will be the same as the desktop. The passed width and height are only used if the window is switched out of fullscreen mode later but will be ignored initially."*

This PR brings X11 and macOS in line with that already-documented contract, so I haven't touched the reference docs in this PR. The separate Windows-only behavior you mentioned — preserving the passed values for a later toggle out of fullscreen — is genuinely undocumented and platform-divergent today; happy to look at that as a follow-up if you'd like, but it feels out of scope for this change.

Rebuilt locally on X11 (CMake + Ninja, Release): `xdisplay.c.o` compiles clean, no new warnings, and the `ALLEGRO_FULLSCREEN_WINDOW` + `al_create_display(0, 0)` path still produces a desktop-sized display end-to-end against the same downstream consumer I tested before.

Derezo:
@SiegeLord I'm using claude code, first time trying it with C and Allegro. It seems to work pretty decently

src/x/xdisplay.c
Derezo: applied this fix

src/x/xdisplay.c
Derezo: added this in

SiegeLord (approved):
Alright, this looks fine. I'll test this on MacOS soonish and if all is well, I'll merge.

SiegeLord:
I finally got around to turning on my Mac. This appears to work fine. Thanks for the contribution!

PR #1746, opened 2026-05-15
Add audio recorder getters

Derezo:
## Summary

`ALLEGRO_AUDIO_RECORDER` is the only audio creator object in the
`allegro_audio` addon without post-creation getters. Every peer object
exposes its full configured state:

| Type | Frequency | Depth | Channels | Buffer |
|---|---|---|---|---|
| `ALLEGRO_SAMPLE` | ✅ `al_get_sample_frequency` | ✅ | ✅ | ✅ `_length` |
| `ALLEGRO_SAMPLE_INSTANCE` | ✅ | ✅ | ✅ | ✅ |
| `ALLEGRO_AUDIO_STREAM` | ✅ | ✅ | ✅ | ✅ `_fragments` |
| `ALLEGRO_MIXER` | ✅ | ✅ | ✅ | — |
| `ALLEGRO_VOICE` | ✅ | ✅ | ✅ | — |
| `ALLEGRO_AUDIO_RECORDER` | ❌ | ❌ | ❌ | ❌ |

The reference example `examples/ex_record.c` lives with the gap by
hardcoding the recorder parameters as module-level constants
(`frequency`, `audio_depth`, `samples_per_fragment` at file scope) and
referring back to those constants when sizing the WAV header and the
playback stream that mirrors the recorder. Application code following
that pattern has the same workaround.

This change adds:

- `al_get_audio_recorder_frequency`
- `al_get_audio_recorder_depth`
- `al_get_audio_recorder_channels`
- `al_get_audio_recorder_fragment_count`
- `al_get_audio_recorder_samples`

Each is a one-line accessor over a field the internal struct already
stores (see `addons/audio/allegro5/internal/aintern_audio.h`). No
behavioural change; the getters return the values the caller passed to
`al_create_audio_recorder`. Doc entries follow the same terse style as
the mixer/stream getters in the existing reference manual, and each new
function carries the existing `[Unstable API]` marker that the rest of
the recorder API has — this PR doesn't tighten or relax that marker.

## Test plan

- [x] `liballegro_audio.so.5.2.12` builds cleanly on Linux (CMake +
  Ninja, Release), with no new warnings on `recorder.c.o`.
- [x] `nm -D` confirms the five new symbols are exported.
- [ ] No new examples added (intentional — the existing `ex_record.c`
  demonstrates the use case and could be refactored to call the new
  getters in a separate change; this PR keeps the scope minimal).

## Notes for reviewers

- I did not touch `misc/liballegro_audio_abi.xml` since that file is
  regenerated from a built `.so` at release time (matches existing
  workflow visible in git history — `Prepare a summary of changes for
  5.2.x` commits).
- No `docs/src/changes-5.2.txt` entry: changelog is maintainer-curated
  at release time, same precedent.
- Choice of `unsigned int` for `frequency`, `fragment_count` and
  `samples` matches the types in `al_create_audio_recorder`'s
  signature and the corresponding fields on the internal struct.

## Motivation

Surfaced while integrating the recorder into a downstream voice-chat
pipeline (Opus encode at 48 kHz mono). The downstream code wanted to
assert at startup that the recorder's actual configuration matched what
its packet header expected. Without getters, that's only possible by
keeping a parallel copy of every value passed in. With these accessors,
the assertion becomes
`assert(al_get_audio_recorder_frequency(r) == 48000);`.

### DIFF
--- a/addons/audio/allegro5/allegro_audio.h
+++ b/addons/audio/allegro5/allegro_audio.h
@@ -432,6 +432,11 @@ ALLEGRO_KCM_AUDIO_FUNC(bool, al_is_audio_recorder_recording, (ALLEGRO_AUDIO_RECO
 ALLEGRO_KCM_AUDIO_FUNC(ALLEGRO_EVENT_SOURCE *, al_get_audio_recorder_event_source,
    (ALLEGRO_AUDIO_RECORDER *r));
 ALLEGRO_KCM_AUDIO_FUNC(ALLEGRO_AUDIO_RECORDER_EVENT *, al_get_audio_recorder_event, (ALLEGRO_EVENT *event));
+ALLEGRO_KCM_AUDIO_FUNC(unsigned int, al_get_audio_recorder_frequency, (const ALLEGRO_AUDIO_RECORDER *r));
+ALLEGRO_KCM_AUDIO_FUNC(ALLEGRO_AUDIO_DEPTH, al_get_audio_recorder_depth, (const ALLEGRO_AUDIO_RECORDER *r));
+ALLEGRO_KCM_AUDIO_FUNC(ALLEGRO_CHANNEL_CONF, al_get_audio_recorder_channels, (const ALLEGRO_AUDIO_RECORDER *r));
+ALLEGRO_KCM_AUDIO_FUNC(unsigned int, al_get_audio_recorder_fragment_count, (const ALLEGRO_AUDIO_RECORDER *r));
+ALLEGRO_KCM_AUDIO_FUNC(unsigned int, al_get_audio_recorder_samples, (const ALLEGRO_AUDIO_RECORDER *r));
 ALLEGRO_KCM_AUDIO_FUNC(void, al_destroy_audio_recorder, (ALLEGRO_AUDIO_RECORDER *r));
 
 #endif
--- a/addons/audio/recorder.c
+++ b/addons/audio/recorder.c
@@ -138,6 +138,46 @@ ALLEGRO_EVENT_SOURCE *al_get_audio_recorder_event_source(ALLEGRO_AUDIO_RECORDER 
    return &r->source;
 }
 
+/* Function: al_get_audio_recorder_frequency
+ */
+unsigned int al_get_audio_recorder_frequency(const ALLEGRO_AUDIO_RECORDER *r)
+{
+   ASSERT(r);
+   return r->frequency;
+}
+
+/* Function: al_get_audio_recorder_depth
+ */
+ALLEGRO_AUDIO_DEPTH al_get_audio_recorder_depth(const ALLEGRO_AUDIO_RECORDER *r)
+{
+   ASSERT(r);
+   return r->depth;
+}
+
+/* Function: al_get_audio_recorder_channels
+ */
+ALLEGRO_CHANNEL_CONF al_get_audio_recorder_channels(const ALLEGRO_AUDIO_RECORDER *r)
+{
+   ASSERT(r);
+   return r->chan_conf;
+}
+
+/* Function: al_get_audio_recorder_fragment_count
+ */
+unsigned int al_get_audio_recorder_fragment_count(const ALLEGRO_AUDIO_RECORDER *r)
+{
+   ASSERT(r);
+   return r->fragment_count;
+}
+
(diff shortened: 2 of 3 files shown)


Derezo:
## Cross-platform / standards validation

Ran a follow-up audit before asking for review. Findings below; no
remediation needed.

**Cross-platform behaviour**

All five fields read by these getters (`frequency`, `depth`, `chan_conf`,
`fragment_count`, `samples`) are populated in `recorder.c:38-42` before
the backend's `allocate_recorder()` is invoked. Greppable across every
backend (`alsa.c`, `pulseaudio.c`, `oss.c`, `openal.c`, `dsound.cpp`,
`aqueue.m`, `sdl_audio.c`, `opensl.c`):

```
grep -n "r->frequency\s*=\|r->depth\s*=\|r->chan_conf\s*=\|
        r->fragment_count\s*=\|r->samples\s*="
```

returns hits only in `recorder.c` (the creator). Every backend mention
is a read-only comparison (`r->depth == ALLEGRO_AUDIO_DEPTH_*` etc.).
The fields are effectively immutable after creation, so getter
behaviour is identical on Linux/ALSA, Linux/Pulse, Linux/OSS,
Linux/OpenAL, Windows/DSound, macOS/AudioQueue, Android/OpenSL, and
the SDL backend.

**Backward compatibility**

Additive only:

- No existing function signature changes
- No struct field changes (`aintern_audio.h` untouched)
- No macro changes
- No removed exports

Soname stays at `liballegro_audio.so.5.2` — adding symbols within an
ABI version is compatible by the standard contract, and matches the
project's own additive-API track record across the 5.2.x point
releases.

**C++ linkage / Windows DLL**

Declarations sit inside the existing `#if defined(ALLEGRO_UNSTABLE)
|| defined(ALLEGRO_INTERNAL_UNSTABLE) || defined(ALLEGRO_KCM_AUDIO_SRC)`
guard and within the file's `extern "C"` block, so C++ consumers get
correct linkage and the unstable-API opt-in is preserved. The
`ALLEGRO_KCM_AUDIO_FUNC` macro is used identically to existing
declarations, so MSVC `__cdecl` + dllexport/dllimport plumbing,
MinGW `extern`, BCC32, and the default `AL_FUNC` path all apply
uniformly.

**Const-correctness**

`const ALLEGRO_AUDIO_RECORDER *r` matches every sibling getter:
`al_get_sample_frequency(const ALLEGRO_SAMPLE *)`,
`al_get_audio_stream_frequency(const ALLEGRO_AUDIO_STREAM *)`,
`al_get_mixer_frequency(const ALLEGRO_MIXER *)`,
`al_get_voice_frequency(const ALLEGRO_VOICE *)`.

**Thread safety**

The five fields are written once at creation (before
`al_start_thread(r->thread)` runs at `recorder.c:81`) and never written
again. Sibling getters across the addon use the same lock-free pattern;
this PR matches.

**House-style compliance**

`.editorconfig` (3-space, LF, no tabs, no trailing whitespace, final
newline) and `indent.pro` (K&R, `-i3`, `-l80`): confirmed clean across
all three touched files. The two implementation lines that go to 82-84
chars match the surrounding file's existing >80-char lines (e.g.
`recorder.c:49`, `:63`, `:136`), so the deviation is consistent with
local style.

**Out-of-scope observation**

While auditing, noticed `recorder.c:40` has `r->frequency = frequency,`
(comma instead of semicolon). It works by accident because the next
line `r->depth = depth;` terminates the expression statement. Pre-
existing, not introduced by this PR, and intentionally left alone to
keep this PR scoped to the additive getter API.

SiegeLord:
Seems reasonable! Thanks!

@Derezo, what AI are you using for this, out of curiosity?

Derezo:
> Seems reasonable! Thanks!
> 
> @Derezo, what AI are you using for this, out of curiosity?

Claude code with opus 4.7 on xhigh -- I'm testing with a game called "Parasite": https://github.com/Derezo/parasite
I have the claude agents / CLAUDE.md committed there, but in my user scoped CLAUDE.md I have a directive that tells it to update the UPSTREAM_ASKS.md file if it finds problems in a dependency, so these bugs were me cleaning up that file. It had found 3 items but one of them turned out to be a false flag and it didn't raise the PR for that one

PR #1747, opened 2026-05-15
Add sample finished events (for mixers)

chruffins:
Addresses #1670 for mixers. Sample instance event source now gets initialized and destroyed, and is used to emit a new event type ALLEGRO_EVENT_AUDIO_SAMPLE_FINISHED. Also wrote up documentation and an example demonstrating usage. Haven't touched voices yet, since detecting when a sample instance finishes seems to need backend specific implementations.


addons/audio/allegro5/internal/aintern_audio.h
SiegeLord: Let's remove this comment, it is no longer accurate.

addons/audio/kcm_instance.c
```diff
@@ -205,6 +207,7 @@ void _al_kcm_destroy_sample(ALLEGRO_SAMPLE_INSTANCE *spl, bool unregister)
       }
 
       _al_kcm_detach_from_parent(spl);
+      al_destroy_user_event_source(&spl->es);
```
SiegeLord: I think this should go into al_destroy_sample_instance (surronded by `if (spl)`  because we also call `_al_kcm_destroy_sample` when let's say a mixer is destroyed, and we never initialized there.

SiegeLord (changes requested):
Thanks! This looks very good. Thank you for adding an example as well.

SiegeLord (approved):
Thanks! This looks great.

PR #1748, opened 2026-05-19
Document Android-specific behavior with resize events

chruffins:
Added documentation for some behavior that I ran into a while ago with running Allegro on Android. It seems that a resize event always follows a drawing resume event due to the implementation in `android_display.c`.


SiegeLord:
Thanks! I wonder if these are bugs we could fix...

PR #1754, opened 2026-06-01
Fix encoding size description for pixel formats

catgofire:
Corrected the encoding size description for DXT1, DXT3, and DXT5 formats from bytes to bits.

See: https://developer.valvesoftware.com/wiki/DirectX_Texture_compression_5 https://en.wikipedia.org/wiki/S3_Texture_Compression


SiegeLord:
Thank you!

PR #1759, opened 2026-07-14
Windows: Ignore first mouse event when window activates

connorjclark:
When the window is not active, clicking on it resulted in a mouse event being sent to the application. This is generally undesirable behavior, and others backends (at least OSX) already ignore such events.

You can validate this with ex_mouse demo:

1. Move some other window to the foreground
2. Click on the mouse demo

Before this commit, it will respond to that first click. After, it does not.

Dragging the window from an unfocused state also still works.

This is a regression relative to Allegro 4's Windows backend.

### DIFF
--- a/src/win/wwindow.c
+++ b/src/win/wwindow.c
@@ -858,6 +858,17 @@ static LRESULT CALLBACK window_callback(HWND hWnd, UINT message,
                break;
          }
          return 1;
+      case WM_MOUSEACTIVATE:
+         /* By default a click on an inactive window both activates it and
+          * gets delivered to the window. Eat client-area clicks so the
+          * click that refocuses the window doesn't register as a mouse
+          * press. This matches OS X, where -acceptsFirstMouse: defaults
+          * to NO. Non-client clicks (title bar, borders) keep the default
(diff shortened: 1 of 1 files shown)


Espyo:
The pre-PR event seems to be how the vast majority of applications work, both from what I remember of when I used Windows as well as my day-to-day usage of Linux (Openbox WM). If this behavior is to be added, it should be locked behind a configuration and disabled by default, in my opinion.

SiegeLord:
If Allegro filters out the event, there's no recourse to do anything else. I glanced at the MacOS/Windows guidelines, and found there are none (yes MacOS defaults to not having clickthrough, but the documentation says that most controls, e.g. buttons, do allow it).

I think we should:

1. Make sure the user is capable of distinguishing first vs second clicks
2. Default to sending first clicks (i.e. bring MacOS into consistency with the other platforms)
3. Add a compatibility flag for MacOS to let apps opt into the older behavior

connorjclark:
The direction is reasonable, but it is more than I'm looking to take on here. I'll leave this PR as-is in case it's a useful starting point for someone who wants to pursue the fuller approach. Feel free to close if it doesn't fit.

PR #1769, opened 2026-08-16
Wayland backend

chruffins:
An insane project that I started in June, but this adds a Wayland driver to Allegro 5. It is still very unfinished but supports the display lifecycle among a few other things, like input.

Current disclaimers:
- I largely learned about developing with Wayland by doing research for this project


SiegeLord:
Thanks for the build-time XML->C conversion, I was going to comment on that. Shouldn't we just use the ones on the system? E.g. for me:

```
$ dpkg -S cursor-shape-v1.xml           
wayland-protocols: /usr/share/wayland-protocols/staging/cursor-shape/cursor-shape-v1.xml
$ dpkg -S xdg-shell.xml      
wayland-protocols: /usr/share/wayland-protocols/stable/xdg-shell/xdg-shell.xml
$ dpkg -S pointer-constraints-unstable-v1.xml
wayland-protocols: /usr/share/wayland-protocols/unstable/pointer-constraints/pointer-constraints-unstable-v1.xml
$ dpkg -S xdg-decoration-unstable-v1.xml     
wayland-protocols: /usr/share/wayland-protocols/unstable/xdg-decoration/xdg-decoration-unstable-v1.xml
```

Still haven't gotten a chance to take a close look at other details.

chruffins:
> Thanks for the build-time XML->C conversion, I was going to comment on that. Shouldn't we just use the ones on the system?

I agree, I switched it over to scan at build-time instead.

PR #1773, opened 2026-09-01
Vendor minimp3 + use mp3 streaming callbacks

chruffins:
This PR changes the MP3 audio stream implementation to use callbacks for streaming!


SiegeLord:
I don't want to start vendoring things.

1. Distros/homebrew don't like people doing that
2. It's actively developed

What should be done is tell the lazy minimp3 maintainer to spend 5 seconds to make a release on github, which is the minimum of what's required to get that stupid library get packaged. This situation is entirely their fault and has been easily avoidable, for years.

As for the streaming... looks like it's using newer API than when that file was written. I see they actually added seeking. It's probably a good idea, but I'd need to carefully test this.

chruffins:
Okay, I reverted the commit that vendors minimp3. I opened an issue on the repository to ask for versioned releases as well, but I wonder if a fork would be necessary to get releases.

pedro-w:
Is there another mp3 implementation that might be easier to work with?

addons/acodec/mp3.c
```diff
@@ -192,126 +183,93 @@ static bool mp3_stream_set_loop(ALLEGRO_AUDIO_STREAM * stream, double start, dou
...
-         if (frame_samples == 0) {
-            mp3_stream_rewind(stream);
-            break;
-         }
-         mp3file->frame_pos = 0;
-         mp3file->next_frame_offset += frame_info.frame_bytes;
-      }
+   samples_read = mp3dec_ex_read(&mp3file->dec, data, samples_needed);
+   if (mp3file->dec.last_error) {
+      ALLEGRO_WARN("MP3 stream decode failed: %d.\n",
```
SiegeLord, line 210: Why is this a warning, does this happen often?

addons/acodec/mp3.c
```diff
@@ -192,126 +183,93 @@ static bool mp3_stream_set_loop(ALLEGRO_AUDIO_STREAM * stream, double start, dou
...
-            mp3file->file_size - mp3file->next_frame_offset,
-            mp3file->frame_buffer, &frame_info);
-         if (frame_samples == 0) {
-            mp3_stream_rewind(stream);
-            break;
-         }
-         mp3file->frame_pos = 0;
-         mp3file->next_frame_offset += frame_info.frame_bytes;
-      }
+   samples_read = mp3dec_ex_read(&mp3file->dec, data, samples_needed);
```
SiegeLord, line 208: This still needs to be a while loop to handle long buffers and short mp3s.

SiegeLord (changes requested):
This looks pretty good aside from the while loop removal.

I tested this and in fact, it returned a more accurate length for the file that I tested with, so it's an improvement in terms of correctness as well!

SiegeLord:
> Is there another mp3 implementation that might be easier to work with?

libmpg123 is probably best option, ~~just a little tricky to package and~~ some people might object to the LGPL nature of it: https://www.mpg123.de/index.shtml

EDIT: By "just a little tricky to package" I mean that it will complicate efforts to package Allegro, obviously libmpg123, being a library developed by civilized people, is widely packaged in Linux distros and homebrew.

addons/acodec/mp3.c
chruffins: It should only happen for invalid data but I added it since I thought it might be helpful for debugging. I just switched it to only emit once per playback though.

addons/acodec/mp3.c
chruffins: Fixed!

SiegeLord (changes requested):
Ok, turns out I was wrong about the while loop. We don't need it (although there is a bug in some of the calling code that I mistook for being a bug with mp3). I think all we need to do here is get rid of that commit (good job on keeping it separate) and we're good to go. Sorry for making you write that while loop!

SiegeLord:
Merged as b0252504482d8d462eec494c393325e9cd991865

Thanks!

PR #1774, opened 2026-09-09
[SDL] Implement set_system_mouse_cursor

connorjclark:
The SDL backend never implemented this, so al_set_system_mouse_cursor never changed the cursor. Map each Allegro system cursor to the closest SDL_SystemCursor and set it with SDL_SetCursor. On Emscripten these become the browser's CSS cursors.

Cursors are created lazily with SDL_CreateSystemCursor and kept for the life of the process.

### DIFF
--- a/src/sdl/sdl_display.c
+++ b/src/sdl/sdl_display.c
@@ -352,12 +352,72 @@ static bool sdl_set_mouse_cursor(ALLEGRO_DISPLAY *display,
    return true;
 }
 
+/* SDL has system cursors of its own (on Emscripten they become CSS cursor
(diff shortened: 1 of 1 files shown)


SiegeLord:
Huh, interesting how neither library covers all the possible cursors. Allegro seems to be missing opposite direction ones while SDL is missing the unidirectional ones. And no question mark for SDL?

We should backfill the missing ones into Allegro.

As for this PR, I'll try testing it soon. Thanks!

SiegeLord:
I tested it, works great!

SDL has some creative choices for some of the icons... but it is what it is.

Thanks again!

PR #1776, opened 2026-09-22
Fix SDL_GameControllerDB parsing when there's unexpected whitespace at the begginning/end of the line.

### LINKED ISSUE #1772: al_set_joystick_mappings_f fails when file includes blank lines
Using the file directly from https://github.com/mdqinc/SDL_GameControllerDB/blob/master/gamecontrollerdb.txt results in an error here:

https://github.com/liballeg/allegro5/blob/master/src/joynu.c#L570

The reason is the file has \n\r newlines, and so Allegro bails when it hits that. Instead, it should ignore lines that just have whitespace (or starts with a #).

SiegeLord:
Fixes #1772


connorjclark (approved):

SiegeLord:
Thanks for taking a look, @connorjclark !

